PR #33, opened 2016-04-30
Define a shell script to generate new examples.

ckundo:
- Use this script from the command line to copy the template.
- This can get more sophisticated over time, or replaced with a more
  robust template generator. For now, it captures in code what would
  otherwise be undocumented or documented poorly in the README.

## Example usage

```
aria-practices % ls examples
alert           button          checkbox        coding-template css      js              link            radio           slider
aria-practices % bin/generate date-picker
aria-practices % head -10 examples/date-picker/date-picker.html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8" />
  <title>ARIA Example: Pattern Name: Coding Example Name</title>
  <meta name="generator" content="BBEdit 10.5" />
  <link href="../css/core.css" rel="stylesheet">
  <link href="../css/table.css" rel="stylesheet">
  <link href="css/codingtemplate.css" rel="stylesheet">
</head>
```

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -1,0 +1,12 @@
+#!/usr/bin/env sh
+
+set -e
+
+example_name="$@"
+
+# Create a directory for the example name.
+mkdir -p examples/"$example_name"
+
+# Copy the template to the new example file.
+cp examples/coding-template/template3.html \
+  examples/"$example_name"/"$example_name".html
--- a/README.md
+++ b/README.md
@@ -15,16 +15,16 @@ future ARIA features we need to implement.
 * To keep code in this repository consistent; editors should use a text editor
 that supports [EditorConfig](http://editorconfig.org/).
 
-## Code examples
+### Writing code examples
 
 - Choose a pattern from the
   [Design Patterns Status](https://github.com/w3c/aria-practices/wiki/Design-Patterns-Status)
   document that is missing a code example.
 - Via the command line, run:
 
-      $ bin/generate <name>
+        $ bin/generate <name>
 
-  Where `<name>` is the camel-cased name of the design pattern example. This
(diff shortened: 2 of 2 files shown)


ckundo:
I'm assuming the best practice here is to copy the `coding-template/template3.html`; this script could easily be updated to use a different file.

MichielBijl:
Can we make this use dashes instead of camel case? Don't think that would be a problem given the variable is already between quotes. I don't know anything about shell script; so not sure it would work with the command itself.

bin/generate
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+
+set -e
+
+example_name="$@"
```
ckundo: @MichielBijl the convention I've seen in shell scripts is to use underscores. Google's styleguide suggests using underscorized variable names (as with function names) https://google.github.io/styleguide/shell.xml?showone=Function_Names#Function_Names

MichielBijl:
@ckundo okay, but that is not the convention we use in the examples; see coding-template and menu-button folder.

ckundo:
@MichielBijl thanks for the followup. I might be misunderstanding your suggestion:

> Can we make this use dashes instead of camel case?

Do you mean the name of the file, or the variable names, should use dashes? In other words, are you referring to `example_name` or `datePicker`?

MichielBijl:
@ckundo sorry for not being clearer. I meant the stuff that is generated. Could you make it generate `date-picker.html` in a `date-picker` folder instead of `datePicker.html` in a `datePicker` folder?

mcking65:
Thanks @ckundo, I've been doing this manually. Not a huge deal. But, I will be stubbing out several examples soon and, with some adjustments,  this might come in handy.

Issue #104, opened 2016-08-29
Develop examples of data grids using the grid design pattern


mcking65:
Related to issue #29.

mcking65:
Integrated initial set of commits into master.
Most recent: commit b050f6431cb7102b8fb71d1bbb8f60b9b46d85a8

Work in progress; making changes to editable content appearance and ARIA properties.

mcking65:
@accdc, @a11ydoer, @annabbott, @jnurthen 

Bryan, Jemma, Ann, James,

I would like each of you to please review the 
[data grid examples page](http://w3c.github.io/aria-practices/examples/grid/dataGrids.html).

Thank you!

mcking65:
Created issue 160 for capturing feedback on the development to make it easier for reviewers to see what needs review and follow the conversation. Closing 104 as initial development is complete.

PR #111, opened 2016-09-01
Fix rowcount

tatermelon:


mcking65:
Thank you @tatermelon; merging in as update to issue #104.

PR #119, opened 2016-10-02
Menubar

a11ydoer:
Changed copyright to w3c and fixed file name errors in JS and HTML


michael-n-cooper:
Thanks, the copyright looks like something we can live with for now.

The automated checks are failing: https://travis-ci.org/w3c/aria-practices/builds/164478496

I don't know if this is a result of the code Jesse added or something else, and whether it's important or not, but in some way we should get that to pass before accepting the pull request, as it will break automated generation of the editors' draft. Or accept the pull request and then go in to fix it, just knowing the editors' draft won't build until it's fixed.

mcking65:
@a11ydoer will be correcting jscs errors and resubmitting so closing.

PR #143, opened 2016-10-30
jscs command fix - missing one dash

a11ydoer:
The jscs command should be `npm run jscs --fix`. The current doc is missing one dash in the command.


mcking65:
@a11ydoer,

The command `npm run jscs-fix` is an alias defined in package.json for the command:

`jscs -c .jscsrc -m 1000 --fix bin examples`

The README.md in master is correct. It is better to use that alias.

So, I am closing this pull request.

a11ydoer:
yeah, I see that from package.json file. 
"jscs-fix": "jscs -c .jscsrc -m 1000 --fix bin examples" ,

It may be my config file issue then. thanks for pointing out package.json file.

Issue #67, opened 2016-07-18
Develop example Accordion


jnurthen:
Change background image for expand collapse to something that works in high contrast mode

jnurthen:
@accdc

MichielBijl:
Which branch is this in?

accdc:
I think we are waiting for Matt to insert the code into the template after his code review.

mcking65:
I added a first draft of the design pattern and integrated Bryan's example code in commit 1ecb9be.

I published it in the master branch, so it can be viewed at:
http://w3c.github.io/aria-practices/#accordion

mcking65:
From August 29, 2016 meeting of APG task force, @jnurthen will remove web fonts and adjust colors .

carmacleod:
This example has an almost-black background, which makes it almost impossible to read.
The problem is the body style:
    background-color: #2E3135;
in file global.css, line 41.
If I completely delete that style, then the example page has a white background, which is way better.

arbydev:
Hi, could you see any issue with making the accordion headings as H2 or H3 depending on the page hierarchy?

mcking65:
@arbydev 

No issue there; heading levels should be appropriate for the page information architecture. @accdc will be addressing this both for this specific example and in the comments that accompany it.

mcking65:
With the latest commits by @jnurthen and myself and the opening of issue 210, this development work is now complete.

Many thanks to @accdc for all the original code for this!

PR #156, opened 2016-11-13
Accordion CSS fixes

jnurthen:
Removed pretty much all the extra CSS. It only confused the example.


jnurthen:
#67 should be more legible now

PR #171, opened 2016-12-02
Update Radio group examples

jongund:
1. Updated documentation to use latest format
2. Fixed bug in radio 1 example, second radio group was not initialized


mcking65:
Thank you Jon!

Issue #160, opened 2016-11-18
Review data grid examples


mcking65:
The data grid examples developed for issue #104 are ready for review. The example page is at:http://w3c.github.io/aria-practices/examples/grid/dataGrids.html

#### Requested Reviews

- [x] Review by @annabbott 
- [x] Review by @accdc 
- [x] Review by @jnurthen

accdc:
Hi Matt, been testing the examples, and some of the grid functionality doesn't always match the instructions at the bottom of each grid, specifically which columns are sortable like Description which is actually static, also which column cells are editable such as Amount.

Another question about aria-rowindex, I thought this was to be used on role=row, <tr> and not role=gridcell <td>, because a rowindex is relative to the row, not the horizontal cell, and what if focus is on the TR instead of the TD, such as when arrowing up and down between selectable grid roles?

accdc:
Looks like GitHub messed up my markup.

Basically I was referring to the scenario when TR rows have focus and the up / down arrow keys are used to move between them, where aria-rowindex is valid. I was unaware that aria-rowindex is valid on TD, because it seems like a lot of duplication, why not just set this once on TR?

mcking65:
@accdc, thank you, thank you. Some excellent catches!

You are right, my notes below the example are out of sync with the example. I'll fix ASAP ... Thanks Giving vacation is about to begin for me so not sure if I will sneak it in before next week.

Also, on rowindex, it is valid on cells, but when on cells, it should also be on the row. We will fix that.  For the purpose of demonstrating assistive technology support, we may want to have it on the rows and not the cells in one example and on both rows and cells in another.

jongund:
Visual styling of what cell the keyboard focus is on also should be improved.   Currently seems to rely on the default browser focus and is difficult to see.

annabbott:
Reviewed Data Grid Examples (http://w3c.github.io/aria-practices/examples/grid/dataGrids.html):

First paragraph:
"Following are three example implementations of the design pattern for grid that demonstrate the keyboard interactions and ARIA features that enable accessible presentation of tabular information. "

Suggested edit:
""Following are three example implementations of the design pattern for grid that demonstrate the keyboard interactions and ARIA features that enable accessible presentation of tabular information, some of which is interactive. "

Without specifying "interactive", there's no difference between a grid and a data table.

annabbott:
[...]

PR #174, opened 2016-12-02
Move rowindex to row instead of col

tatermelon:


mcking65:
Thank you @tatermelon!! This is exactly what we need to address the feedback on rowindex in issue #160.

PR #190, opened 2016-12-07
Proposal for alternative format of presentation section

jongund:
Matt,

My github skills are improving, so you are probably getting more feedback than you want on the presentation section.   But I made some edits that I think make it clearer to new authors what effects are and applicable rules, without really calling them rules.   I think rules belong more in the ARIA spec and in validation tools.  I think we should use HTML elements in the examples rather than role values, because that is 99% of the use case in the real world.   Most developers do not think in terms of roles, they think in terms of elements.

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -3430,14 +3430,16 @@ helper.stopEvent(evt); //stop the event if not a tab keypress
       <h3>
         <code>presentation</code> role examples
       </h3>
-      <p>
-        The
+      <ul>
+        <li>The
         <a href="examples/presentation/PresentationRoleExamples.html">
           <code>presentation</code> role examples page
         </a>
         illustrates the effects of these rules in a variety of scenarios and provides more detailed
-        explanations of the rationale behind them.
-      </p>
+        explanations of the rationale behind them.</li>
+        <li><a href="examples/menu-button/menu-button-2/menu-button-2.html">Menu Button Example: links</a>
+        <li><a href="examples/menubar/menubar-1/menubar-1.html">Navigation Menubar Example</a>
+      </ul>
     </section>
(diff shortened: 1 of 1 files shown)


mcking65:
@jongund, 

Thank you ... I will do just a little tweaking of your commit in a presentation branch and then share out. You make some good points. But, there are a few issues with the text I would like to address.

BTW, you have a clause in there that states alt text on images is remains visible to AT ... it is important to understand that alt text does indeed evaporate as it is an attribute of an element that is completely removed from the tree. When an element is presentational, all of its attribs are wiped out.

That is one of the reasons, btw, for the global attrib exclusion. It has the power to wipe out an element, but if there is info not specific to the role, then presentation loses its fire power.

jongund:
If the IMG element does not have role=presentation and has ALT test and is a descendant of an element with role=presentation, that ALT text will be part of the content text of the element with role="presentation"

mcking65:
@jongund,

Yes, if an image is inside of a presentational element, it is not affected by presentation. But, that is not what the text was saying. Presentation only affects an image if it is directly applied to the image. It is not possible for an image to inherit presentation.

mcking65:
@jongund,

I have merged some of your ideas. Take a look at master now.

I have also made some changes to the language to be more accurate in a "specky" sort of way but without adding "specky" language ... I hope. I hope that will make the implications of the language more clear.

mcking65:
@jongund ... in case you are wondering ...

I didn't make a separate section for inheritance because inheritance itself is a core aspect of the effects. 

I didn't include a heading that summarizes conditions for ignoring presentation role because wording of the conditions in short title form creates ambiguities that are dangerous if misinterpreted.

PR #191, opened 2016-12-07
Fixed bug with focus not going to first item with JAWS

jongund:


mcking65:
@jongund, thank you! The fix is working!

PR #198, opened 2016-12-09
Updated Menu button to match current template

jongund:


mcking65:
@jongund, thank you very much for these updates
!

PR #199, opened 2016-12-09
Updated Radio button examples to conform to current example template

jongund:


mcking65:
@jongund, thank you for these improvements!

PR #200, opened 2016-12-09
Updated Checkbox examples to implement current example template and titling

jongund:


mcking65:
@jongund, thank you for these updates too! Look good.

PR #201, opened 2016-12-10
Updates for the accordion example. 

jnurthen:
Filled in the roles/properties section and updated the aria-level of the accordion to 3


jnurthen:
@mcking65 can you review. note - the close and reopen of the pull request is to get the travis build to pass.... it always fails the first time but if you do it again after a short amount if time it passes.

PR #203, opened 2016-12-10
Fix broken links

jnurthen:
Hopefully this fixes all of the broken links - it is hard to test until we get the gh pages version.

There were a couple of extra changes I had to make to add notes - or remove links to content which hasn't yet been written - and then a few lines got changed due to my editorconfig (spaces removed at the end of lines) but they shouldn't affect the output.

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -187,7 +187,7 @@
           </li>
 
           <li>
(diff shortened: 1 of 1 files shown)


mcking65:
@jnurthen,

Thank you James. Wow, I think it was me that accidentally left out a lot of hashes from the URIs ... sorry about that.

And, many thanks for doing this in a separate branch.

I am going to merge this in now.

PR #208, opened 2016-12-11
Updated checkbox examples description tables to conform to template and edited descriptions

jongund:


jnurthen:
@jongund  - just as an fyi I find that if I close and re-open a PR then the travis job gets restarted. It will normally complete the 2nd time. It is not that you have done anything wrong - just in case you are wondering.

PR #262, opened 2017-01-27
removed keypress events for treeview examples

jongund:


mcking65:
Thanks @jongund, this looks good. And, it fixes the issue with not being able to tab out of the treeview.

PR #263, opened 2017-01-30
Removed keypress event handler from menu button examples

jongund:


mcking65:
@jongund, thank you! I pushed 1 commit to fix jscs errors. I tested w/JAWS and Firefox, and it is working.

Issue #187, opened 2016-12-07
Review updates to breadcrumb example page


mcking65:
Make the following updates to  the [breadcrumb example page](http://w3c.github.io/aria-practices/examples/breadcrumb/index.html):

- [x] Implement the current [example template](https://github.com/w3c/aria-practices/blob/master/examples/coding-template/Example-Template.html).
- [x] Shorten title to "ARIA Breadcrumb Example"
- [x] Fix typo "markedup "
- [x] Add visual decoration separators via CSS (so they are not announced by SRs)

danieldafoe:
Is there an example of one of the example pages implemented with the new template?

MichielBijl:
@danieldafoe you mean like an actual page where this is implemented?

danieldafoe:
When I cloned the repo and wanted to follow the new template given, I just wanted to see an existing page that has implemented the new template (page structure) that should be followed. As a reference, say.

Hopefully that's clear enough.

MichielBijl:
Ahh, the [tabs example](http://w3c.github.io/aria-practices/examples/tabs/tabs.html) uses the new template although it’s missing the “Accessibility Features” section.

danieldafoe:
Thank you! I'll refer to that in the future.

And no, go right ahead. I'll take a look at yours once completed as well so I know for next time. Thanks for the help.

MichielBijl:
This is fixed in https://github.com/w3c/aria-practices/commit/4648a41260c2d7a63fe86ba99e78c7dfea50eff4 & https://github.com/w3c/aria-practices/commit/b00aeb6acada77fcdc6bebc2bb0aea033011aa93 in [the breadcrumb branch](https://github.com/w3c/aria-practices/tree/breadcrumb).

There is [pull request for this issue](https://github.com/w3c/aria-practices/pull/266).

- [ ] Review @mcking65 
- [ ] Review @jnurthen

mcking65:
I added 3 more commits; please have a look.

I see one more issue that I didn't touch. The separators are made with the content property so they are still announced by screen readers. I think an image would be much better so we can get rid of the unnecessary screen reader verbosity. If we don't do this, then we should change the accessibility properties subsection. However, I strongly believe those separators are pure decoration, especially since we have the list structure.

MichielBijl:
Last issue fixed in https://github.com/w3c/aria-practices/pull/283.

mcking65:
[...]

PR #266, opened 2017-01-31
Breadcrumb

MichielBijl:
Fix issues in https://github.com/w3c/aria-practices/issues/187. Please squash the commits in this PR.

### DIFF
--- a/examples/breadcrumb/index.html
+++ b/examples/breadcrumb/index.html
@@ -66,6 +66,18 @@
           <tr>
             <td></td>
             <th scope="row">
+              <code>aria-label=<q>Breadcrumb</q></code>
+            </th>
+            <td>
+              <code>nav</code>
+            </td>
+            <td>
+              Provides a label that describes the type of navigation provided in the <code>nav</code> element.
+            </td>
+          </tr>
+          <tr>
+            <td></td>
+            <th scope="row">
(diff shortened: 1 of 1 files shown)


mcking65:
@MichielBijl, 

I added 3 more commits.

If you and @jnurthen are OK with the changes, I see no need to keep all the individual commits in the history ... we should squash it.

MichielBijl:
I’m fine with all but the removal of the links to properties and elements. I for one find it very useful to have cross references to other specs and the like. This is also a step towards the more “developer reference”-ness of the APG. It makes it a lot easier to look up elements and attributes that you might not be familiar with.

We can discuss where it should link, but I figured TR would be the correct location.

Any arguments to take it out?

mcking65:
We do not have the role/state/property links in the example pages because we do not have a way of ensuring they point to the correct place. The example pages are not respec documents. During the editors' call, I had a discussion with Michael about isolating just the reference generation capabilities of respec for certain pages so we could do things like that. We do not want to turn the example pages into respec docs.

In general, I don't think we should link to other specs unless there is something unusual that justifies it. For example, I think it is better not to have links to the spec for common html elements. They could be more of a distraction than a help to readers.

Would you be interested in taking on issue #270?

PR #282, opened 2017-02-13
updated disclosure examples to restore use of DL/DT/DD and use button element 

jongund:
updated example to:
1.  Use images instead of entities for visual state information 
2. Removed the use of DL/DT/DD elements in FAQ example


examples/disclosure/disclosure-faq.html
```diff
@@ -226,7 +226,7 @@ <h2 id="sc1_label">HTML Source Code</h2>
       aria-label="End of"
     ></div>
     <script>
-          sourceCode.add('sc1', 'ex1');
+          sourceCode.adiv('sc1', 'ex1');
```
mcking65: Why this change? It breaks display of source code.

mcking65:
@jongund, thank you.

I made a few more changes and pushed them to this branch.

I put the answers to the questions in the FAQ inside of `p` elements that are contained by the `dd` elements so that the `dl` is always well-formed, even if the answers are all hidden.

I made some editorial revisions and corrections.

I'll merge into master. Please have a look.

jongund:
Sounds good.  Thanks jon.

PR #284, opened 2017-02-13
Treeview has been updated to use images instead of entities for visual state information in example 2

jongund:


mcking65:
@jongund, thank you!

PR #302, opened 2017-02-21
Grid indicators

tatermelon:


mcking65:
@tatermelon, looks great! 

I can think of one reason why we may want to load the nux only when focus is inside the grid and the user presses tab. That is that the nux is for keyboard users. There is no reason to show it to everyone, and I would imagine that most sites would not want to show to everyone who comes to the page.

MichielBijl:
Can I ask what a NUX is? Assuming this isn’t about the Mad Max character.

tatermelon:
@MichielBijl NUX stands for 'New User Experience'. In this case, it is a contextual dialog that explains how to use the grid keyboard behaviors:

Screen shot of NUX dialog

PR #303, opened 2017-02-22
Check for this.deleteButton before trying to set an attribute

thomascorthals:
After a select item is toggled, attribute aria-disabled on this.deleteButton was always set. This results in an error if no deleteButton is set.


thomascorthals:
I don't know why the Travis CI build failed twice. My change seems simple enough to not cause any issues.

MichielBijl:
@thomascorthals Travis is a bit of a PITA because it tries to curl some W3C URL and often fails for some reason. This all looks good though, thank you :).

PR #312, opened 2017-03-11
Disclosure: Fixed bug with space bar

jongund:


mcking65:
@jongund, looks good. Many thanks for speedy fix!!

PR #313, opened 2017-03-13
Slider: Added tabindex to source code and updated focus styling to include rail

jongund:


mcking65:
@jongund, it looks like you did not update your fork from master before editing the html files. As a result, you were editing old versions of the html files. For one of them, it appears to git you have changed all lines.

You could attempt to merge master into your branch. But, in the current state, that looks harder then simply replacing the html files in your branch with the latest from master and then re-making only the changes that are necessary.

jongund:
Submitted a new pull request, hopefully works as expected this time

PR #314, opened 2017-03-13
Updated code generation script 

jongund:
Updated code generation script to:
1. Exclude container element with id
2. Added the inclusion of comments in source code
3. Fixed bug of HTML code in script elements not being included


mcking65:
@jongund, thank you!!

PR #315, opened 2017-03-13
Disclosure: Updated alt text of image, text content of figure caption

jongund:
I also fixed the problem with the figcaption tag, it was in the document as caption


mcking65:
Looking at 
[spec for figcaption](https://www.w3.org/TR/html5/grouping-content.html#the-figcaption-element),
the `figcaption` has to be the first or last element inside the figure.

jnurthen:
I think the center align on the caption makes the text really hard to read. Any objection to left aligning it?

jnurthen:
Looking at the spec for figcaption it states that it can include flow elements or text.  Button is not a flow element so I'm not sure this example is valid. Of course we could put the button inside a div and that would be valid (I think) but I think we need a change.

mcking65:
@jnurthen, the [spec for button](https://www.w3.org/TR/2011/WD-html5-20110525/the-button-element.html) 
says it is in the flow content category and the 
[spec for flow content](https://www.w3.org/TR/2011/WD-html5-20110525/content-models.html#flow-content-0) 
lists button. So, I think we are OK with a button inside of a figcaption.

jnurthen:
You are right. Serves me right for looking at an inaccurate document.... http://w3c.github.io/html-reference/common-models.html#common.elem.flow

jongund:
Matt I have made the changes and it looks like this branch is clean for a merge

mcking65:
@jongund, yes, this pull does indeed make the needed changes and tests clean. Thanks! I am merging.

PR #318, opened 2017-03-15
Updated button to fix bug in source code rendering

jongund:


mcking65:
@jongund, when looking at the source, I personally find the info that says which is action and which is toggle useful. It was treating the set of two buttons as one example block rather than two separate examples.

To me, the state without this change is more of a feature than a bug. 

What is the problem the this PR resolves?

jongund:
I was trying to fix the problem of source code not rendering, but it seems you have done this so I didn't need to do.   So if your happy with your fix, you can disregard mine.

mcking65:
@jongund, thanks. closing then.

PR #319, opened 2017-03-15
Checkbox Example (Mixed State): Hide bullets in checkbox list

jongund:


mcking65:
@jongund, since the review process for the checkbox examples is closed and we did not have a bug raised on this, I want to make sure we have multiple editors review before accepting.

@jnurthen, this looks good to me. Please comment with your take.

PR #320, opened 2017-03-15
Updated slider to include tab index in source code 

jongund:
Also improves focus styling of rail


mcking65:
Thank you @jongund, this one looks good.

I added commit a12f281  to make a modification to the roles, states, and properties table. Previously, I had added a bullet about the tabindex being dynamically added; I removed that.

This is now ready to merge.

Issue #328, opened 2017-03-17
Update table Style in example template


jnurthen:
Group decided to add the full grid lines based on decision at https://www.w3.org/2017/03/13-aria-apg-minutes.html

MichielBijl:
The CSS can be simplified to:

```
table.attributes tbody th,
table.attributes tbody td {
    border: 1px sold silver;
}
```

There is nothing in the base.css (W3C default style) that prevents borders on specific elements. We only care about tables with the `.attributes` class; this is the class applied to all role / attribute tables.

mcking65:
@MichielBijl, so, I will merge pull #329 since it has the desired effect. However, we will leave this issue open so that either you or James can take care of the implementation detail of simplifying the CSS. My understanding from today's meeting is that you would make this change. So, assigning this issue to you.

MichielBijl:
Fixed in https://github.com/w3c/aria-practices/commit/e389a15c91c0d2f529cfd35015b108bd3ec99159

PR #329, opened 2017-03-17
Added Grid Lines for the Role, Property, State, and Tabindex Attributes Tables

jnurthen:
In the example documents added grid lines for the Role, Property, State, and Tabindex Attributes Tables


jnurthen:
To resolve #328

Issue #189, opened 2016-12-07
Updates needed for button example page


mcking65:
The [button example page](http://w3c.github.io/aria-practices/examples/button/button.html) 
needs the following changes.

- [x] Update to use the current [example template](https://github.com/w3c/aria-practices/blob/master/examples/coding-template/Example-Template.html)
- [x] Make the toggle button with anchor instead of span.
- [x] Drop the 3rd "show alert" button example.
- [x] Remove the "non-compliant" example or change it to follow the pattern; the task force has agreed to not include anti-patterns.
- [x] Change the title to "Button Examples"
- [x] Fix misspellings.
- [x] Task force decision on whether a note discouraging use of role button in favor of html buttons is warranted. If yes, agree on a wording.

annabbott:
If the task force has agreed to not include anti-patterns, then the non-compliant pattern "a element with href="#"" should be deleted.

mcking65:
I am not comfortable with the wording of the following note so I am removing it for now, and I am adding a task to the above checklist to address this point.

>Note: When using HTML, use the button element. 
>It is recommended that authors not re-purpose other elements to create buttons. 

If we are going to have a note like this, the note should either include a crisp statement of rationale or link to an explanation.

jongund:
Does the mute button need to change the accessible when it toggles from pressed to unpressed (e.g. mute to unmute)?

mcking65:
@jongund wrote:
>Does the mute button need to change the accessible when it toggles from pressed to unpressed (e.g. mute to unmute)?

No, in fact, the design pattern explicitly warns people not to do that. We may want to remind of them of that on the example page in the properties table.

mdahamiwal:
The first button example "Print Page" doesn't specify `tabindex` and is not part of tab order.

mdahamiwal:
@mcking65, can you review the fix?

jnurthen:
@mdahamiwal thanks.

mcking65:
1. Make the toggle button with anchor instead of span.
2. Drop the 3rd "show alert" button example.

jnurthen:
Resolved in 3/13/2017 meeting not to add a note discouraging button role

mcking65:
@jongund, while working on editorial clean up, I noticed that the svg for the toggle is inline but not part of the example div. Since it is a part of the example, it should either be inside that div, which might be distracting to people reading the html, or in an svg file. I sort of prefer the latter for both clarity and consistency with other examples.

[...]

PR #333, opened 2017-03-20
Button: Moved SVG for mute button to an external file

jongund:


mcking65:
Thank you @jongund. This fixes the issue I pointed out in issue #189. 

I pushed one additional commit because one of the parameters of the html source display script was wrong; id of the example div on this page is "example" instead of "ex1". I am not sure why that change was not merged in from master; perhaps it was a conflict and you deleted the way that it was in master when resolving. At any rate, it is working now.

PR #343, opened 2017-03-23
added polyfill for remove and checks for focus to fix IE11 issues

tatermelon:


mcking65:
@tatermelon, I pushed one more commit to fix a jscs error in utils.js. Merging now. thanks!

PR #356, opened 2017-04-10
Update menubar-2.html

jongund:
Should fix problem with both keyboard and menu width issue


mcking65:
Jon, I don't think this PR contains what you expected. It has just one immaterial change to the html file.

PR #368, opened 2017-04-20
Treegrid

aleventhal:


mcking65:
This is pretty cool!!

Duplicating aria-expanded seems confusing from a screen reader perspective. 
So, if rows are focusable, I think it would be better if aria-expanded is on the row only. And if rows are not focusable, it should be on the first cell.

in the third model where rows are not focusable, aria-expanded on the cell is working great! Now, we have a very flexible set of options!

PR #374, opened 2017-04-24
Fixed left arrow bug that caused the tree leaf to collapse when focus moved to parent treeitem and modifier key problem

jongund:


mcking65:
Thank you @jongund!!

PR #387, opened 2017-04-27
removed from Javascript setting a tabindex=0

jongund:
I fixed a bug of the Javascript adding a tabindex=0, this is not needed since the button is already part of the tab order of the page


mcking65:
Looks good; merging.

PR #390, opened 2017-04-29
Treeview: Fixed bug with right arrow not going to first child treeitem when on a expandable treeitem

jongund:


mcking65:
@jongund, working great!

Issue #158, opened 2016-11-14
Develop accordion example that includes support for the optional keyboard commands


mcking65:
#### Objectives

Develop an additional accordion example page and enhance the accordion js to:

- [ ] Support the design pattern's optional keyboard commands (listed below).
- [ ] Implement accordion controls with HTML button elements as recommended in issue #210 by @aruns07 
- [ ] Demonstrate a panel with a nested accordion.
- [ ] Demonstrate panels with headings in the content.
- [ ] Demonstrate headers with non-text content, e.g., thumbnails of images.
- [ ] demonstrate action menus for accordion headers.

#### Optional Accordion Keyboard Commands are

1. Down Arrow (Optional): If focus is on an accordion header, moves focus to the next accordion header. If focus is on the last accordion header, either does nothing or moves focus to the first accordion header. 
2. Up Arrow (Optional): If focus is on an accordion header, moves focus to the previous accordion header. If focus is on the first accordion header, either does nothing or moves focus to the last accordion header. 
3. Home (Optional): When focus is on an accordion header, moves focus to the first accordion header. 
4. End (Optional): When focus is on an accordion header, moves focus to the last accordion header. 
5. Control + Page Down (Optional): If focus is inside an accordion panel or on an accordion header, moves focus to the next accordion header. If focus is in the last accordion header or panel, either does nothing or moves focus to the first accordion header. 
6. Control + Page Up (Optional): If focus is inside an accordion panel, moves focus to the header for that panel. If focus is on an accordion header, moves focus to the previous accordion header. If focus is on the first accordion header, either does nothing or moves focus to the last accordion header. 

#### References

1. [Accordion pattern](http://w3c.github.io/aria-practices/#accordion)
2. [First accordion example page](http://w3c.github.io/aria-practices/examples/accordion/accordion1.html)
3. Accordion javascript:
    * [setup.js](https://github.com/w3c/aria-practices/blob/master/examples/accordion/js/setup.js)
    * [helpers.js](https://github.com/w3c/aria-practices/blob/master/examples/accordion/js/helpers.js)

mcking65:
This was completed with pull #397. Closing.

PR #397, opened 2017-05-26
Issue #158 Add Optional Keyboard Commands to Accordion Example

gerardkcohen:
Added the following (optional) keyboard commands based on issue #158:

Up/ Down: Cycle focus through accordion toggle controls/ headings
Home/ End: Focus first/ last accordion toggle control/ heading
Control + Page Up/ Down: Move focus from accordion panel to accordion heading, and then cycle through headings.

Also refactored JS and CSS to be simpler to understand.


mcking65:
@gerardkcohen, thank you! This looks excellent. I really like the refactoring, especially the use of data attribs to control the expand/collapsing behaviors and the css icons.

On windows, Ctrl+pageUp/Down are not working for me; the browser is grabbing them. Also it looks like you are not filtering out modifier keys when not applicable, e.g., ctrl+home/end do same as home/end. Should be ignoring all modifiers on those.

Nonetheless, I don't want to keep this out of the publication we are working on for this week, so I am merging and opening a separate review issue to collect this and other feedback.

PR #398, opened 2017-06-13
removed aria-label from menuitems

jongund:


mcking65:
@jongund, looks good. Thank you!!

PR #399, opened 2017-06-13
Updated treeview examples and document to remove the use of aria-label on treeitems

jongund:


mcking65:
@jongund, looks great! thank you!!

PR #411, opened 2017-06-21
fixed tree view bugs

jongund:
* Removed default keyboard focus styling from treeview widget elements
* Changed the use of character entities to images in treeview-2 examples


mcking65:
@jongund, thank you! This is looking much better!

PR #422, opened 2017-07-14
Change JSCS to ESLint

nschonni:
JSCS project was merged with ESLint a year ago http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint. It's probably good to switch to the actively maintained project.
1. I used `polyjuice` for a basic automated rule conversion, so it might be a little lossy.
1. Used `npm run fix` to do the basic whitespace and other autofixing
1. Manually fixed the remaining 2 errors
1. Used the `node` tag on Travis since it was still running 0.10, which is also EOL


MichielBijl:
@nschonni thanks for this! The errors are pretty random. @michael-n-cooper looked into it at one point, but can’t remember what the outcome of that was. I’ll try to re-run the test and merge when that’s done.

PR #423, opened 2017-07-17
Uncheck a radio

NicoR91:


MichielBijl:
@NicoR91 can you please elaborate on what the purpose of this PR is? From the code I gather that it’ll allow users to uncheck a radio button once it has been checked. This isn’t something that we find on other platforms; you can’t uncheck a radio button on either Windows or macOS for example.

NicoR91:
@MichielBijl Thanks for your response! Yes, I wanted to allow users to uncheck a radio button because I had cases where the response is non-mandatory. And if the user check by error one response, I wanted he could uncheck the radio button.

Probably the solution is to provide another response like « No Response » and in this way, I could satisfy the rule you mentioned and my needs.

Sorry for the inconvenience and thanks for your work and your explanations!

MichielBijl:
No problem, thank you for your contribution 👍

PR #428, opened 2017-07-20
chore: typo in "nvda" aria-labelledby value

nschonni:
The typo was causing the labelledby to point at the wrong/missing ID

### DIFF
--- a/examples/landmarks/at.html
+++ b/examples/landmarks/at.html
@@ -94,7 +94,7 @@
             
             </section>
 
-            <section aria-labelledby="ndva">
+            <section aria-labelledby="nvda">
             <h2 id="nvda">NVDA Screen Reader for Windows</h2>
                <p>The following commands are available in NVDA screen reader (since version 2014.2) for navigating landmarks:</p>
 


examples/landmarks/at.html
```diff
@@ -281,4 +281,4 @@ <h2 id="id3">Related W3C Documents</h2>
     </script>
     <script type="text/javascript" src="js/SkipTo.min.js"></script>      
   </body>
-</html>
```
jnurthen: Can you please use an editor that supports .editorconfig so we get consistent formatting. Files should end in a newline

nschonni: I was using the web editor, which does actually support EditorConfig.
This diff is actually saying that I fixed the missing eof newline

jnurthen: ok... I get it now.... That is confusing of github to add a symbol for a missing newline character ;)

nschonni: Yup, that character they use is pretty odd. I resisted commenting the same thing on your last PR 😝

Issue #427, opened 2017-07-20
Landmark AT Example page using percentage based image widths


nschonni:
Found this running the Nu validator https://travis-ci.org/w3c/aria-practices/builds/253487231#L418
Looks like those widths need to be in pixels in HTML5 https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#attr-width

MichielBijl:
Fixed in https://github.com/w3c/aria-practices/commit/e3547be1243455697e6dfac62a21a96112e97d5e

nschonni:
Technically the PR hasn't landed, but the changes look 💯

MichielBijl:
Discussed the change I made yesterday with @jnurthen and he approves.

PR #431, opened 2017-07-20
Change width attribute to inline style

jnurthen:
#427


MichielBijl:
I actually removed the attributes and put a rule in the CSS to limit images to have a max width of 100%.

PR #433, opened 2017-07-20
chore: use Travis Github Pages deployment

nschonni:
They have a built in function for this now https://docs.travis-ci.com/user/deployment/pages/.
Needs some more testing because the respec generation is currently failing on Travis.


jnurthen:
I think the respec generation simply times out. Builds normally complete if you just restart them. Can you try adding something from here  https://docs.travis-ci.com/user/common-build-problems/#Timeouts-installing-dependencies to see if it resolves the issue?

nschonni:
Add 3 retries which seemed to work around it.
Noticed one thing that is different with this approach. It keeps the `gh-pages` as an orphan branch with a single commit instead of keeping any history. Not sure if anyone sees any issue with that since it is the generated content.
I also didn't try to cleanup an of the other folders like `bin` and the dotfiles either

MichielBijl:
I don’t think we need the history for the gh-pages branch.

nschonni:
Should I add some `rm -rf` for the other files, or is it ok as-is?
/cc @michael-n-cooper

michael-n-cooper:
Hmm, I'm not sure what this approach gains us - it copies files over, but the previous version of the script was already doing that. It would wind up copying aria-practices.html over, which isn't needed in gh-pages (albeit not harmful). And while in theory it's probably not a big deal to blow away the history of gh-pages, in practice it's useful sometimes to see the history, how often snapshots were posted, etc. It might also be easier to get meaningful diffs between old commits on that branch than on the master branch. So on the whole, I'm not seeing value in using this new feature, but if you all prefer it, go ahead.

The adding 3 retries to the curl command could be very helpful, I didn't know that was a feature. I would certainly keep that part of the change if you decide not to do the rest.

.travis.yml
```diff
@@ -3,34 +3,22 @@ language: node_js
...
   - "node"
 
-env:
-  global:
-    - GH_REF: github.com/w3c/aria-practices.git
-
 git:
   depth: 3
 
 branches:
```
nschonni: Just realized with this approach, this filter can be removed so PRs outside `master` can be built. I was scratching my head to why 440 wasn't running :stuck_out_tongue:

nschonni:
Closing in favour of #445

Issue #425, opened 2017-07-17
Accordion bug within addEventListener target


be-codified:
Just noticed, when using Chrome and clicking on panel's title, target is:

`<span class="Accordion-title">Personal Information</span>`

instead of: 

```
<button aria-expanded="false" class="Accordion-trigger" aria-controls="sect1" id="accordion1id">
    <span class="Accordion-title">Personal Information</span><span class="Accordion-icon"></span>
 </button>
```
This will fix it:

```
accordion.addEventListener('click', (event) => {
    const target = event.target.classList.contains('Accordion-trigger') ? event.target : 
    event.target.parentNode;
    ...
```

Well played in Firefox and Safari though.

gerardkcohen:
@be-codified I have a fix coming soon.

mcking65:
Fixed with commit 74887d9. Many thanks to @gerardkcohen!

Issue #424, opened 2017-07-17
Accordion data-allow-multiple & data-allow-toggle not working


be-codified:
Hi,

I am trying to use 

```
<dl id="accordionGroup" role="presentation" class="Accordion" data-allow-multiple data-allow-toggle>
```

but nothing happens. Can you please check?

nschonni:
Are you running on an older browser? I found in https://github.com/w3c/aria-practices/pull/422/commits/88c142d48dad51220357104e90c661d14b913889 that the example was using some ES6 stuff

be-codified:
No, I am using latest Chrome or Firefox. @nschonni is script working for you?

gerardkcohen:
@be-codified  @nschonni I have a fix coming soon.

mcking65:
Closed with commit 74887d9.

PR #434, opened 2017-07-22
Accordion fixes

gerardkcohen:
This commit addresses issues #424 and #425.

Note: I needed `data-allow-multiple` to also mean `data-allow-toggle` otherwise you could end up in a situation where you can open all accordions but not close them.

### DIFF
--- a/examples/accordion/accordion.html
+++ b/examples/accordion/accordion.html
@@ -41,11 +41,13 @@
             Allow for each toggle to both open and close individually
 
           data-allow-multiple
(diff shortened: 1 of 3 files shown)


gerardkcohen:
Ping @mcking65

mcking65:
Sorry @gerardkcohen and others for the very long delay on this. I am back in the game full force now. 

Everything looks good with one exception, but I will merge anyway and raise a separate issue. The issue is that when we do not have data-allow-toggle, we need aria-disabled on the expanded header.

PR #435, opened 2017-07-27
Navigation Menubar Example: Remove second example with dynamically added ARIA attributes

jongund:
This bug fixes a problem in the source code view and I don't think we needed a label for the role=menu or a tabindex=-1 on the menu role


mcking65:
@jongund, I will put this on agenda for Monday. 

We want labels on the menu elements, but it looks like they are in the HTML source so the js you are removing does nothing. Example 2 states that the ARIA is not in the HTML, but it actually is. I am thinking we should consider removing the second menubar example that is on the same page ... it may be more distracting than valuable ... and permits simplification of the js.

mcking65:
@jongund, thank you, thank you! Much cleaner. Note: I pushed a commit to your fork that removed a few more vestages of the second example.

PR #448, opened 2017-08-28
Fixed some behavior bugs and added documentation for aria-disabled on font smaller and larger; also fixed the aria-expanded bug with link menubar and second level popups

jongund:
fixed bugs in:
1. Italic style menu button did not toggle off italics and now it does
2. Activating the font larger or smaller menu item was not updating the list of menubuttonradio items for font size and now there are updated
3. Updated documentation to include information about the use of aria-disabled for the font smaller and larger options

### DIFF
--- a/examples/menubar/menubar-1/js/PopupMenuLinks.js
+++ b/examples/menubar/menubar-1/js/PopupMenuLinks.js
@@ -241,7 +241,7 @@ PopupMenu.prototype.open = function () {
     this.domNode.style.zIndex = 100;
   }
 
-  this.controller.domNode.setAttribute('aria-expanded', 'true');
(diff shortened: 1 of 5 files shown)


mcking65:
@jongund, thank you. 

There were some ESLint errors that I was able to fix automatically so I pushed another commit to your menubar branch.

I am going to squash and merge this so please delete your menubar branch and make a new branch off the latest master for any additional pull requests related to menubar.

The first set of changes look good. However, at least as far as I can tell with JAWS in Firefox 56, issue #472 is not yet fixed. It looks like the parent menuitem for Facts and Campus Tours still have aria-expanded=false even when the submenus are expanded.

Issue #401, opened 2017-06-18
Review accordion example updated with implementation of optional keyboard interactions


mcking65:
@gerardkcohen refactored the 
[accordion example](http://w3c.github.io/aria-practices/examples/accordion/accordion.html) 
to simplify the code and add support for the optional keyboard interactions, including Up Arrow, Down Arrow, ctrl-PageUp, ctrl-PageDown, Home, and End.

This issue is to capture feedback on the updated example.

DavideTriso:
The implementation actually does not prevent all accordions to be collapsed at the same time, as stated in the description. Click on the expanded accordion collapses it.

Decrepidos:
A couple of thought:

**Accordion triggering element (Button)**
WAI-ARIA Roles, States, and Properties

**In response to:** If the button is a toggle button, it has an aria-pressed state. When the button is toggled on, the value of this state is true, and when toggled off, the state is false.

**Reason:** aria-pressed is not included in the accordion design pattern.

I suggest the following be added to the spec for clarity.

_The difference between a button that toggles and does nothing else, like Enable Cookies, where the value is true or false, as opposed to a button that expands or collapses another visible section on the page. The first would be best suited to use aria-pressed, whereas the second is best suited to use aria-expanded. Therefore, if a toggle button contains a mechanism to convey the state of the button to the user i.e. aria-expanded, aria-pressed can be omitted._

carmacleod:
The accordion example has odd mouse behavior, in that clicking on either the Accordion-title or Accordion-icon span inside the Accordion-trigger button do not trigger the open/close action.
Unfortunately, the title text or the drop-down arrow icon are the 2 most likely places that a mouse user will click to open/close the panel. (This behavior seen on Windows Chrome).

Decrepidos:
@carmacleod

The only way I can work around this is to use a fake button.

<dt role="button" aria-expanded="true" class="Accordion-trigger" aria-controls="sect1" id="accordion1id" tabindex="0">
    <h3 class="Accordion-title">Personal Information</h3>
</dt>

Then for the toggle icons use
#accordionGroup .Accordion-trigger.open:before {
    font-family: FontAwesome;
    content: "\f077";
    float: right;
    color: #fff;
}
#accordionGroup .Accordion-trigger:before {
    font-family: FontAwesome;
    content: "\f078";
    float: right;
    color: #fff;
}

[...]

PR #466, opened 2017-09-24
#401 IE11 Compatible JS

gerardkcohen:
Removed ES6 JS and make compatible with IE11, as mentioned in #401


mcking65:
Thank you @gerardkcohen!

PR #473, opened 2017-10-10
Styled buttons and toolbar in listbox example

tatermelon:
Updated style of toolbar and buttons 

screen shot 2017-10-10 at 2 10 45 pm


mcking65:
Thank you @tatermelon!

Issue #475, opened 2017-10-11
Accordion Example: Roles, States, and Properties table incorrectly says the header control uses an `<a>` element


pudgereyem:
> **DISCLAIMER**  
> This is my first contact with this repository. Please correct me if I'm missunderstanding stuff. :)
> It would be nice to contribute, and this seems like a good start.

## Issue
The [Accordion Example](https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html) uses `<button/>`, but the documentation tells us that it uses an `<a/>` (anchor) element

## Solution
Change the Documentation in https://github.com/w3c/aria-practices/blob/master/examples/accordion/accordion.html to use `<button>` instead of `<a>`.

> I would like to open up a PR for this, but please answer to this issue if you'd like.

## Background / Discussion
I looked through the activity in this github project to understand the current status. Below is what I found. I wrote this down for myself to get a bit of context.

- Discussion about wether or not to use `<button>` or `<a>` element.
- 2017-02-16: @mcking65 writes in https://github.com/w3c/aria-practices/issues/210#issuecomment-280441653  that `<a>` element should be used
- Issue https://github.com/w3c/aria-practices/issues/158 that has todo that `<button>` element should be used
- 2017-06-17: Pull Request Merged https://github.com/w3c/aria-practices/pull/397 that now uses `<button>` component

mcking65:
@pudgereyem commented:

> This is my first contact with this repository. Please correct me if I'm missunderstanding stuff. :)
> It would be nice to contribute, and this seems like a good start.

Victor, thank you for both catching this error and offering to help!!

I haven't put this in the readme yet, but it is a good idea to raise an issue or comment in an issue, as you did, before going through the work of changing code and submitting a PR. That will help ensure you aren't working on something that someone else is already in the process of changing. It can also make sure that your fix is aligned with the spec, task force objectives, etc.

That said, if you see anything in our [current milestone](https://github.com/w3c/aria-practices/milestone/6) 
that might interest you, please comment in the issue and say so!

> The Accordion Example uses  <button/> , but the documentation tells us that it uses an  <a/>  (anchor) element.

Excellent catch!

pudgereyem:
Hey @mcking65, thanks for the warm welcome!

I will definitely check out https://github.com/w3c/aria-practices/milestone/6, and see if I can be of any help. I'm not that "experienced" with a11y, so contributing to this project would be a great learning experience for me.

Cheers!

[...]

PR #477, opened 2017-10-11
Accordion Example: Update documentation for change to button from link

pudgereyem:
Fixes issue https://github.com/w3c/aria-practices/issues/475

ping @mcking65

### DIFF
--- a/examples/accordion/accordion.html
+++ b/examples/accordion/accordion.html
@@ -271,14 +271,6 @@
           </td>
         </tr>
         <tr>
-          <th scope="row"><code>button</code></th>
-          <td></td>
-          <td><code>button</code></td>
-          <td>
-            Button inside the accordion header that provides hide/show functionality.
-          </td>
-        </tr>
-        <tr>
           <td> </td>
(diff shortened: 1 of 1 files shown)


examples/accordion/accordion.html
```diff
@@ -273,31 +273,31 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
         <tr>
           <th scope="row"><code>button</code></th>
           <td></td>
-          <td><code>a</code></td>
+          <td><code>button</code></td>
           <td>
-            Identifies the element as a button inside the accordion header that provides hide/show functionality.
+            Button inside the accordion header that provides hide/show functionality.
```
mcking65: Actually, we need to delete this row of the table because we do not apply the role button to the button element. This table lists only explicit uses of ARIA, not implicit roles.

pudgereyem: Ahh, I actually thought about that. But wasn't sure. Fixed this in commit b35d3a1

mcking65:
Looks great now, Thanks!!

PR #478, opened 2017-10-11
Menu button has been updated to not use SPACE key to activate popup menu items

jongund:


mcking65:
@jongund, thank you! This is looking good. I see just one issue. The fix seems to apply to two of the examples but space bar is still activating menu items in menu-button-actions-active-descendant.html.

mcking65:
@jongund, thank you!!! So glad to have this one knocked off the list! All looks good.

As usual, since this branch has multiple commits with messages that are not necessarily helpful in our history, I am squashing and merging. So, after I merge, please delete the menu-button branch from your fork.

PR #479, opened 2017-10-11
Update keyboard shortcuts for listbox

tatermelon:
- Updated Add and Important buttons to be triggered by Enter instead of Delete
- Added Alt up/down shortcuts for shifting items in Important list

### DIFF
--- a/examples/listbox/js/listbox-rearrangeable.js
+++ b/examples/listbox/js/listbox-rearrangeable.js
@@ -14,15 +14,15 @@ window.addEventListener('load', function () {
     document.getElementById('ex1-up'),
     document.getElementById('ex1-down')
   );
-  ex1ImportantListbox.setupDelete(document.getElementById('ex1-delete'), ex1UnimportantListbox);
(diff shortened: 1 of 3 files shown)


mcking65:
@tatermelon, 

I am experiencing a strange bug. I tested in Firefox. Here are repro steps:

1. Tab to available upgrades listbox.
2. Press End key.
3. Press space to select, up arrow and press select to select 2nd to last item as well.
4. Press enter to add.

Result: Focus jumps to chosen upgrades. Pressing back-tab twice returns focus to chosen upgrades. Can not get to available upgrades.

Expected result: The two selected items are added to chosen upgrades and focus remains in available upgrades. 

The bug occurs if two or more consecutive items are selected and one of the selected items is the last item.

mcking65:
@tatermelon, one other question ... are there any tooltips or anything like that on the toolbar buttons that can reveal the key shortcuts to sighted users?

mcking65:
It's working great now!! Merging.

PR #483, opened 2017-10-16
Add expandable listbox example

tatermelon:


mcking65:
@tatermelon, thanks! I am having one issue. Enter is not choosing an option. I can choose by tabbing out. But, if I expand, arrow to an option and press enter, nothing seems to happen in response to the enter key. I tried in both Firefox and Chrome.

PR #485, opened 2017-10-17
Live arranging

tatermelon:


mcking65:
@tatermelon, I merged master in and resolved conflicts in listbox-rearrangeable.js. However, that broke both the keyboard shortcuts and the live regions! It is not obvious to me what is wrong. I pushed the merge commit to your branch. Please do a git pull and take a look.

Issue #486, opened 2017-10-18
Mixed state checkbox example does not correctly recall previous mixed state in some scenarios


mcking65:
The [mixed state checkbox example](http://w3c.github.io/aria-practices/examples/checkbox/checkbox-2/checkbox-2.html) does not correctly recall the last mixed state of checkboxes if the mixed state was created by changing the state of controlled checkboxes from the all checked or no checked states.

The code should record the set of checkbox states any time the master checkbox is in the mixed state, regardless of what event caused it to get to that state. Then, restore that state any time the master checkbox is in the unchecked state and it is activated to go back to the mixed state.

Example tests 1:
1. Load page.
2. Check master all so all boxes are checked.
3. Uncheck tomato so master is in mixed state.
4. check master to return to all checked state.
5. Check master 2 more times; should return to state in step 3.

Example test 2:
1. Load page.
2. Check master all so all boxes are checked.
3. Check master so no boxes are checked.
3. Check one or more boxes so master is in mixed state.
4. check master to return to all checked state.
5. Check master 2 more times to return to mixed state; should return to states in step 3.

Issue reported by @racer2207 and @annabbott in issue #471.

mcking65:
This is fixed with ca272bf and thanks to @gerardkcohen!

PR #491, opened 2017-10-23
Fix issue 486: Update mixed state on change

gerardkcohen:
Fix issue #486.

The selected state of each individual checkbox was only recorded on click of that individual checkbox. In order to properly track the state of all the checkboxes a routine is added to update the state of all controlled checkboxes when the master is set to mixed state, which happens every time a change occurs in the controlled checkboxes.

I would have preferred a cleaner approach where the controller stored the mixed states, but did not want to be too destructive with my changes.

cc: @mcking65


mcking65:
Yay, works excellently!! Thank you @gerardkcohen!

PR #510, opened 2017-10-31
Fixed bug with TAB key not closing menus or tabbing out of menubar

jongund:


mcking65:
Excellent, seems to be working correctly now. I fixed a couple of ESLint errors with the auto fix, and it's now good to merge.

Thank you @jongund!

PR #518, opened 2017-11-03
Copyright and comments

jnurthen:
- Added W3C Copyright to all the js files (except those in the JS directory which have other licenses and are only referenced)
- Removed all Author declarations in the JS files
- Added Removed authors to the Other Contributors section of acknowledgements.


mcking65:
@jnurthen, looks good! Thank you for this!

Issue #524, opened 2017-11-16
Editor Menubar Example: Enter not closing menus when performing a command and aria-checked not correctly set on some items


mcking65:
This issue is for two bugs in the [editor menubar example](http://w3c.github.io/aria-practices/examples/menubar/menubar-2/menubar-2.html):

1. When Enter is pressed to activate a menuitem that performs an action, e.g., change font, it should both perform the action and close the menu. In this particular example, the focus should logicaly go to the text area after the action is performed.

2. When menuitem radios and menuitemcheckbox items are activated, sometimes the aria-checked state is not set to true. And, if a menuitemradio, the previously checked item does not get aria-checked set to false. This was working correctly prior to commit 0de93bd652d86b035697b554adf4ae088af20661.

mcking65:
With thanks to @sh0ji, this is resolved with commit ba84ee6.

PR #525, opened 2017-11-17
Menubar2 fixes

### LINKED ISSUE #521: Menuitems duplicated in Editor Menubar Example
In the Editor Menubar Example, the menuitem elements are duplicated.

https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-2/menubar-2.html

`<li role="menuitem" aria-haspopup="true" aria-expanded="false">
  <span tabindex="0" role="menuitem" aria-haspopup="true" aria-expanded="false">Font</span>`

I assume LI should only have role="none".

Twitter references:
https://twitter.com/jnurthen/status/930852962582151169
https://twitter.com/jnurthen/status/930872266664984576

sh0ji:
This fixes #521 & #524.

For #521, I placed the attributes directly on the HTML and removed the JavaScript that was adding various attributes (part of `MenubarItemAction.init()`).

#524 was relatively straight-forward, but the first point did seem to be deliberately coded into the example.

Note that I also changed the `aria-haspopup` elements to anchors from spans to match the [menubar-1](http://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html) pattern. My .editorconfig also appears to have removed some unnecessary whitespace from the HTML.

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -1934,8 +1934,19 @@
           <li>Items in a menu may be divided into groups by placing an element with a role of <a href="#separator" class="role-reference">separator</a> between groups. For example, this technique should be used when a menu contains a set of <a href="#menuitemradio" class="role-reference">menuitemradio</a> items. </li>
           <li>All <a href="#separator" class="role-reference">separators</a> should have <a href="#aria-orientation" class="property-reference">aria-orientation</a> consistent with the separator's orientation. </li>
           <li>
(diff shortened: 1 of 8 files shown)


sh0ji:
Just noticed that 34894e8985729103e20a1647d3297b8b1066b5be was what added the "toggle but don't close" behavior, which describes that space should do this, but enter should still close. That should now be working.

mcking65:
@sh0ji, Evan, looking great! Many thanks!!

PR #528, opened 2017-11-18
Radio Group with Roving tabindex Example: Clean up example HTML and correct supporting documentation

kleinfreund:
I noticed some grammar errors in this example and tried to correct them. The indentation of code was also fixed.

When I find the time, I try to have a more focused look at the example (and potentially others) and see whether they need any care.

### DIFF
--- a/examples/radio/radio-1/radio-1.html
+++ b/examples/radio/radio-1/radio-1.html
@@ -14,238 +14,226 @@
     <link rel="stylesheet" href="../css/radio.css">
     <script src="js/radioGroup.js" type="text/javascript"></script>
     <script src="js/radioButton.js" type="text/javascript"></script>
-
   </head>
   <body>
     <main role="main">
-      <h1>Radio Button Example</h1>
+      <h1>Radio Group Example</h1>
 
-     <p>This example implements the features of the <a href="../../../#radiobutton">Radio Button Design Pattern</a> for selecting a pizza crust and delivery method.</p>
+      <p>This example implements the features of the <a href="../../../#radiobutton">Radio Button Design Pattern</a> for selecting a pizza crust and delivery method.</p>
 
       <p>Similar examples include: </p>
+      
       <ul>
         <li><a href="../radio-2/radio-2.html">Radio Button Example (aria-activedescendant)</a>: Radio button group using <code>aria-activedescendant</code> for managing keyboard focus.</li>
         <!--  list other examples that implement the same design pattern. -->
       </ul>
 
 
-     <section aria-labelledby="example_label">
-      <h2 id="example_label">Example Start</h2>
-      <div id="ex1">
+      <section aria-labelledby="example_label">
+        <h2 id="example_label">Example</h2>
 
-        <div  role="radiogroup" aria-labelledby="group_label_1" id="rg1">
+        <div id="ex1">
+          <div  role="radiogroup" aria-labelledby="group_label_1" id="rg1">
             <h3 id="group_label_1">Pizza Crust</h3>
 
-            <div role="radio"
-                aria-checked="false"
-                tabindex="0">
+            <div role="radio" aria-checked="false" tabindex="0">
               Regular crust
-           </div>
-           <div role="radio"
-                aria-checked="false"
-                tabindex="-1">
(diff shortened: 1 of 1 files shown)


mcking65:
Phillipp,

Thank you for the contribution. You caught quite a few problems.

While reviewing, I found quite a few other editorial problems I need to address on this page. It looks like this one somehow slipped through the editorial review process; lots of mistakes. Thank you very much for calling this to our attention.

Before I go further with this PR, I am checking if there is a reason to use the entity code for double quotes in our code snippets. I know that it is not technically necessary, but I want to make sure the spec document generation tools we use do not require it for some reason.

BTW, we have not been indenting the head and body elements. It's common to leave them at the left margin to reduce the overall indent.

Since we are about to make a few global changes across many files, it would be a good idea to raise an issue or comment on an existing issue before going through the work of editing and generating a PR. That can help avoid conflicts that are excessively time consuming to resolve.

It would be wonderful if you can contribute more. As you can easily see from our issues and milestones, we have a lot of work!

kleinfreund:
I tried to keep this contribution in a managable state. It’s supposed to be a first, coarse step. Let me know when this is ready to move forward. Thank you.

pudgereyem:
Hey @mcking65 and @kleinfreund,
I'm just about to start implementing a Radio Group in our UI library.

When looking at the WAI-ARIA Authoring Practices 1.1 section about [Radio Group](https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton) that has two examples:

1.  [Radio group example using roving tabindex](https://www.w3.org/TR/wai-aria-practices-1.1/examples/radio/radio-1/radio-1.html) and
2. [Radio group example using managed focus with aria-activedescendant.](https://www.w3.org/TR/wai-aria-practices-1.1/examples/radio/radio-2/radio-2.html) 

I'm wondering which one to use. When testing on my own with VoiceOver on Mac, I think example two is better because it tells me how many Radio-buttons the Radio Group holds in total. But I don't know. Which one do you prefer?

## On a sidenote
On a sidenote, I would very much learn how people actually are using Accessibility programs, such as VoiceOver. So I can evaluate different patterns and examples myself in a better way. If anyone knows of any resources/tutorials please comment.

I've found this course; [Start Building Accessible Web Applications Today](https://egghead.io/courses/start-building-accessible-web-applications-today) from [egghead.io](https://egghead.io/) but would like to find more resources.

---

Thanks!

kleinfreund:
@pudgereyem I’m not yet used to screen reader programs, so I didn’t try it myself. However, I think the difference between the first and second example is the usage of an `<ul>` element instead of a `<div>` element. I would imagine screen readers announce the number of elements in a list?

All in all, it would be nice if the two examples, which are implementing the same component, would use the same markup,code, etc. wherever possible, only differing in the technique (i.e. roving tabindex vs. activedescendant)

Regarding your question for resources: Not specific to assistive technology itself, but in my opinion absolutely essential for strategies on how to implement accessible components: https://inclusive-components.design by Heydon Pickering.

jongund:
It looks like the keyboard focus styling of the associated accessible names for each checkbox has be removed.
The keyboard focus should also highlight the text associated with the accessible name, especially if radio buttons are going to be horizontal.  The hover should also highlight the accessible name, since it can be used to activate the state of the button.

mcking65:
@jongund, of course, I can't see those issues, but it does not look like they have anything to do with this diff. I will raise a separate issue after merging in these improvements.

mcking65:
@kleinfreund, thank you for this. This is a definite improvement for this page.

pudgereyem:
@kleinfreund Yes, you're right. I would like to know the pros/cons and what solution is preferred. Thanks a ton for the link to https://inclusive-components.design/, great read.

@mcking65 do you have an opinion on which one of the two examples works best?

1.  [Radio group example using roving tabindex](https://www.w3.org/TR/wai-aria-practices-1.1/examples/radio/radio-1/radio-1.html) and
2. [Radio group example using managed focus with aria-activedescendant.](https://www.w3.org/TR/wai-aria-practices-1.1/examples/radio/radio-2/radio-2.html)

PR #530, opened 2017-11-19
Issue 499: Fix LEFT/ RIGHT arrows wrapping to proper row

gerardkcohen:
Small refactor to determine appropriate row by moving logic to a reusable method, which can be used for either UP/DOWN or LEFT/RIGHT arrow keys.


mcking65:
All seems to be working well. Thank you Gerard!

PR #543, opened 2017-11-27
removed title attribute from multi-thumb slider

jongund:


mcking65:
@jongund, there are 2 conflicts. Could you please resolve them?

jongund:
@mcking65 

Fixed conflicts

jongund:
@mcking65 @jnurthen 

Even worse W3C validation problem with ALT attribute, when ALT equal to empty string it says the only role that can be used is "none".  So I will remove the ALT attribute and the less confusing error message.

mcking65:
@jnurthen, @jongund, the [HTML spec for img](https://www.w3.org/TR/html/semantics-embedded-content.html#the-img-element) says:

> Allowed ARIA role attribute values: ‘presentation’ role only, for an `` element whose alt attribute’s value is empty `(alt="")`, otherwise Any role value.
>
> Allowed ARIA state and property attributes:
> * Global aria-* attributes
> * Any aria-* attributes applicable to the allowed roles.

So, seems to me the validator has a bug if it requires `alt` on an `img` that has a role other than `img`. Would you agree?

mcking65:
Jon, the changes look good. I made one minor editorial change. I am merging now. I recommend deleting this branch from your fork because I am squashing all the commits together.

jongund:
Matt,
I agree if the IMG element is no longer has the role=img it should not require alt attribute, but you will be fighting the WCAG group on that one.
ALT on IMG element seems to be a sacred cow in the WCAG working group, they do not seem to want to consider any other way to provide an accessible name for an image.

Jon

Issue #540, opened 2017-11-26
Toolbar Example: aria-selected should not be used.


mcking65:
In the [toolbar example](http://w3c.github.io/aria-practices/examples/toolbar/toolbar.html),
`aria-selected` is being applied to the elements in the toolbar.
This should not be done. The elements in the toolbar are focusable, but not selectable.

Remove aria-selected from the static HTML and the code that sets it from the javascript.

mcking65:
@gerardkcohen  has volunteered to submit a PR for this issue.

mcking65:
Fixed with commit 0fe1784 with thanks to @@gerardkcohen!

PR #546, opened 2017-11-28
Remove use of aria-selected

gerardkcohen:
As mentioned in #540, removed use of `aria-selected`


mcking65:
Gerard, thanks for squashing this! Looks good.

PR #548, opened 2017-11-28
Fix Grid Layout Instructions

### LINKED ISSUE #531: Layout Grid Examples: Keyboard Tutorial NUX does not close when using IE 11.
@shirsha [commented in issue 227](https://github.com/w3c/aria-practices/issues/227#issuecomment-340527273)
that when using IE 11, the keyboard NUX does not close with either keyboard or mouse.

We might want to consider using a button element instead of `<a`.

gerardkcohen:
Fix #531 issue with NUX instructions not closing as expected with IE11.

A loop was created on focus, where the instructions where displayed and then dismissing the instructions set focus to the element that displayed the instructions on focus. This happened because the `once` option was used for the event listener, which is not supported in IE11.

Created an event listener that, when fired, removed the event listener and setup the instructions and close listeners/ handlers.

Tested with IE11 on Win7 and Win8, as well as Chrome/ Safari/ Firefox on Safari.

### DIFF
--- a/examples/grid/js/layoutGrids.js
+++ b/examples/grid/js/layoutGrids.js
@@ -29,31 +29,24 @@ window.addEventListener('load', function () {
 
   var gridNUX = document.getElementById('grid-nux');
   var firstGridCell = document.querySelector('#ex1 [tabindex="0"]');
-  firstGridCell.addEventListener(
-    'focus',
-    function () {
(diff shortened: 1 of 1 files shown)


mcking65:
@gerardkcohen , looks great! Many thanks! And, thank you for the comment explaining the problem.

PR #549, opened 2017-11-28
Fix Dialog Overlays

### LINKED ISSUE #523: Modal Dialog Example: Dialogs underneath other dialogs are not grayed out
In the [modal dialog example](https://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html),
when a dialog appears on top of another dialog, the lower z dialog is not grayed out and is not inert.
For example,  open the "add delivery address" dialog. Then activate the "Verify Address" button. Then, activate the "accepting an alternative form" button, which opens the "end of the road" dialog. When the "end of the road" dialog is displayed, the underlying content from the "verify address" dialog is not grayed out and is still active.

The lower z dialog should always be grayed when another modal is opened on top of it. And clicking outside the currently active dialog should either do nothing or cancel the currently active dialog.

This was raised by @annabbott in a [comment on issue 503](https://github.com/w3c/aria-practices/issues/503#issuecomment-340482191).

sh0ji:
Fixes #523. This is more substantive than I was expecting it to be and probably needs close review.

It was not really possible to fix the issues outlined in #523 by using just one backdrop element, which was previously the main `#dialog_layer` container. Instead, it's now wrapping each dialog in its own backdrop container on open, and then unwrapping them on close. This functions similarly to the native `<dialog>` elements, which each have their own individual [`::backdrop` psuedo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop).

I also cleaned up some of the CSS and replaced the unnecessary `.dialog_default` selector with a `[role="dialog"]` selector.


examples/dialog-modal/js/dialog.js
```diff
@@ -212,14 +217,14 @@ aria.Utils = aria.Utils || {};
     this.dialogNode.className = 'hidden';
     this.focusAfterClosed.focus();
 
+    // unwrap the dialog and remove the backdrop
+    this.backdropNode.parentNode.insertBefore(this.dialogNode, this.backdropNode);
```
mcking65: When the dialog closes and using JAWS or NVDA, focus is no longer going back to the focusAfterClose element. Wondering if order of operations is a problem here.

mcking65:
@sh0ji, did you happen to test on IOS? This is one example where we did a fair amount of work to make it mobile compatible. Most of this was in the CSS. For example, for small screens, the active dialog covers the screen so you don't see the lower z layer scrolling when you scroll the foreground. That was a problem we had to solve for scrolling the long "Verify Address" second layer dialog. So, it would be good to bring it up on a phone and just make sure.

annabbott:
With https://rawgit.com/sh0ji/aria-practices/523-fix/examples/dialog-modal/dialog.html:
When I click with mouse outside "Add Delivery Address" dialog, visible focus does not remain within the dialog. An extra TAB keystroke is required to return visible focus to the dialog. 

Same thing happens in "Verification Result" and "End of the Road" dialogs.

Vastly better than http://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html WRT what each separate dialog displays!

annabbott:
@mcking65 You provided a great explanation for why, in the "Address Added" dialog, the link for "your profile" didn't announce as a link when initially read by screen reader (like aria-describedby is plain text so the link semantic isn't announced). Unfortunately I'm at a loss to find your original explanation...<grrrr>. I think some explanation for this needs to be added to the example - I'm certain I won't be the only one who notices and will question...

sh0ji:
@mcking65 You're right, I had overlooked the fact that it's now using `margin: 10vh auto;` for large screens, but not updating that for small screens. I fixed that in the most recent commit. The scrolling and everything else works as expected on mobile.

sh0ji:
@annabbott If I understand you correctly, when you click outside the modal and then press tab, focus starts at the cancel button; but if you do it again, focus starts on the first form element ("Street" input). It alternates between those every time you click outside the modal.

From what I can tell, this bug has existed a while. It happens on all previous versions of this page that I've tested and in all browsers I've tested it with. I'll look into this as a secondary bug with this PR.

annabbott:
FF 57.0 on Win10

For the "Add Delivery Address" dialog, initial focus is set to "Street" form element which I left empty. Using mouse, click on inert page behind dialog causes "Street" to loose focus and I can't tell where the focus has gone. TAB keystroke returns focus to "Street" form element which I left empty. TAB to "City" form element which I leave empty, click using mouse on inert page behind dialog causes "City" to loose focus and I can't tell where the focus has gone. TAB keystroke returns focus to dialog, but to the "Street" form element. BTW: none of the form elements are indicated as "Required" so I felt leaving empty was a valid test. I was not able to replicate what you indicate as the "Close" button getting focus after clicking on the inert page behind the dialog.

sh0ji:
@annabbott It's my understanding that the behavior you describe is expected. When a user clicks out of a widget, focus usually reinitializes on the next tab keystroke rather than picking up where it left off. If that's not the case, we need to explain that in the design pattern.

sh0ji:
@mcking65 (re: https://github.com/w3c/aria-practices/pull/549#discussion_r153674450) I had some issues verifying the issue described. I observed it on NVDA/IE on my work computer, but not NVDA/Firefox at home.

Still, rather than manipulating the DOM after the backdrop has been inserted, I opted to simply add an "active" class and keep the DOM in place. This seems to solve the issue of JAWS/NVDA getting lost on close.

mcking65:
@sh0ji, With JAWS and Firefox 52.5.2, JAWS behaves differently; the VC loses its place. However, in Chrome and IE it works fine. And, with NVDA, it works fine in all 3 browsers.

So, I think there is some kind of JAWS bug that is only manifest with this version of the code. The version in master works fine with Firefox 52.5.2 and JAWS 18. I am curious to find out what is different. One approach is that I could branch master before merging this just to keep a named location of the previous behavior handy for later comparison.

jnurthen:
From the limited testing I did it seems to be an impovement.

annabbott:
Dialogs are better in Evan's latest version plus the user's point of regard is maintained when canceling all the way out of the dialogs.

However, the issue of mouse clicking on the inert page when the dialog is open (this happens in all three dialogs) remains - focus goes 'somewhere' on mouse click, then returns to illogical field in dialog on next TAB keystroke.

I'm proposing that this issue is a very edge case and can wait till next version of the APG for a fix. Thus I vote we accept Evan's latest version for this release.

mcking65:
@sh0ji, thank you, thank you for all this work!! We got it in for the publication!

PR #583, opened 2018-01-23
552 fix

### LINKED ISSUE #552: Modal Dialog Example: Background not entirely inert; it allows selection.
In the [modal dialog example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html), the background is not entirely inert.

In Safari and Chrome (macOS) I can make a selection with my mouse that starts in the dialogs, but by dragging further, I can include the content in the background. I believe this can be fixed in CSS with user-select: none (and variants) on the `base_window_layer` container.

In Chrome using VoiceOver I can reach items in the background (headings, links, etc) by using the rotor and navigating the page item by item. I believe this can be fixed with `aria-hidden="true"` on the `base_window_layer` container.

sh0ji:
Fixes #552, as recommended by @erikkroes.

To test with a keyboard: open a dialog, move focus to a non-editable element, and press ctrl-a to select all. Inspect the contents of that selection.

While this does technically disable user-selection when a dialog is active, it also adds coupling to the HTML structure outside of the dialog, which concerns me. Specifically, prior to this PR, dialogs operated as standalone units that didn't affect other elements on the page. With this PR, `user-select: none;` needs to be applied to everything that we don't want to be selected. That's just the `#base_window_layer` here, but that's only a coincidence of how the HTML is structured.

The only other alternative I can imagine would be messing about with selection listeners, and I don't like that very much either.

I'd be okay with delaying this or not merging it.

### DIFF
--- a/examples/dialog-modal/css/dialog.css
+++ b/examples/dialog-modal/css/dialog.css
@@ -23,6 +23,10 @@ body:not(.toc-inline) {
   padding: 1.6em 0 2em;
 }
 
+.dialog-active #base_window_layer {
+  user-select: none;
+}
+
 #base_window_layer main, #base_window_layer nav {
   max-width: 50em;
   margin: auto;
--- a/examples/dialog-modal/js/dialog.js
+++ b/examples/dialog-modal/js/dialog.js
@@ -131,6 +131,8 @@ aria.Utils = aria.Utils || {};
         'Dialog() requires a DOM element with ARIA role of dialog.');
     }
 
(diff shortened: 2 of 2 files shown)


mcking65:
Based on discussion with Evan and other members of the task force in the [February 5, 2018 meeting](https://github.com/w3c/aria-practices/wiki/February-5%2C-2018-Meeting), closing this PR with a decision not to fix  issue #552.

Issue #451, opened 2017-09-07
Menubar Examples: When focus is in menubar, Escape should close any open submenus


mcking65:
In the 
[navigation menubar example](http://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html), 
when users tab into the menubar, the submenu associated with the focused menubar menuitem is displayed, covering content underneath. There is no way to hide the submenu without tabbing out of the menubar. Shouldn't the Escape key hide the submenu?

To reproduce:

1. After page load, tab to the menubar. The "About" menu item is focused and the "About" menu is expanded.
2. Press escape.

Expected result: The "About" menu collapses.
Actual result: Nothing happens.

Similarly, while the "About" menu item has focus, continue by:

1. Press down arrow. Focus moves to "Overview".
2. Press right arrow. The "About" menu collapses, focus moves to "Admissions", and the "Admissions" menu expands.
3. Press escape.

Expected result: The "Admissions" menu collapses.
Actual result: Nothing happens.

devarshipant:
It seems to work with keyboard (IE 11). Pressing the Escape key sets keyboard focus on the parent menu item.
Sidebar - This maybe JAWS / IE 11 specific, but JAWS only speaks ‘use up and down arrow keys’ when keyboard focus is on the menu item. A helpful hint, but might help if it said use arrow keys instead. There are left and right arrow keys that make the menu navigation complete.

tatermelon:
I saw this issue on Mac Chrome. Esc works when focus was inside one of the submenus, but not when focus is on the menubar and a submenu is shown. You can repro by pressing right/left when a submenu is shown, then pressing Esc

jongund:
I am not observing the problem with escape key not closing menus in the current menubar branch

a11ydoer:
@mcking65 In your second testing scenario, it is testing the escape key function when the focus is on parent menu, menubar item "Admissions",  not on submenu(role=menuitem).
As it was mentioned in the "keyboard support" section, escape key should work on submenu(menuitem), not in the parent menu, menubar item. According to this, the current example does not have any problem with escape key support.

a11ydoer:
ok. Group agrees that escape key should close wherever the focus is. The keyboard part of Menu, not submenu and coding will be updated.

mcking65:
@jongund, With d9a1f79, this issue is fixed in the editor menubar (menubar2).

However, we still have two open Escape key issues with the navigation menubar (menubar1):
[...]

PR #589, opened 2018-01-29
escape key now closes open popup menus when focus is on a menubar men…

jongund:
This change fixes the issue  #451


mcking65:
@jongund, for the editor menubar example, this was resolved with pr #593.  We want the fixes that are in #593 for the navigation menubar, so closing this.

Issue #408, opened 2017-06-21
Menubar examples: Make mouse behaviors and parent menu item appearance change triggers more consistent with desktop conventions


mcking65:
The following two problems with the 
[editor menubar example](http://w3c.github.io/aria-practices/examples/menubar/menubar-2/menubar-2.html) 
and the [navigation menubar examples](http://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html) 
were raised in issue #144.

#### Problem 1 described by @carmacleod in issue #144:

> When moving a mouse horizontally across a menubar, hovering over each menu button to see what is in its menu, 
> each newly-opened menu appears instantly (good) 
> but the previously-open menu hesitates before closing, 
> which looks like both menus are open for a while, which feels really odd. 
> This odd behavior does not happen when I use right/left arrow keys to traverse through the menus.

#### Problem 2 described by @carmacleod in issue #144:

> Using either keyboard or mouse, when the user opens a menu and then traverses to a menu item inside that menu, 
> the menu button in the menubar no longer looks selected. 
> For example, open the Font menu, and arrow down to the Serif menu item. 
> The selected look of the Font menu button disappears. 
> This is different from the look of a desktop menubar, which would show the Font menu button as "pushed in" (for Windows menubars) or "blue" (for Mac menubars) 
> while the Font menu is open.

#### Consensus resolution documented in issue #144

Agreement on the following mouse behaviors is documented in issue #144.

1. Open menus remain open if the user moves the mouse away; remove timeouts that close menu when the mouse moves away.
2. Mouse behaviors that cause an open menu to close are:
    A. User clicks anywhere outside the menu. 
    B. User clicks an action item in the open menu.
    C. User hovers over a different parent menu item in the menubar, triggering display of that parent's submenu.
3. Submenus open on hover over a parent item in the menubar only if another submenu is already open. That is, if all submenus are closed, a click on a parent menu item is required to display a submenu.

The task force also agrees to the behavior proposed by @carmacleod to resolve problem 2 described above. A parent menu item should retain it's activated appearance while its submenu is open.

carmacleod:
The above commit fixes the [menubar-2 example (Editor Menubar)](http://w3c.github.io/aria-practices/examples/menubar/menubar-2/menubar-2.html).
If you like these changes, then I will fix the [menubar-1 example (Navigation Menubar)](http://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html).
[...]

PR #593, opened 2018-02-01
Menubar examples: Make mouse behaviors and parent menu item appearanc…

carmacleod:
…e change triggers more consistent with desktop conventions #408

This commit fixes the menubar-2 example.

It ensures the following behavior:
- open menus stay open on mouseout
- remove timeouts that close menu on mouseout
- close menu when user:
  - clicks outside menu
  - clicks again on menubar item
  - clicks on menu item in menu
  - types tab or shift+tab
  - types space or enter on menu item (but not for space on radio/checkbox)
  - hovers over another menubar item (other menu opens)
- menubar item retains its activated appearance while its menu is open

Details of the code changes:
- deleted hasFocus and hasHover state booleans because they are no longer needed
- added popupMenu.isOpen() which checks for aria-expanded=true on the corresponding menubar item's DOM element
- added ESC key handling on menubar items (closes menu if open)
- clicking on menubar item opens menu, clicking again on same menubar item closes it
- make sure browser doesn't scroll when clicking on menubar item (default behavior for clicking on link)
- removed focus, blur, and mouseout event handling and timeouts from menubar items
- removed focus, blur, mouseover, and mouseout event handling and timeouts from popup menu items
- added focusout on menubar item and popup menu item to close menu if focus will be going outside of the menubar
- if a menu is open, then hovering over a different menubar item opens its menu instead
- removed check for anchor element children of menubar items (was probably an artefact from a previous implementation)
- removed mouseover and mouseout events from popup menu
- removed 'force' parameter from popup menu close() function
- deleted some vars that were declared but not used, or were declared twice
- fixed a few comments that didn't quite match the code
- declared some vars as local that were declared as global (i.e. without var keyword) but did not need to be
- removed some redundant checks, i.e. while (e) { if (e ...) }
- changed != to !== and == to === where appropriate
- removed css for hovering over menubar item (not needed because hovering focuses menubar item)


mcking65:
@annabbott, @sh0ji, @shirsha, @jongund,

You can view/test the result of these changes in the [RawGit view of the editor menubar example page in Carolyn's repo](https://rawgit.com/carmacleod/aria-practices/issue408/examples/menubar/menubar-2/menubar-2.html).

For your convenience, here is the [current published version of the editor menubar on w3/tr](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/menubar/menubar-2/menubar-2.html).

@carmacleod, thank you very, very much for this contribution. It's a big change. So, I am having several people look closely before commiting to master.

With multiple reviewers, there could be some churn in the comments. Please do not make changes unless we have clear agreement that a change is needed.

mcking65:
1. Windows Firefox/JAWS:
    * All kb commands work as described
    * All roles, states, and properties being set/changed as described.
2. Windows Chrome/JAWS:
    * All kb commands work as described
    * All roles, states, and properties being set/changed as described.

Note: above tests are of the example code, not the screen reader.

annabbott:
Tested with Win10/FF ESR 52.6.0/JAWS 2018 (build 1712.10 ILM)
All good!

annabbott:
Tested with Win10/FF ESR 52.6.0/kbd only (no screen reader)
All good!

annabbott:
Tested with Win10/Chrome Version 64.0.3282.140 (Official Build) (64-bit)/JAWS 2018 (build 1712.10 ILM)
Menubar: expected result - number of menuitems on menubar is announced (x of x) as navigated with arrow keys.
Menubar: actual result - number of menuitems on menubar is NOT announced correctly as navigated with arrow keys (always announces 'one of one' for each of four menuitems.

This is specific to Chrome. It is announced correctly on FF.

All good!

annabbott:
Tested with Win10/Chrome Version 64.0.3282.140 (Official Build) (64-bit)/kbd only (no screen reader)
All good!

Note: I was unable to replicate Jon's kbd-only test results.

sh0ji:
As a follow-up, I'd actually argue that the screen reader experience is better than the keyboard or mouse experience on this one now, though not because of changes made in this PR. We may want to review the visual design for that reason. Specifically:
* There is no visual indicator that the submenu items are a checkbox or radio. You just have to test it or guess to know whether clicking one will de-select another.
* Similarly, sub-groups (`[role="group"]` elements) have a label that makes it clear how the items are related, but that label is inaccessible to everyone except screen reader users. It would be useful for everyone.

carmacleod:
Thank-you all so much for your comments and testing!

I have a few replies:
1) The issue that @jongund mentioned was there originally, and I agree that it is odd. I believe that the intent was for the menu to stay open if the user selects a radio or checkbox menu item using the space key, and to close if the user selects a push button menu item with space, or any menu item with enter. This is listed as "Optional" behavior in the [APG design pattern for menu](http://w3c.github.io/aria-practices/#menu). It works the same whether the screen reader is running or not.
To test, use space to select any item in the first 3 menus (these are all radio or checkbox menu items), and then use space to select "Smaller" or "Larger" in the "Size" menu (these are the only push button menu items in this example). Please let me know if you would like this behavior kept or changed.

2) I opened a [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=807758) for the "menu 1 of 1" issue that @annabbott mentioned. It would be great for it to have a few more "Stars" to hopefully raise the priority a bit.  :)

3) I agree with @sh0ji that simply hovering over a menubar item - when no menus are already open -
 should not focus the item. I will fix that. Do you prefer that I open a new issue for this, or add another commit to this PR?

4) Good point about no visual indicator when submenu items are checkbox or radio, @sh0ji . I looked at the old Windows Notepad menu, and it has the same behavior, i.e. the user cannot tell, other than perhaps by context, that a menu item is a checkbox. For example, View -> Status Bar, or Format -> Word Wrap. I'm not sure what the best solution to this problem is; does Mac have anything nice? One possibility is to use a red X (or an empty box) when the checkbox is not selected, and a green checkmark (or green checkmark in a box) when it is selected. It would be nice to distinguish between radio and checkbox as well, so, for example, an empty circle (unselected) or blue dot within circle (selected) for radio.

5) Good point also about the grouping labels, however I think we need to be careful not to get too verbose in both sighted and screen reader menus. For example, very few menus in the past ever had group labels. If there truly was a need for a group label, then one might argue that the group would be better implemented as a submenu. In this example, every separator is followed by a group label. If you take a look at the menus in Notepad on Windows (I'm on Windows 7 - I don't know if the Windows 10 Notepad is the same) you can see that there are several separators, but no group labels. For example, the Edit menu has Undo, separator, Cut, Copy, Paste, Delete, separator, Find..., Find Next, Replace..., Go To..., separator, Select All, Time/Date. If we said that separators delineate groups, and groups need labels, then it would be very difficult to come up with a good label for each of those groups, and it would add unnecessary verbosity to the Edit menu.

mcking65:
@carmacleod, thank you for opening the chromium bug!!

@carmacleod commented:

> I agree with @sh0ji that simply hovering over a menubar item - when no menus are already open - should not focus the item. I will fix that. Do you prefer that I open a new issue for this, or add another commit to this PR?

We would like a commit added to this pull request for this focus on hover problem before merging.

> Good point about no visual indicator when submenu items are checkbox or radio, @sh0ji.

This would be a separate issue and PR. I opened issue #605.

> Good point also about the grouping labels, however I think we need to be careful not to get too verbose in both sighted and screen reader menus. For example, very few menus in the past ever had group labels. If there truly was a need for a group label, then one might argue that the group would be better implemented as a submenu. In this example, every separator is followed by a group label. If you take a look at the menus in Notepad on Windows (I'm on Windows 7 - I don't know if the Windows 10 Notepad is the same) you can see that there are several separators, but no group labels. For example, the Edit menu has Undo, separator, Cut, Copy, Paste, Delete, separator, Find..., Find Next, Replace..., Go To..., separator, Select All, Time/Date. If we said that separators delineate groups, and groups need labels, then it would be very difficult to come up with a good label for each of those groups, and it would add unnecessary verbosity to the Edit menu.

I agree. But, in this particular case, the group labels are particularly useful because they provide a screen reader analog to the visual separation. Remember that separators are not announced. That is generally fine since most of the time separators are there to help with visual scanning as opposed to communicating a specific intent. Here, labels like "none" aren't particularly good unless they are accompanied by the group label. And, btw, NVDA is the only screen reader that reads the group labels when they should be read.

I am guessing that sighted users can easily dicern "none" in the text decoration group means "no text decoration" because of its proximity. If that is not the case, then maybe that label should be changed to "No decoration". On the other hand, the visual separation of colors is surely adequate and there is no need for a visual label of "text color". I didn't open a new issue for this aspect of the visual design; I am not confident there is an issue.

I did note that we didn't document the use of aria-label on the groups so I created issue #606, which I can fix later after we merge.

sh0ji:
> I agree with @sh0ji that simply hovering over a menubar item - when no menus are already open -
should not focus the item. I will fix that. Do you prefer that I open a new issue for this, or add another commit to this PR? (@carmacleod)

> One issue to fix before merging raised by @sh0ji: hover should not move focus. (@mcking65)

I'm generally okay with the current implementation, which I'll summarize: when not active, hover either does nothing or provides visual feedback (it currently does this by inverting colors and changing to a pointer cursor), but it does not focus anything until the user activates the menubar through either a focus or click action. Once the user activates the menubar, focus follows the hovered item.

This is similar to how operating system menus work--macOS actually does nothing on hover until the user activates the menubar, but otherwise works identically to the current example.

The pattern of a hover event triggering a focus event is nearly unheard of on the web, though. And I argue that it's similar to [SC 3.2.1 in WCAG](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-receive-focus.html), but with the `hover` event replacing the `focus` event: "When any component [is hovered], it does not initiate a change of context." That rule exists because unexpected changes of context can be disorienting for users. It doesn't include hover, though, which is why I would be okay with the current implementation. Feedback from others would be useful here. I know @jongund may have some thoughts about this.

sh0ji:
@mcking65 commented:
> On the other hand, the visual separation of colors is surely adequate and there is no need for a visual label of "text color".

While I agree for our specific case, I don't want to assume this for the general case. There's a cognitive load associated with looking at a grouping of items and determining how they're related. That cognitive load can be mitigated by having an explicit label for the group, regardless of how the user experiences it. On a similar note, if we're okay with the cognitive load required to understand the grouping for sighted users, does it need a label at all?

I don't have answers for this. I just wanted to point out that the screen reader user has access to explicit information for a couple items that are only ever implicit for sighted users.

carmacleod:
Good news: the chromium "menu 1 of 1" bug is fixed, and will be in Chrome 67.

carmacleod:
This latest commit does not change any behavior; I only changed css. (and a comment)
Please give it a try. (I'm hoping you are all at CSUN and you can try it together :)

Please don't worry about focus following hover inside an open menu, because when the menu closes, focus goes back to the parent menuitem anyhow. So there's no possibility of loss of context, because we never left the context of the menubar.

Note again that focus does not follow hover unless the menubar or one of its menus already has focus.

carmacleod:
Latest fixes now available on rawgit. Please try and let me know.
https://rawgit.com/carmacleod/aria-practices/issue408/examples/menubar/menubar-2/menubar-2.html

Issue #535, opened 2017-11-22
Develop example of a tri-state toggle button


mcking65:
The ARIA 1.1 specification refers readers to ARIA Practices for examples of toggle buttons that have `aria-pressed="mixed"`. Develop an example of this and ensure appropriate guidance is present in the button pattern.

kleinfreund:
Hey, I’d like to work on this. I have a few questions.

The examples on [Button Examples](http://w3c.github.io/aria-practices/examples/button/button.html) are not organized well. I would like to separate the example JS/CSS into their own directories (i.e. `button/`, `toggle-button-twostate/` and `toggle-button-tristate/`. If this is something you want, should it be done in a separate pull request? If so, I would prefer to do that first, because it would allow me to write some generic button styles that the specific examples can reuse.

Also, I wonder what the correct spelling of *tristate*/*tri-state* is. The spec language uses `tristate` when they refer to the *value type* ([reference](https://w3c.github.io/aria/aria/aria.html#valuetype_tristate)). However, in natural language, I believe it should be *tri-state* (and *two-state* respectively).

I’m also having a hard time figuring out what a good example might be for a *tri-state toggle button*. My concept looks like this:

---

A sound control with the following elements:

- A main button, labelled "Sound"
- Two additional buttons "Audio" and "Vibration"

The main button is the one that can exists in three states ("true", "false" and "mixed"). Toggling it cycles through the two usual states of a toggle button, "true" and "false". Whenever the main button is toggled, the dependant buttons receive the same state (e.g. turning off "Sound" also turns of "Audio" and "Vibration").

The dependant buttons are able to bring the main button in the "mixed" state. This happens whenever the dependant buttons don’t share the same state. Otherwise, the main button will receive the shared state of the dependant buttons.

---

What do you think? Originally, I thought I could implement the tri-state buttons that are commonly used for phones’ audio control: Tapping the button cycles through the modes "Sound", "Vibration" and "Silent". However, the specification doesn’t allow a toggle button to trigger the "mixed" state on itself ([reference](https://w3c.github.io/aria/aria/aria.html#aria-pressed)).

mcking65:
@kleinfreund commented:

> Hey, I’d like to work on this. I have a few questions.

Excellent! Thank you! And, thank you for raising your questions and proposal specifics here before working on a PR.

[...]

PR #609, opened 2018-02-21
Clean up button example code

kleinfreund:
*In preparation for developing examples for tri-state toggle buttons (see issue #535).*

This pull request addresses/changes the following:

- Fixes indentation issues in the markup
- Removing `type="text/javascript"` (some scripts have it, some don’t. I chose one.)
- Removed `/` in the meta charset tag (some examples have it, some don’t).

Please note: Due to indentation changes, the GitHub diff doesn’t show some changes. For example,I changed `<th>` to `<td>` elements when inside `<tbody>`. It may have been incorrect to do so.

### DIFF
--- a/examples/button/button.html
+++ b/examples/button/button.html
@@ -37,7 +37,7 @@
     <ul>
       <li><a href="../menu-button/menu-button-links.html">Navigation Menu Button</a>: A button that opens a menu of items that behave as links.</li>
       <li><a href="../menu-button/menu-button-actions.html">Action Menu Button Example Using <code>element.focus()</code></a>: A button that opens a menu of actions or commands where focus in the menu is managed using <code>element.focus()</code>.</li>
-      <li><a href="../menu-button/menu-button-actions-active-descendant.html">Action Menu Button Example Using <code>aria-activedescendant</code></a>: A button that opens a menu of actions or commands where focus in the menu is managed using aria-activedescendant.</li>
+      <li><a href="../menu-button/menu-button-actions-active-descendant.html">Action Menu Button Example Using <code>aria-activedescendant</code></a>: A button that opens a menu of actions or commands where focus in the menu is managed using <code>aria-activedescendant</code>.</li>
     </ul>
 
     <section>
@@ -73,11 +73,11 @@
         </thead>
         <tbody>
           <tr>
-            <td><kbd>Enter</kbd></td>
+            <th><kbd>Enter</kbd></th>
             <td>Activates the button.</td>
           </tr>
           <tr>
-            <td><kbd>Space</kbd></td>
+            <th><kbd>Space</kbd></th>
             <td>Activates the button.</td>
           </tr>
         </tbody>
--- a/examples/button/button.html
+++ b/examples/button/button.html
@@ -1,18 +1,19 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-<meta charset="utf-8" />
-<title>Button Examples | WAI-ARIA Authoring Practices 1.1</title>
+  <meta charset="utf-8">
+  <title>Button Examples | WAI-ARIA Authoring Practices 1.1</title>
 
-<!--  Core js and css shared by all examples; do not modify when using this template. -->
-<link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/base.css">
-<link rel="stylesheet" href="../css/core.css">
-<script src="../js/examples.js" type="text/javascript"></script>
-<script src="../js/highlight.pack.js"></script>
-<script src="../js/app.js"></script>
+  <!-- Core JS and CSS shared by all examples. Do not modify when using this template. -->
+  <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/base.css">
+  <link rel="stylesheet" href="../css/core.css">
+  <script src="../js/examples.js"></script>
+  <script src="../js/highlight.pack.js"></script>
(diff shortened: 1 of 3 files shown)


kleinfreund:
@mcking65 I did not consider that. Apologies. Ideally, I would have made two separate commits; one for whitespace, one for other changes. I will provide you with a list of changes that are hidden by whitespace changes. I will refer to lines by updated line numbers.

- line 4: Omitting slash in self-closing tag
- line 10, 16: Omitting `type="text/javascript"` in `script` tag.
- line 15: Moving `rel` attribute before the `href` attribute to match lines 8 and 9.
- line 76, 80: Replaced `th` tags with `td`. I will revert this change as per your review. I will have a look into the side-effect of having the `th` there overriding the `kbd` styles.

A single extra newline was added in the following lines: 47, 51, 60, 66, 89, 156, 171, 173, 175, 177 and 184.

A personal note: You’re not to blame for not being able to make out the changes. The diff does not provide any cue for some of the changes made inside of blocks that also change whitespace. Also, there isn’t a visual cue. One would have to read the actual contents in order to notice. In the future, I will try not to mix whitespace changes with other changes to avoid such issues.

Thank you for reviewing this.

kleinfreund:
I pushed two commits with the following changes:

- Reverted the change from the first commit where some `th` tags were incorrectly replaced with `td`.
- Wrapped an occurence of `aria-activedescendant` in a `code` tag
- The core CSS was changed so that `kbd` elements are always styled. Previously, the selector only targetted `kbd` elements that were not inside a table heading. Please note that I’m concerned about this change as there might be a good reason for the rule being as explicit as it was. From my stand point, it just stopped `kbd` elements from being styled in places where they should be.

mcking65:
Thank you for rebasing on latest! That always helps.

When multiple commits come from a fork, I squash when I merge ... no need to do it on your end.

I want to have a sightee or two review the core.css change before merging. The rest all looks good to me!

mcking65:
@jnurthen, would you please review the changes to core.css in commit e7c83a2cba94fd8bad2a90874018d4b3084693b9?

kleinfreund:
First of all, I agree with @sh0ji: A default styling should be applied to `kbd`, not `:not(th) > kbd`.

With a default styling for `kbd`, it seems odd to reset these for `th kbd`. It would be easy if `all: unset` could be used, but the support for that is lacking (e.g. not available in Edge or Safari). Instead, we would need to set almost all properties from the default styling to `initial`. That seems unnecessary. Also, I don’t think keeping the styles inside table headers makes the table too busy.

Anyway, should I just revert back to `:not(th) > kbd` (avoids unnecessarily overwriting styles) or keep the default styles for `kbd` and reset them explicitly for `th kbd`?

MichielBijl:
@sh0ji I changed that because the column is called “Key”. There’s no need for a visual cue to indicate these are keys.

sh0ji:
@MichielBijl You're absolutely right that the current selector is the shortest way of expressing "everywhere except X." I prefer a more verbose approach for two reasons:

1. It takes me longer to understand what `:not(th) > kbd` means compared to individual rules for `kbd`, and then another for `th > kbd`. It didn't take me too long to figure it out, but I can imagine someone less familiar with CSS spending significantly longer trying to understand it.
2. Keeping the specificity for root element styles as low as possible can help decrease the likelihood of future specificity errors.

"Base statement with low specificity -> override of base statement with higher specificity" is almost idiomatic of CSS.

kleinfreund:
> @kleinfreund why did you change background to background-color? Since we define new styles for the buttons it’s important to reset all old styles—this is exactly what happens when we use the shorthand background.

@MichielBijl I wasn’t aware that some user agents set other background properties than the color. Is that the case? I read that code as "set a background color", not "reset all background properties and set a background color". In fact, `background: <color>` resetting the other properties was the reason for that change. It seemed like an unintentional side-effect.

As for the selector, I agree with @sh0ji.

MichielBijl:
From the [CSS Backgrounds and Borders spec](https://www.w3.org/TR/css-backgrounds-3/):

> Given a valid declaration, for each layer the shorthand first sets the corresponding layer of each of background-image, background-position, background-size, background-repeat, background-origin, background-clip and background-attachment to that property’s initial value, then assigns any explicit values specified for this layer in the declaration. Finally background-color is set to the specified color, if any, else set to its initial value.

There’s a note with an example in the section I link to. 

---

As for the `kbd` thing, it's a core style with supplied with a comment meant to help the reader understand what it does. That piece of CSS is only used on the example pages—it's reader base is already pretty small I reckon. I really don't want to add more CSS to that than necessary. If you feel the current notation is unclear even with the provided comment we should address that comment.

When it comes down to specificity it doesn't really matter in this case—the specificity of the original selector is already really low and can easily be overwritten where necessary. Here's a [CodePen example that demonstrates the specificity](https://codepen.io/Michiel/pen/RyZyLW/).

kleinfreund:
> The reset is intentional. From the CSS Backgrounds and Borders spec:

I’m aware of the behavior and its intentions with regards to the specification. I was talking about the author perspective: Was it intentional resetting the background styles? Why? Presumably due to some user agents using background properties that are not `background-color`. I’m not sure there are any, but I can only check a few (Chrome, Firefox and Edge just have a color, nothing else). 

---

Regarding specificity, this is really not an issue in this project.

Thank you for reviewing this. I reckon much was discussed around a very small issue/non-issue without any big necessity. I will revert the changes to the CSS file tomorrow.

Issue #622, opened 2018-03-18
ARIA 1.1 Combobox examples have extraneous "null" entry in IE/Edge


patrickhlauke:
Both https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html and https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/grid-combo.html have an extra "null" at the start of the results in both IE and Edge (but not in Chrome, Firefox, etc)

![aria-combobox-edge-ie](https://user-images.githubusercontent.com/895831/37565101-94c599ee-2a9a-11e8-88e1-3766eb70cc99.png)

patrickhlauke:
```
<ul class="listbox" id="ex1-listbox" role="listbox" aria-labelledby="ex1-label">
null
<li class="result" id="result-item-0" role="option">Radish</li>
<li class="result" id="result-item-1" role="option">Rhubarb</li>
</ul>
```

mcking65:
@patrickhlauke, thank you very, very much for catching this bug and fixing it!

patrickhlauke:
no worries. a strange webcompat bug with IE/Edge there, though. may be worth scouring the repo for more ` ... = null` cases

PR #623, opened 2018-03-18
Fix combobox examples markup, fix IE/Edge "null" entry bug

patrickhlauke:
Closes https://github.com/w3c/aria-practices/issues/622


patrickhlauke:
(just force-pushed a cleaner solution to the "null" entry problem in IE/Edge)

mcking65:
@tatermelon, could you please review this?

Issue #579, opened 2018-01-22
Fix HTML Validation issues in layoutGrid


jnurthen:
Get following error:
"file:/home/travis/build/w3c/aria-practices/examples/grid/LayoutGrids.html":375.11-375.57: error: Bad value “true” for attribute “disabled” on element “button”.

mcking65:
Resolved with commit 31c7b7f and thanks to @fALKENdk.

PR #645, opened 2018-04-16
Fix html validation issues in layoutgrid #579

fALKENdk:
Fixed error "Bad value true for attribute disabled on element button." described in #579 
Fixed Nu HTML warnings "The type attribute is unnecessary for JavaScript resources."


mcking65:
@fALKENdk, thank you very much for this contribution!! Looks good.

PR #646, opened 2018-04-16
Escape html tags inside comments in examples

fALKENdk:
Right now when a HTML tag is used in a comment it is not escaped when the example is rendered by example.js
Causing a bug on http://w3c.github.io/aria-practices/examples/accordion/accordion.html where is renders to dl-elements in the example code.


mcking65:
@fALKENdk, thanks for the pr ... but this was addressed along with other issues in pr #653.

PR #653, opened 2018-04-20
Refactoring examples.js (Fixing indentation, various bug fixes)

kleinfreund:
This pull request refactors the [examples.js](https://github.com/w3c/aria-practices/blob/master/examples/js/examples.js) script and addresses various issues:

- Fixed: The outer-most elements being indented by two extra spaces
- Fixed: Source code starting with unnecessary empty line
- Fixed: Indentation of multi-line text nodes (contents of `<body>` and `<script>` in particular)
- Fixed: Omission of closing tag for elements that had no child nodes (e.g. `<div role="separator">` wouldn’t be closed in the output). Closing tags are now omitted from void elements (see [HTML 5.3, The HTML syntax, Void elements](http://w3c.github.io/html/syntax.html#void-elements)) based on the actual tag name. Note: This doesn’t cover SVG elements like `<use>`.
- Fixed: Comments starting with `<--` instead of `<!--`
- Fixed: The characters `<` and `>` not being replaced with their respective HTML entities (supersedes #646).
- When attributes are wrapped, they are now aligned to the column of the first attribute. Previously, two spaces of indentation were missing.

I had a look at two thirds of the example pages to test whether this introduces some regressions. Apart from some examples that break due to other reasons, everything is looking good.

### DIFF
--- a/examples/js/examples.js
+++ b/examples/js/examples.js
@@ -5,7 +5,7 @@
 var aria = aria || {};
 
 /* ---------------------------------------------------------------- */
-/*                  ARIA Widget Namespace                        */
+/*                  ARIA Widget Namespace                           */
 /* ---------------------------------------------------------------- */
 
 aria.widget = aria.widget || {};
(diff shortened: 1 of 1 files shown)


jnurthen:
@kleinfreund You mentioned "Apart from some examples that break due to other reasons, everything is looking good." 
Did those examples break with the previous version too - or are they only broken with this one?

kleinfreund:
@jnurthen I was referring to examples that break due to problems unrelated to my changes. I haven't observed any examples breaking with the changes.

mcking65:
kleinfreund commented:

>I was referring to examples that break due to problems unrelated to my changes.

Which examples are the ones breaking due to problems unrelated to these changes? How are they breaking?

mcking65:
Just wondering ... why would we fix problems in the original HTML when displaying it? Seems like we should faithfully reproduce the HTML that is there. If the HTML needs to be cleaned up, that should be taken care of in the original source of the example before example.js comes along and displays it. So, for instance, why change `<--` to`<!--`? Why remove closing tags from voide elements?

kleinfreund:
> Just wondering ... why would we fix problems in the original HTML when displaying it?

The pull request doesn’t do that. It fixes problems with the algorithm that takes the example HTML and displays its markup in a code block.

For example, the [Modal Dialog Example](https://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html#sc1_label) page.Note that in the source HTML, the comment is properly started with `<!--` but in the output, it’s `<--`.

> So, for instance, why change `<--` to `<!--`?

Because the algorithm handles comments wrong. It takes comment nodes’ content and displays them starting with `<--`. The algorithm is wrong, not the example code.

> Why remove closing tags from voide elements?

This is not an added feature. The algorithm did that before, but it incorrectly assumed that void elements are all elements that have no child nodes. This is wrong for `<div></div>`, but the algorithm would omit `</div>` in the ouput. Now, the algorithm bases this decision on the list of elements that are void per the specification.

> Which examples are the ones breaking due to problems unrelated to these changes? How are they breaking?

For example, the [Modal Dialog Example](https://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html) is rendered differently than other example pages. The sidebar is a couple of pixels rendered off from the right screen border and the main column is more to the right than with the typical example pages. But again, those are issues that are not affected or dealt with by this pull request.

mcking65:
@kleinfreund, thank you! I am sometimes amazed at some of the things that go unnoticed for so long until another person comes along with fresh perspective. How, for example, did I never hear the missing `!` in the comment opening. Thank you for this great attention to detail.

I also was not aware that the formatting algorithm was removing closing tags from void elements. I think the format/display algorithm should reproduce the HTML as written so would happily accept another PR that removed that feature from the algorithm.

kleinfreund:
> I also was not aware that the formatting algorithm was removing closing tags from void elements. I think the format/display algorithm should reproduce the HTML as written so would happily accept another PR that removed that feature from the algorithm.

I haven’t evaluated this properly, but I think when I worked on that particular piece of code, I had a reason to *fix it* instead of *removing it*. I think I wasn’t able to tell by the node object that is being processed whether there was a closing tag or not in the source. So I think I had to make a choice: Remove it or don’t. I think the version in this pull request is *more correct* although it might not be perfect. I happily look into it again once work on this pull request is done.

kleinfreund:
> I think the format/display algorithm should reproduce the HTML as written so would happily accept another PR that removed that feature from the algorithm.

@mcking65 Coming back to this, with this pull request being merged, things are working like this: Nodes are being examined by looking at [`Node.childNodes`](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes) and whatever is in that collection. The nodes in that collection don’t specifically carry the information whether their source has a closing tag. One could look at `node.outerHTML` and figure it out analytically.

I agree that it would be ideal to just *mirror* the content from the original example code without applying any stronger correcting logic. In this case, I don’t think any harm is done by leaving out the closing tags of void elements. If an example editor wrote `<input></input>`, the output will only have `<input>`. Only void HTML elements with an unneeded closing tag will be "corrected" and not contain that closing tag. Nothing will be added.

I don’t think the original HTML needs to replicated that strictly. What do you think? If you still want that, I’m able to implement it.

mcking65:
@kleinfreund, After more consideration, I agree with you; I think not including the closing tag on void elements even if present in the original source is an innocuous alteration. It  does not interfere with the purpose of the HTML source display. So, let's just leave as is.

PR #421, opened 2017-07-14
chore: add VNU checking

nschonni:
From the README:
> All HTML should validate in the NU HTML Validator. Exceptions to this rule are warnings and errors related to future ARIA features we need to implement.

Current errors are below, but may be related to current work
```console
"file:examples/landmarks/at.html":93.18-93.119: error: Bad value “80%” for attribute “width” on element “img”: Expected a digit but saw “%” instead.
"file:examples/landmarks/at.html":126.20-126.121: error: Bad value “80%” for attribute “width” on element “img”: Expected a digit but saw “%” instead.
"file:examples/landmarks/at.html":187.16-187.115: error: Bad value “80%” for attribute “width” on element “img”: Expected a digit but saw “%” instead.
"file:examples/landmarks/at.html":97.13-97.44: error: The “aria-labelledby” attribute must point to an element in the same document.
"file:examples/listbox/listbox.html":65.13-65.117: error: Bad value “” for attribute “aria-activedescendant” on element “ul”: An ID must not be the empty string.
"file:examples/listbox/listbox.html":124.13-130.42: error: Bad value “” for attribute “aria-activedescendant” on element “ul”: An ID must not be the empty string.
"file:examples/menu-button/menu-button-links.html":59.15-59.30: error: Bad value “none” for attribute “role” on element “li”.
"file:examples/menu-button/menu-button-links.html":64.15-64.30: error: Bad value “none” for attribute “role” on element “li”.
"file:examples/menu-button/menu-button-links.html":69.15-69.30: error: Bad value “none” for attribute “role” on element “li”.
"file:examples/menu-button/menu-button-links.html":74.15-74.30: error: Bad value “none” for attribute “role” on element “li”.
"file:examples/menu-button/menu-button-links.html":79.15-79.30: error: Bad value “none” for attribute “role” on element “li”.
"file:examples/menu-button/menu-button-links.html":84.15-84.30: error: Bad value “none” for attribute “role” on element “li”.
"file:examples/toolbar/toolbar.html":41.13-41.95: error: Attribute “aria-selected” not allowed on element “div” at this point.
```

To ignore work in progress the [--filterpattern](https://www.npmjs.com/package/vnu-jar#--filterpattern-pattern) option could be used till patches are landed in the validator.

/cc @sideshowbarker

PR #654, opened 2018-04-23
bug(toolbar): bad mime type for CSS links

nschonni:
Cherry-picked this out of #421 since it can land while that's still being discussed


MichielBijl:
Thanks Nick :)

PR #655, opened 2018-04-23
chore: Update ESLint task globbing and autofix

nschonni:
Just noticed that the globbing that ESLint wasn't catching everything. There is already a `.eslintigore` file to exclude third party files that should be excluded.
Ran the `fix` target to apply the formatting after extending the glob pattern.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,7 +1,4 @@
 **/*.min.js
+common/
 examples/landmarks/js/visua11y.js
-examples/landmarks/js/SkipTo.min.js
-examples/landmarks/js/bootstrap-accessibility.min.js
-examples/landmarks/js/bootstrap.min.js
-examples/landmarks/js/jquery-2.1.1.min.js
 examples/js/highlight.pack.js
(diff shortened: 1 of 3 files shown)


nschonni:
@mcking65 sorry this is going to be an ugly one to review. The github view with [`?w=1`](https://github.com/w3c/aria-practices/pull/655/files?w=1) might be a little less noisy.
I believe all the changes outside of the package.json are:
- Double quotes to single quote conversion. Looks like the `--fix` option just escapes internal double quotes after the external quote is changed to single
- Indenting to 2 spaces
- I manually had to fix a few hard tabs that It couldn't figure out. That might be preventable in the futrure by adding the whitespace configuration in the `.eslintrc.json` in addtion to the `.editorconfig` file
- Few swaps from leading to trailing comma
- Some spacing corrections around the `function` keyword

The only nice thing after this, is you shouldn't have to do another whitespace fixing PR here

jnurthen:
@nschonni I believe this is all coming from aria-common - https://github.com/w3c/aria-common

We do not modify these files in this repository as these are common files across all ARIA specs. As such they should be excluded from eslint in this repository.

nschonni:
OK, I'll add that to the ignore and rebase

Issue #660, opened 2018-04-24
Typo in grid element focus section heading


carmacleod:
The heading for "[Whether to Focus on a Cell Or an Element Inside IT](http://w3c.github.io/aria-practices/#gridNav_focus)" has a typo.
The word `it` is all uppercase, and should just be capitalized (i.e. the `t` should be lowercase.

mcking65:
Fixed with commit df74455.

PR #661, opened 2018-04-24
Typo in grid element focus section heading #660

carmacleod:


jnurthen:
I think the capitalization of our headings is weird - but this is certainly an improvement!

PR #668, opened 2018-04-26
Correct text error "withthe" should be "with the"

KevinTCoughlin:
Fixes a minor text error I encountered while reading the modified files. The text "withthe" should be "with the".

Hope you don't mind the patch. Love the site and work, very helpful!


MichielBijl:
Thank you very much for your contribution Kevin! We couldn’t do it witouththe help of the community 😀

Issue #225, opened 2016-12-14
Review Navigation Treeview Example Using Computed Properties


mcking65:
All clean up of the [Navigation Treeview Example Using Computed Properties](http://w3c.github.io/aria-practices/examples/treeview/treeview-2/treeview-2a.html) is complete and it is ready for final task force review.

#### Task Force Member Reviews Requested as of Nov 20, 2017

- [x] Review by Ann (@annabbott)
- [ ] Review by Siri (@shirsha)
- [x] Review by Evan (@sh0ji)

#### Required Updates

- [x] Editorial clean up
- [x] Issue #364: Navigation Tree View examples: enter key and modifier keys are not correctly handled
- [x] Issue #365: Tree View Examples: Fix Incorrect Left Arrow Behavior
- [x] Issue #404: Disable default focus ring and implement hover and focus states in css
- [x] Issue #406: Plus/Minus characters are being read by screen readers 
- [x] Issue #712: Treeview Examples: Hovering to right of tree causes blinking

annabbott:
In the Keyboard Support section, currently the following is listed:
Home Moves focus to first node.
End Moves focus to the last node that can be focused without expanding any nodes that are closed.

Question: does Home expand any nodes that are closed?

Also, none of the links with role="treeitem" causes a page to load either via keyboard or mouse.

mcking65:
Two more issues that need to be resolved before we can complete this review:

1. Issue #364: Navigation Tree View examples: enter key and modifier keys are not correctly handled
2. Issue #365: Tree View Examples: Fix Incorrect Left Arrow Behavior

mcking65:
This issue is now dependent on issues #365, #385, and #386.

annabbott:
Keyboard Support  > last row:
Shows an asterisk character, which for low vision users could be mistaken for a spot of dust on the screen.

Suggest adding "(asterisk)" without the double quotes after the character for clarity.

annabbott:
No other issues noted.

sh0ji:
The `hover` class is being added and removed very rapidly when hovering over the `[role=treeitem]`'s section that is outside the width specified in the corresponding CSS selector. The result for the user is white/gray flashing that exceeds 3 flashes per second.

This appears to be because the corresponding span (`firstElementChild` of the tree item) is full-width on mouseover, and then CSS immediately sets it to `16em`, triggering the mouseout event.

```css
[role="treeitem"].hover,
[role="treeitem"] span.hover {
[...]

PR #689, opened 2018-05-23
Treeview examples: fix blink on hover for issue #712

aristov:
Fix a disgusting blinking effect appearing on hover to the right of a treeitem.

### DIFF
--- a/examples/treeview/treeview-1/css/tree.css
+++ b/examples/treeview/treeview-1/css/tree.css
@@ -27,6 +27,7 @@ ul[role="tree"] {
 
 [role="treeitem"],
 [role="treeitem"] span {
+  width: 9em;
   margin: 0;
   padding: 0;
   border: 2px transparent solid;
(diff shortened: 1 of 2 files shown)


aristov:
I use Chrome Version 66 on Mac OS X.
Go to [the example page](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html) and place a cursor to the right of [the Treeview widget](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html#ex1).
All treeitems have 100% width, but on hover they become highlighted their width become 16em. Since their width becomes less they become not hovered and 100% width again. This process cycles and I see blinking effect, which looks buggy.

sh0ji:
I brought up this issue a while back in https://github.com/w3c/aria-practices/issues/225#issuecomment-351110621.

This looks good to me, but it's also happening in [treeview-1](http://w3c.github.io/aria-practices/examples/treeview/treeview-1/treeview-1a.html). @aristov, would you mind adding this fix to that example as well?

mcking65:
@sh0ji, thanks for taking a look. 

The css change looks straight forward. But, given this is purely visual, I'm going to rely on the sightee judgments here. Did you also verify in Firefox or Safari, or in any Win browsers?

sh0ji:
I tested this in Chrome, Firefox, and Safari on Mac, but haven't had a chance to check on Windows. The blinking issue was more prevalent in Chrome, and this fixes it in all three.

aristov:
Done.

mcking65:
@aristov, thank you, thank you for this contribution!

PR #694, opened 2018-05-24
Checkbox grouping label Issue 604

jongund:
We have added the role=group and aria-labelledby to reference the grouping label.
We have also updated the example documentation to include the role=group and aria-labelledby additions.

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,3 +1,0 @@
-[branch]
-  autosetuprebase = always
-
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,3 @@
+[branch]
+  autosetuprebase = always
+
--- a/examples/checkbox/checkbox-1/checkbox-1.html
+++ b/examples/checkbox/checkbox-1/checkbox-1.html
@@ -37,8 +37,8 @@
       <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
 
       <div  id="ex1">
-        <h3 id="sc1">Sandwich Condiments</h3>
-        <div role="group" aria-labelledby="sc1">
+        <h3 id="id-group-label">Sandwich Condiments</h3>
+        <div role="group" aria-labelledby="id-group-label">
           <ul class="checkboxes">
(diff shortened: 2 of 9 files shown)


.gitconfig
```diff
@@ -0,0 +1,3 @@
+[branch]
```
mcking65: Please remove this file from the PR. If you want to set this option, best to do it in .git/config (your local repo config file)

jongund:
Matt,

Git config should be removed now.

Jon

examples/checkbox/checkbox-1/checkbox-1.html
```diff
@@ -37,8 +37,8 @@ <h2 id="ex_label">Example</h2>
       <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
 
       <div  id="ex1">
-        <h3>Sandwich Condiments</h3>
-        <div>
+        <h3 id="id-group-label">Sandwich Condiments</h3>
+        <div role="group" aria-labelledby="id-group-label">
```
mcking65: Sorry I missed this before ... the states and properties table says that the group role is on the UL element. It is actually on a div here. I'm not sure which is better. The advantage of the div is the list semantics that says there are 4 checkboxes is preserved. The disadvantage is that is just more verbage on the page that may not be particularly valuable. I could go either way. We just need the documentation and the code to be consistent.

examples/checkbox/checkbox-1/checkbox-1.html
```diff
@@ -101,6 +101,35 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
         </thead>
         <tbody>
             <tr>
+              <th scope="row"></th>
+              <td></td>
+              <td><code>h3</code></td>
+              <td>
+                <ul>
+                  <li>Provides a grouping label for the group of checkboxes.</li>
+                </ul>
```
mcking65: This is the entry that is inconsistent with the code.

jongund:
Matt,

I updated the documentation to reference DIV element.

I think we should document in the example why we want the list semantics are not being overridden.

Jon

Issue #592, opened 2018-01-30
Navigation Menubar Example: Parent items are siblings of their submenu but pattern says they should be children


katiehockman:
From the authoring practices:
"If activating a menuitem opens a submenu, the menuitem is known as a parent menuitem. A submenu's menu element is contained in or owned by it's parent menuitem."
https://www.w3.org/TR/wai-aria-practices/#menu

Published example with error: https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html#
This example has a "Facts" menuitem with an attribute aria-popup='true'. This indicates it has a submenu. However, the submenu is a sibling of the "Facts" menuitem, rather than a descendant. This contradicts the suggestion in the APG.

mcking65:
This was discussed in the [February 12, 2018 task force meeting](https://github.com/w3c/aria-practices/wiki/February-12%2C-2018-Meeting). The task force consensus is to keep the guidance, and treat this issue as a bug.

The primary reason for establishing the parent/child relationship between a parent menuitem and the submenu it opens is to provide correct reading order when exploring by touch or with a reading cursor. This is possible with aria-owns, which effects the order of elements in the accessibility tree.

katiehockman:
Sounds good. The guidance from the APG makes sense, so I agree that fixing the menubar example would be preferable.

mcking65:
@jongund has proposed pull #707. This moves the menuitem role to the `<li>` elements, e.g.:

~~~
<li role="menuitem" aria-haspopup="true"><a href="#">parent menu name</a>
  <ul role="menu">
    <li role="menuitem"><a href="...">submenu item 1</a></li>
    <li role="menuitem"><a href="...">submenu item 2</a></li>
</ul>
</li>
~~~

That structure avoids a link containing links if the parent menuitem element is a link.

Pull #707 does this for all menuitems, even if they are not parent menuitems. That means that the `<li>` elements are focused and the links do not receive focus. One problem I see with that is that it removes the ability to access the link context menu with the keyboard.

So, I wonder, wouldn't a better structure be:

~~~
<li role="menuitem" aria-haspopup="true"><span>menu name</span>
  <ul role="menu">
    <li role="presentation"><a role="menuitem" href="...">submenu item 1</a></li>
    <li role="presentation"><a role="menuitem" href="...">submenu item 2</a></li>
</ul>
</li>
~~~

Based on discussion during today's practices teleconference, @accdc likely has a perspective on this.

[...]

PR #707, opened 2018-06-07
Navigation Menubar Example: Update structure to have parent/child relationships between parent menus and submenus

jongund:
Updated the code and CSS to support the parent child relationships for menus and sub-menus.

Note that the memuitem role has been moved from anchor elements to the listitem elements

I also added the feature requested in issue 159 that the expandable menuitem can also be links.


mcking65:
Closing based on discussion in issue #592.

PR #688, opened 2018-05-22
Add Alert Dialog Example

sh0ji:
Addresses #101.

I originally had this in the alert folder, but it primarily uses code from the dialog example, so I moved it there. Since it alters some of the code from the dialog example, we'll need to review that as well.

## Overview
* It is a "Confirmation" dialog used to confirm whether the user wants to take a destructive action.
* Defaults to "no" by placing that button first in the alertdialog (least destructive action).
* Confirming ("yes") will remove the contents in the `<textarea>`.
* Pressing the "discard contents" call to action when there are no contents triggers an alert. I chose to do this to illustrate the difference between an alert and an alertdialog, but it could easily be removed if it's out of scope.

## Changes
I had to change some pieces of the dialog.js and dialog.css used by the dialog-modal example in order to accommodate this new example. Those changes are documented here.
* Changes the way dialogs are positioned to better fit their contents (9b3e274)
* Added a box-shadow to the dialog element (7ddd9c5)
* Fixed a sentence fragment in dialog.html (3c92514)
* Adds `alertdialog` to the role validator in the Dialog constructor (6de2fd0)

b64883f changes a few things:
* Removed all body/`#base_window_layer` CSS from dialog.css.
  * These declarations tied dialog code very tightly to the document structure, which was causing issues in the new alertdialog example.
  * This was also overriding the default body layout defined by [base.css](https://www.w3.org/StyleSheets/TR/2016/base.css), which was actually causing some undocumented layout issues in the dialog-modal example. It can be seen in the current TR.
* dialog.js now adds a class to the body in order to disable body scroll when a dialog is open. `#base_window_layer` was previously handling scroll.
* Layout changes meant that the `#base_window_layer` was no longer doing anything, so I removed it from dialog.html

## Rawgit Links
* [alertdialog](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/alertdialog.html)
* [dialog](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/dialog.html)

***
<a href="https://pr-preview.s3.amazonaws.com/sh0ji/aria-practices/pull/688.html" title="Last updated on Jun 18, 2018, 5:01 PM GMT (e739b6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/688/a84d4d3...sh0ji:e739b6f.html" title="Last updated on Jun 18, 2018, 5:01 PM GMT (e739b6f)">Diff</a>

Issue #101, opened 2016-08-29
Develop example of alert dialog design pattern


mcking65:
Develop example of the [design pattern for alert dialog](http://w3c.github.io/aria-practices/#alertdialog).

#### Links to WIP in Evan's repo:

* [Alert dialog](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/alertdialog.html)
* [Modal dialog](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/dialog.html)

dfkaye:
@mcking65 - I had been working on an example of this pattern recently when I found this issue - would any of the code in the following gist suit your needs?

see https://gist.github.com/dfkaye/31939cd9fd3d00be68ffd5ab42622a35

mcking65:
Consider whether we need a separate folder or can add this to the dialog example folder.

mcking65:
All, please, test Evan's new code. Here are links to pages in the Rawgit view of Evan's repo:

* [Alert dialog](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/alertdialog.html)
* [Modal dialog](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/dialog.html)

mcking65:
Evan, I really like the simplicity of the approach for this example -- a discard text function.

Rather than an alert when there is no text to discard, seems like it would better demonstrate good practice to mark the button disabled. To include an alert, perhaps we put a 100 character limit on the field and then warn the person when they get to 90 with an alert and another alert when they get to 100. I suggest reasonably low numbers so it doesn't take a long time to experience the alerts.

annabbott:
[Modal Dialog Example](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/dialog.html) > HTML Source Code: all text is displayed in red instead of black, blue and red like in [Alert Dialog Example](https://rawgit.com/sh0ji/aria-practices/alertdialog-example/examples/dialog-modal/alertdialog.html) > HTML Source Code.

annabbott:
No other issues noted with Modal Dialog Example with FF 60.0.1 and Chrome 66.0.3359.181 on Win10. No screen reader testing performed.

annabbott:
Tested with FF 60.0.1 and Chrome 66.0.3359.181 on Win10.

[...]

PR #713, opened 2018-06-13
Fix and clean up modal dialog example

kleinfreund:
This pull request addresses the following issues:

**Clean up HTML**:

- Fixed highlight.js not being triggered due to code markup using a `<div>` element instead of `<pre>` and `<code>` elements.
- HTML head:
  - Consistent attribute order
  - Omitting unnecessary type attributes and closing slashes
- Content:
  - When used inside code, replaced `<q>` elements with regular double quote characters (i.e. `"`)
  - Added manual line breaks to the example code so that the example code block doesn’t have a scrollbar on medium to large screens
- Added missing `</li>` closing elements in some places
- Added a bunch of empty lines between some chunks of code for easier readability
- Fixed some minor indentation issues in some places

**Clean up JavaScript**:

- Fixed some documentation comments using JSDoc
- Removed the ability to pass an ID to the `focusAfterClosed` parameter because it isn’t used at all. This added unnecessary complexity to the example.
- Removed the ability to pass an element node to the `focusFirst` parameter. Same reason as above.

**Fixed Layout Issue**:

Removed the `div#base_window_layer` element wrapping the main content that was used to disable scrolling when any modal dialog is open.

The new implementation disables scrolling on the root element as soon as one dialog is open and restores the original property as soon as *all* dialoges are closed.

This fixed a major visual discrepancy of the dialog example page layout compared to the appearance of all other example pages. Most notably, the main content wasn’t aligned to the left of the screen and the main scrollbar was rendered a couple of pixels inside the viewport instead of the viewport border.

**Potential bug**:

The following code looks broken.

[**`aria.Utils.attemptFocus`**](https://github.com/w3c/aria-practices/blob/master/examples/dialog-modal/js/dialog.js#L67-L72):

```js
try {
  element.focus();
}
catch (e) {
}
aria.Utils.IgnoreUtilFocusChanges = false;
```

Is the last line supposed to be called inside the `catch` block? If not, what’s the purpose of the try-catch block if nothing happens?

### DIFF
--- a/examples/dialog-modal/css/dialog.css
+++ b/examples/dialog-modal/css/dialog.css
@@ -1,76 +1,26 @@
-html, body {
-  height: 100%;
-}
-
-body {
-  max-width: 100%;
-  padding: 0;
-  margin: 0;
-  overflow: hidden;
-}
-
-body:not(.toc-inline) {
-  padding: 0;
-}
-
-#base_window_layer {
-  box-sizing: border-box;
-  height: 100%;
-  width: 100%;
-  /* Use overflow scroll and webkit-overflow-scrolling to add momentum scroll*/
-  overflow: scroll;
-  -webkit-overflow-scrolling: touch;
-  padding: 1.6em 0 2em;
-}
-
-#base_window_layer main, #base_window_layer nav {
-  max-width: 50em;
-  margin: auto;
-  padding-right: 1.5em;
-  padding-left: 50px;
-  padding-left: calc(26px + 1.5em);
-}
-
-@media screen and (min-width: 78em) {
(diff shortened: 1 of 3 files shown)


mcking65:
@kleinfreund, thank you, but we are getting close to merging #688 by @sh0ji for issue #101, which includes at least one of the same changes while adding the alert dialog example for APG 1.1 release 2. I suggest that after we merge #688 that you rebase and resubmit this.

kleinfreund:
Ah, yikes, I did look at that pull request but disregarded it due to its title and first couple of commit messages. Will have a look into it once that pull request is merged.

Issue #588, opened 2018-01-29
Menu Button Examples: Shift+Tab from menu does not close menu


mcking65:
As pointed out in issue #577 by @brooksienoodlesoup, in the following 3 menu button examples, when focus is in the open menu, pressing shift+tab should close the menu and refocus the button. The button gets focus, but the menu remains open.

1 [Navigation Menu Button Example](http://w3c.github.io/aria-practices/examples/menu-button/menu-button-links.html)
2. [Action Menu Button Example Using element.focus()](http://w3c.github.io/aria-practices/examples/menu-button/menu-button-actions.html)
3. [Action Menu Button Example Using aria-activedescendant](http://w3c.github.io/aria-practices/examples/menu-button/menu-button-actions-active-descendant.html)

Owlbertz:
If there isn't already someone taking care of this I would like to tackle this one and get in touch with this project.

mcking65:
@Owlbertz, are you still interested in helping with this? Very sorry that I did not ask sooner.

jongund:
Should be fixed in pull request 716

jongund:
Should be fixed in pull request 716

mcking65:
Thank you @jongund, this is now fixed in commit 8bc433e.

Issue #396, opened 2017-05-12
Navigation Menu Button Examples: Unreachable code in PopupMenuLinks.js


racer2207:
`  while (childElement) {
    var menuitem = childElement.firstElementChild;
    if (menuitem && menuitem === 'A') {
      throw new Error(msgPrefix + 'has descendant elements that are not A elements.');
    }
    childElement = childElement.nextElementSibling;
  }`

Line 53 of PopupMenuLinks.js

I believe the conditional statement will never equal true since menuitem is an element and not a string. Wouldn't this work better?

` if (menuitem && menuitem.tagName !== 'A') {`

mcking65:
@racer2207, thank you for catching this and providing a detailed description of the problem.

@jongund, thank you for implementing the fix.

It's almost embarrassing this sat unfixed for more than a year, but it is now fixed in commit 8bc433e, which is landed in master. It will be published to w3.org in release 2 of APG 1.1 in mid July.

PR #716, opened 2018-06-15
Navigation Menu Button Example: fix unreachable code and shift-tab not closing menu

jongund:
Issue #396: Fixes the unreachable code problem
Issue #588: Fixes shift-tab not closing the menu
Updated CSS focus styling

#### Preview link

[Navigation menu button example in Jon's fork](https://rawgit.com/jongund/aria-practices/issues-396-588-menubutton/examples/menu-button/menu-button-links.html)

### DIFF
--- a/examples/menu-button/css/MenubuttonLinks.css
+++ b/examples/menu-button/css/MenubuttonLinks.css
@@ -3,16 +3,20 @@
 }
 
 .menu_button button {
+  margin: 1px;
   font-size: 110%;
   display: block;
-  padding: 0.25em;
-  border: inset 2px gray;
+  padding: 0.375em;
+  border: solid 0.125em gray;
+  border-radius: 8px;
   width: 12em;
   text-align: center;
-  background-color: #EEEEEE;
-  text-decoration: none;
+  background-color: #EEE;
   color: black;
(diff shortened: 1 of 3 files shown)


mcking65:
@jongund, the fix for issue  588 looks good. I think 396 as well; @sh0ji, could you look at that?

I'm curious about the css changes. They were not part of either of these issues. Please describe them.

sh0ji:
The 396 fix looks good to me.

The style changes look purely cosmetic to me, but there are at least two issues:
* The border of the `role=menu` is not connecting at the top-left and top-right corners the way it should.
* The `role=menuitem` boxes resize on hover a tiny bit, causing the text to shift.

I'd recommend doing the style changes in a different PR if that's possible.

mcking65:
@sh0ji, thank you! @jongund, I too would prefer a separate PR for the styling. In general, I prefer one PR per issue ... it doesn't have to always be that way for small issues, but it is a good rule of thumb.

Please back out the CSS file changes and I will merge.

Then, if you want to address Evan's feedback in another PR, that would be good.

jongund:
I removed the CSS changes from the original pull request

PR #717, opened 2018-06-15
Issue 559 combobox 1.0 fixed escape key not clearing textbox

jongund:


mcking65:
@jongund, it seems your issue-559-combobox-1.0 branch had a dirty history with stray commits and lots of eslint errors. However, it appeared that your commit a0cb251 has the fix, so I cherry-picked it into a clean branch and replaced this PR with PR #734.

PR #719, opened 2018-06-15
Initial Changes for v1.2

jnurthen:
Changed version numbers for v1.2. 
Removed Changelog

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/719.html" title="Last updated on Jun 18, 2018, 8:51 PM GMT (3c276fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/719/c555c83...3c276fb.html" title="Last updated on Jun 18, 2018, 8:51 PM GMT (3c276fb)">Diff</a>

### DIFF
--- a/respec-config.js
+++ b/respec-config.js
@@ -1,6 +1,6 @@
 var respecConfig = {
   // Embed RDFa data in the output.
-  doRDFa: '1.1',
+  doRDFa: '1.2',
   includePermalinks: true,
   permalinkEdge: true,
   permalinkHide: false,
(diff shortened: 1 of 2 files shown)


respec-config.js
```diff
@@ -30,7 +30,7 @@ var respecConfig = {
   // previousDiffURI: "",
 
   // If there a publicly available Editors Draft, this is the link
-  edDraftURI: 'https://w3c.github.io/aria-practices/',
+  edDraftURI: 'https://w3c.github.io/aria-practices-1.2/',
```
mcking65: This won't work ... the path in the gh-pages URI is the repo name. We don't have a separate repo for 1.2. The only option here is to use rawgit. Or, another possibility is to restructure the entire gh-pages branch to have a 1.1 dir and a 1.2 dir and then reconfig travis so that it puts stuff in the right dir ... hmm, but then that would mean restructuring master as well. We would have to separate dirs in master. For now, the simplest, I guess, is to use rawgit, which is not the greatest idea b/c github says not to drive public traffic to rawgit.

PR #722, opened 2018-06-18
Add multi-select tree examples

jongund:
These include two multi-select treeview examples.  One for computed and one for declared set size, position in set and level.   The two examples also differ when a file is selected.  The 1a example set aria-selected=false and the 1b example removes the aria-selected attribute all together.


mcking65:
@jongund, thank you ... lots of work here. It looks like it could be a good addition.

I searched issues, and I don't find any issues or plans related to a multi-select tree. If we're going to do this, we should have an issue where we can discuss and put it in the [tree project](https://github.com/w3c/aria-practices/projects/17).

I made a feature branch, add-multiselect-tree where we can work on this. Please edit this PR and [change the base branch](https://help.github.com/articles/changing-the-base-branch-of-a-pull-request/) from master to add-multiselect-tree.

One thing I think we should consider is combining this with one of the existing examples; we already have 4 very similar examples; that is a lot to maintain. Rather than adding another example, I think we could make the file selector tree into a multi-select without losing anything. But, first, just change the base branch, I will merge, than we can tweak more.

mcking65:
@jongund, I just figured out that I can change the base branch on any PR ... so I am doing that and merging.  We will track this project in issue #735.

PR #723, opened 2018-06-18
Editor menubar example: Make Escape close submenu when focus in menubar

jongund:
For issue #451.


mcking65:
Thank you @jongund. It is working.

Issue #125, opened 2016-10-16
Develop example of spinbutton design pattern


mcking65:
The spinbutton design pattern is at:
http://w3c.github.io/aria-practices/#spinbutton

#### Link to Work in Progress

[View the example spinbutton in the feature branch](https://rawgit.com/w3c/aria-practices/issue125-add-spinbutton-example/examples/spinbutton/spinbutton.html)

mcking65:
@gerardkcohen is working on this issue.

eknowles:
@mcking65 is this not describing an input[type=number] ? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number apart from the page up/down large step function?

mcking65:
@eknowles, yes, HTML input type number is mapped as a spin button. So, if the native html element meets your needs, do not make a custom ARIA spin button.

jongund:
Matt,

My student and I are working on a spin button example.   If that is ok with you.

Jon

mcking65:
@jongund, Gerard has already submitted #724.

mcking65:
@gerardkcohen, do you have time this week to wrap up js and css for this example? I could help out with the documentation. Wondering if we can get it done soon. I would consider delaying release 2 a few days if so. We need time on calendar for working group review.

mcking65:
When focus is in the spin button, it is capturing all key events. Can not tab out, cannot use modified keys like alt+d to go to addressbar, etc.

gerardkcohen:
Hi Matt, I am currently on vacation until 7/23. I don’t think I will have
time to get through this, but I’ll give it a try. Sorry about the timing.

Gkc

mcking65:
Closed with commit 8a3fea1.

The [new spin button example](http://w3c.github.io/aria-practices/examples/spinbutton/datepicker-spinbuttons.html) is in the editor's draft.

PR #724, opened 2018-06-19
Spinbutton - Initial commit, work in progress

gerardkcohen:
This is the initial commit for the Spinbutton issue #125, very much a work in progress.

Testing without AT (VoiceOver/ Safari), UP/DOWN will increment/ decrement as designed. With AT, VoiceOver suggests different keys for operation. I am looking for advice on how to proceed.

With VO, CTRL+OPTION+UP/DOWN is required to increment/ decrement the counter. CTRL+OPTION is not firing, so while VO is announcing the change I am not capturing any events to respond to.

cc: @mcking65

### DIFF
--- a/examples/spinbutton/js/spinbutton.js
+++ b/examples/spinbutton/js/spinbutton.js
@@ -14,7 +14,8 @@ Array.prototype.slice.call(document.querySelectorAll('.Spinbutton')).forEach(fun
 
     if (event.key.match(/[\d]|Backspace/)) {
       // updateValueNow(target,);
-    } else {
+    }
+    else {
       event.preventDefault();
     }
 
--- a/examples/spinbutton/js/spinbutton.js
+++ b/examples/spinbutton/js/spinbutton.js
@@ -1,0 +1,76 @@
+/*
+*   This content is licensed according to the W3C Software License at
+*   https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
+*
+*   Simple spinbutton pattern example
+*/
+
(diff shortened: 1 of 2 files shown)


mcking65:
@gerardkcohen,

I'd like to develop this in a feature branch before landing in master. I made branch issue125. Could you please close this and resubmit with issue125 as the base rather than master?

Also, please run npm test. There's an eslint error.

WRT the VO slider keys, it is a bug in VO that it provides that keyboard help for a web slider. Those keys only work on native sliders. Without AOM, it is not possible for us to capture VO slider increment and decrement events in JavaScript.

gerardkcohen:
@mcking65 In a bit of a bad cycle here. I am getting the error "Element input is missing one or more of the following attributes: type." when trying validate HTML.

Ok, add `type="text"` and get "Bad value spinbutton for attribute role on element input." with "Attribute aria-valuemin not allowed on element input at this point."

I can use `type="number"` and get off with a warning of "The spinbutton role is unnecessary for element input whose type is number." but then example changes with the native browser spinbutton UI.

This seems like a validator issue. @stevefaulkner, any comments?

mcking65:
Raised w3c/aria-html#111.

stevefaulkner:
@gerardkcohen will review and update where necessary

gerardkcohen:
@stevefaulkner Thanks for the quick turnaround. Since the default type for input is `text` would omitting the type also pass?

stevefaulkner:
@gerardkcohen WFM

PR #731, opened 2018-06-22
For menu button links example, collapse menu when clicking button

robfentress:
For the menu button links example, currently, with the menu expanded, clicking the button trigger does not cause the menu to collapse. It does, however, do this in the menu button actions example, and I think this is the correct behavior, especially since, without that, this means, on touch devices, there is essentially no intuitive way of collapsing the menu. The issue appears to be that line 111 of Menubutton2.js is not passed the true boolean argument, as it is on Menubutton.js. I've fixed this and it appears to work now.


mcking65:
@robfentress, looks good. Thanks!

PR #740, opened 2018-06-25
Issues 451 595 menubar navigation

jongund:
This pull request addresses issues 451 and 595.

Also removed a console.log commands used for debugging.


mcking65:
@jongund, excellent ... escape as working perfectly! The only change I made was to remove the .gitconfig file.

PR #742, opened 2018-06-25
Minimal framework for regression tests of APG Examples

spectranaut:
Hi Matt and everyone,

This is the basic framework of the regression tests, using selenium's webdriverjs and avajs test framework. Regressions test for each example page will be added to the `test/tests/` directory, which only contains tests for link/link.js for now!
 
To run the tests:
```$ npm run regression```

About `tests/test/link.js`: This is a tentative format with minimal abstractions for now, I'll write tests for a few more examples (specifically, we discussed `combobox/aria1.1pattern/listbox-combo.html` and `grid/LayoutGrids.htm`) before deciding which abstraction to make. The test should closely follow the described behavior in the "Keyboard Support" and "Role, Property, State, and Tabindex Attributes" tables of the `link/link.js` example page.

### CODE REVIEW

File: test/index.js [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+
+test.beforeEach((t) => {
+  t.context.session = session;
+  t.context.By = webdriver.By;
+  t.context.Key = webdriver.Key;
+  t.context.until = webdriver.until;
+});
+
+test.after.always(() => {
+  return Promise.resolve(session && session.close())
```
jessebeach: Add a catch here?

test/tests/link.js
```diff
@@ -0,0 +1,107 @@
...
+
+ariaTest('link/link.html', 'link-role', async (t) => {
+
+  for (let i = 0; i < pageExamples.length; i++) {
+    let ex = pageExamples[i];
+    let linkLocator = t.context.By.css(ex.linkSelector);
+    let linkElement = await t.context.session.findElement(linkLocator);
+
+    t.is(
+      await linkElement.getAttribute('role'),
```
jessebeach: Rather than a raw await here, can we provide a utility function that affords an error boundary for unresolved promises? 

I'm thinking ahead to future maintenance. Unresolved promise errors are a devil to debug, so we can catch them at a boundary and emit a human-readable warning, at least we might know where to start looking. Something to ponder, at least. It may not be an ideal approach in this case.

test/util/start-geckodriver.js
```diff
@@ -0,0 +1,80 @@
...
+
+    child.on('close', giveUp);
+
+    (function poll () {
+      if (stopPolling) {
+        return;
+      }
+
+      if (timeout - (Date.now() - start) < 0) {
+        reject(new Error('Timed out while waiting for WebDriver server'));
```
jessebeach: 👍

test/util/start-geckodriver.js
```diff
@@ -0,0 +1,80 @@
...
+      return { stop, port };
+    });
+};
+
+/**
+ * Start a GeckoDriver server using a dynamically-determined available TCP/IP
+ * port.
+ *
+ * @param {Number} port - the TCP/IP port from which to begin search
+ * @param {Number{ timeout - the number of milliseconds to attempt to create a
```
jessebeach: Wrong-wase turned bracket.

test/index.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {String} page - path to the example file
+ * @param {String} testId - unique identifier for the documented behavior
+ *                          within the demonstration page
+ * @param {Function} body - script which implements the test
+ */
+const ariaTest = (page, testId, body) => {
+  const absPath = path.resolve(__dirname, '..', 'examples', ...page.split('/'));
+  const url = 'file://' + absPath;
+  const selector = '[data-test-id="' + testId + '"]';
```
mcking65: I like this association between the documentation and the test ... very nice!

test/index.js
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Declare a test for a behavior documented on and demonstrated by an
+ * aria-practices examples page.
+ *
+ * @param {String} page - path to the example file
+ * @param {String} testId - unique identifier for the documented behavior
+ *                          within the demonstration page
+ * @param {Function} body - script which implements the test
+ */
+const ariaTest = (page, testId, body) => {
```
sh0ji: As the primary interface for writing tests, I'd prefer it if this function did as little as necessary even if it means more code repetition inside the tests themselves. This will help keep parity with the Ava API, and make the tests more readable as discretely composed units.

test/tests/link.js
```diff
@@ -0,0 +1,107 @@
...
+      'link',
+      '[role="link"] attribute should exist on element select by: ' + ex.linkSelector
```
sh0ji: Placing the assertion message here instead of in the test title means that a human-readable description of the test is only ever output if it fails. I'd prefer tests to always describe themselves.

examples/link/link.html
```diff
@@ -113,7 +113,7 @@ <h2 id="kbd_label">Keyboard Support</h2>
         </tr>
       </thead>
       <tbody>
-        <tr>
+        <tr data-test-id="key-enter">
```
sh0ji: I like that you're using the tables found in the examples to inform the tests. We'll want to keep a close eye on whether that's sufficient.

test/index.js
spectranaut: I hear where you are coming from -- I prefer readability to DRYness in tests myself. I wasn't particularly expecting to add much more functionality to ariaTest, but to be clear, do you think the subtest that currently exists in the ariaTest interface doesn't belong there, or are you just stating a preference not to see ariaTest grow too much?

test/tests/link.js
spectranaut: Also I want to be sure I understand what you are saying here -- the test title here is quick simple, it's `link/link.html [data-test-id="link-role"]`, constructed in ariaTest. The purpose that title is to tie  the description of the test to the description of the attribute/role/keypress that exists in the link/link.html example page itsself.

It sounds like you prefer the title of the test to be more verbose than this. What about titles like: `Test the 'role' attribute exists as documented in link/link.html [data-test-id="link-role"]`

test/index.js
sh0ji: My concern here is because maintainers/users will primarily only ever be interacting with the unit tests, so they should ideally be able to understand everything that's happening from there without ever having to dig into ariaTest's internals.

One option would be to require ariaTest to be instantiated, similar to Webdriver's initialization process, and do the abstractions as part of that setup. Each page would require a declarative setup, but that can help with readability and DRYness.

test/tests/link.js
sh0ji: The first part of the title is great. I just meant that the actual description of what the test is testing only appears if the test fails. For example:
```bash
# success
  √ link » link/link.html [data-test-id="link-role"]
# failure
  × link » link/link.html [data-test-id="link-role"] [role="link"] attribute should exist on element select by: #ex1 span
```

PR #744, opened 2018-06-26
Form examples to the form input assistance section 

jongund:
There are two examples, one for ordering a pizza and the other for inputing a credit card.


.gitconfig
```diff
@@ -0,0 +1,3 @@
+[branch]
```
mcking65: Jon, we don't want this file in the shared repository. You can set your local config with the git config command ... read up on the options for that.

mcking65:
@jongund, there is 1 eslint error.

.gitconfig
jongund: Ok.  I have removed this file from my fork.

jongund:
I can update the example to be one form

mcking65:
@jongund, there are validation errors that need to be fixed. Note that you can push more commits to this same branch on your fork and they will be added to this PR automatically; you do not need to open another PR.

jongund:
Validation errors should be fixed now.

PR #754, opened 2018-07-04
link/link.js feedback - minor fixes

spectranaut:
Fix testnames based on @sh0ji 's suggestions and fix one small comment.

I left the ariaTest abstraction for now (including the check for data-test-id) -- we can revisit it later. I did change the "t.is" in ariaTest to an assert in order to not effect the test count for test planning.

### DIFF
--- a/examples/link/link.html
+++ b/examples/link/link.html
@@ -147,7 +147,7 @@
             </ul>
           </td>
         </tr>
-        <tr data-test-id="tab-index">
+        <tr data-test-id="tabindex">
           <td></td>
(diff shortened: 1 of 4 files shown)


mcking65:
@sh0ji, would like your input on latest revisions; do they address your feedback?

sh0ji:
Yes, this looks great!

My only lingering question is whether ariaTest's `page` argument is likely to ever be different within the scope of a test file. If not, it might be worth making it configurable (e.g. `ariaTest.page('link/link.html')`).

PR #755, opened 2018-07-05
Work in progress: add tests for LayoutGrid

jugglinmike:
(This depends on gh-754.)

For this series of tests, we wanted to push the envelope on the idea of flexibility.

The statements made in the test examples document are fairly generic, describing the expected functionality for all cells and focusable elements. To ensure these are honored, this suite of tests are highly dynamic: they query the document and make assertions according to the structure they find.

A bug in code like this could cause entire branches to be skipped. That is particularly dangerous in test code because it can lead to false positives (since it can be difficult to differentiate between passing assertions and mistakenly-skipped assertions). In order to avoid false positives in such cases, we are using the test framework's `expect` method: this gives us confidence that the procedural code is operating as intended (at the expense of some maintenance overhead--modifications to the test may require manually updating those numbers).

Neither @spectranaut nor I feel as though this is a tenable approach to testing. Beyond being labor-intensive to write, these tests are difficult to understand, and that diminishes their value as documentation.

More concerning is their failure mode. Although we attempted to account for future alterations to the examples, we can't guarantee that the tests will be resilient to any arbitrary change. In the event that a contributor attempts to modify an example in a way that violates a test, that contributor will be forced to learn the complexities of this logic and to extend it.

We're submitting tests in this form for review in order to demonstrate those pain points and to verify that simpler patterns are preferable. Specifically: we would prefer to write tests that take the current structure of the examples for granted. Tests written in this way will contain more duplication, but they will also contain less abstraction and less conditional logic. That will make them easier for others to read and to modify. They may be more "brittle" (i.e.  more susceptible to invalidation by future changes to the examples), but correcting a more-or-less declarative list of assertion statements is a far more reasonable task for drive-by contributors.

### DIFF
--- a/examples/grid/LayoutGrids.html
+++ b/examples/grid/LayoutGrids.html
@@ -455,7 +455,7 @@
         </tr>
       </thead>
       <tbody>
-        <tr>
+        <tr data-test-id="key-right-arrow">
           <th>
             <kbd>Right Arrow</kbd>
           </th>
@@ -468,7 +468,7 @@
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="key-left-arrow">
           <th>
             <kbd>Left Arrow</kbd>
           </th>
@@ -481,7 +481,7 @@
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="key-down-arrow">
           <th>
             <kbd>Down Arrow</kbd>
           </th>
@@ -496,7 +496,7 @@
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="key-up-arrow">
           <th>
             <kbd>Up Arrow</kbd>
           </th>
@@ -511,7 +511,7 @@
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="key-page-down">
           <th>
             <kbd>Page Down</kbd> (Example 3)
           </th>
@@ -523,7 +523,7 @@
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="key-page-up">
           <th>
             <kbd>Page Up</kbd> (Example 3)
           </th>
@@ -535,25 +535,25 @@
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="key-home">
           <th>
             <kbd>Home</kbd>
           </th>
           <td>Moves focus to the first cell in the row that contains focus.</td>
         </tr>
-        <tr>
+        <tr data-test-id="key-end">
           <th>
             <kbd>End</kbd>
           </th>
           <td>Moves focus to the last cell in the row that contains focus.</td>
         </tr>
-        <tr>
+        <tr data-test-id="key-control-home">
           <th>
             <kbd>Control + Home</kbd>
           </th>
           <td>Moves focus to the first cell in the first row.</td>
         </tr>
-        <tr>
+        <tr data-test-id="key-control-end">
           <th>
             <kbd>Control + End</kbd>
           </th>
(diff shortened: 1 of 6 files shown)


examples/grid/LayoutGrids.html
```diff
@@ -641,13 +641,13 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
             </ul>
           </td>
         </tr>
-        <tr>
+        <tr data-test-id="gridcell-role">
           <th scope="row"><code>gridcell</code></th>
           <td></td>
-          <td><code>span</code></td>
+          <td><code>span, div</code></td>
```
spectranaut: I wanted to point on this minor change -- another place where the table was out of date with the examples. "tab-index" can be found on both span and div elements.

test/tests/LayoutGrids.js
```diff
@@ -0,0 +1,593 @@
...
+  'grid/LayoutGrids.html', 'grid-role', async (t) => {
+
+    t.plan(3);
+
+    for (let exId in pageExamples) {
+      let ex = pageExamples[exId];
+      let gridLocator = t.context.By.css(ex.gridSelector);
+
+      t.is(
+        (await t.context.session.findElements(gridLocator)).length,
```
spectranaut: It's not clear to me what you would like test for "role" attributes -- should I simply be testing whether an appropriate element type with an appropriate role exists, as I have in these tests for `role=grid`, `row` and `gridcell`?

test/tests/LayoutGrids.js
```diff
@@ -0,0 +1,593 @@
...
+  for (let [exId, selector] of Object.entries(cellSelectors)) {
+    let gridcellElements = await t.context.session.executeScript(findFocusable, selector);
+
+    // Find the first focusable element
+    let activeElement = await t.context.session.executeScript(focusWithin, gridcellElements[gridcellElements.length - 1]);
+    if (!activeElement) {
+      throw new Error('Could not focus on element or any decendent in the first gridcell: ' + selector);
+    }
+
+    // Test focus moves to next element on arrow key
```
tatermelon: previous element

test/tests/LayoutGrids.js
```diff
@@ -0,0 +1,593 @@
...
```
tatermelon: previous element

test/tests/LayoutGrids.js
```diff
@@ -0,0 +1,593 @@
...
+
+ariaTest('PageDown key moves focus', 'grid/LayoutGrids.html', 'key-page-down', async (t) => {
+  t.plan(12);
+
+  const cellSelectors = {
+    first: '#ex3 [role="row"] [role="gridcell"]:nth-child(1)',
+    second: '#ex3 [role="row"] [role="gridcell"]:nth-child(2)',
+    third: '#ex3 [role="row"] [role="gridcell"]:nth-child(3)'
+  };
+  const jumpBy = 5;
```
tatermelon: You could use the data-per-page value here

test/tests/LayoutGrids.js
```diff
@@ -0,0 +1,593 @@
...
+        correctActiveElement,
+        'Example ' + exId + ': gridcell `' + (await gridcellElements[index].getText()) + '` should receive focus after arrow right'
+      );
+
+      activeElement = await t.context.session.executeScript(() => {
+        return document.activeElement;
+      });
+    }
+
+    // Test arrow key on final element
```
tatermelon: Here you can use the data-wrap-cols attribute to check whether or not the arrows should wrap

examples/grid/LayoutGrids.html
mcking65: I've addressed this in master.

mcking65:
@jugglinmike, after merging linktest changes, we now have some conflicts.

spectranaut:
blocked on merge of https://github.com/w3c/aria-practices/pull/766

So I can merge master into bocoup and add submit the last few tests on this branch for review!

spectranaut:
Closing -- replaced with PR: #784 
Reviews all addressed there!

Issue #132, opened 2016-10-17
Develop example of tree grid design pattern


mcking65:
This issue depends on issue #91.

### URL for Viewing Development Branch

This example is being developed in the treegrid branch. View it at:
https://rawgit.com/w3c/aria-practices/treegrid/examples/treegrid/treegrid-1.html

aleventhal:
Here's an example I coded up:
https://github.com/aleventhal/treegrid-example

aleventhal:
Better example, with more discoverable keyboard model:
https://github.com/aleventhal/treegrid-tab-active-cells

mcking65:
Current work on this example is now in the treegrid branch of w3c/aria-practices at:
https://rawgit.com/w3c/aria-practices/treegrid/examples/treegrid/treegrid-row-nav-primary-1.html

mcking65:
In pull request #710, @jongund is working on modifying the example in the treeggrid branch, which is at:
https://rawgit.com/w3c/aria-practices/treegrid/examples/treegrid/treegrid-1.html
to complete the example documentation using the standard example template.

Jon's work can be previewed at:
https://rawgit.com/jongund/aria-practices/treegrid/examples/treegrid/treegrid-1.html

aleventhal:
@mcking65 I believe aria-hidden is just used for the collapsed rows which are hidden. Just semantically indicating that they are in fact hidden.

mcking65:
@aleventhal commented:

>I believe aria-hidden is just used for the collapsed rows which are hidden. Just semantically indicating that they are in fact hidden.

Thanks ... It sounds like we should remove it then.

aleventhal:
@mcking65 Why is that? Does it work better with the virtual buffer?

mcking65:
@aleventhal, no, just because it does not do anything; it is redundant code. 

In order to avoid the risk of aria-hidden unintentionally hiding things in the event it were somehow to get out of sync, we have adopted the practice of avoiding using aria-hidden at all unless it is necessary to hide something from AT that is visually rendered -- a practice we also avoid like the plague.

jongund:
I can update the code to not use aria-hidden, will use the class name "hidden".

jongund:
I have updated the code to remove the use of aria-hidden attribute and should be available as part of pull request 710

annabbott:
HTML Source Code > HTML Source Code > aria-level="3" > aria-posinset="1" currently is:
[...]

PR #762, opened 2018-07-11
Add Treegrid Example

mcking65:
For issue #132, merging review-ready version of treegrid example developed by @aleventhal and edited to fit example template by @jongund.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/762.html" title="Last updated on Jul 12, 2018, 6:51 AM GMT (514f612)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/762/a49121c...514f612.html" title="Last updated on Jul 12, 2018, 6:51 AM GMT (514f612)">Diff</a>


examples/treegrid/treegrid-1.html
```diff
@@ -0,0 +1,567 @@
...
+    <div id="ex1">
+      <table id="treegrid" role="treegrid" aria-label="ARIA treegrid e-mail inbox example">
+        <colgroup>
+          <col id="treegrid-col1">
+          <col id="treegrid-col2">
+          <col id="treegrid-col3">
+        </colgroup>
+        <thead>
+          <tr>
+            <th role="columnheader">Subject</th>
```
mcking65: @aleventhal, @jongund, why is role columnheader added here. This is not necessary and not recommended to duplicate native semantics. Can we safely remove without breaking any selectors or js?

examples/treegrid/treegrid-1.html
```diff
@@ -0,0 +1,567 @@
...
+            <th role="columnheader">Summary</th>
```
mcking65: @aleventhal, @jongund, why is role row added here and on subsequent tr elements? This is not necessary and not recommended to duplicate native semantics. Can we safely remove without breaking any selectors or js?

examples/treegrid/treegrid-1.html
aleventhal: Actually, a th can be a columnheader or a rowheader. I don't think it's redundant to specify. This way the browser does not need to fall back on heuristics (which different browsers may disagree about).

aleventhal: Actually we can probably use scope="col".

examples/treegrid/treegrid-1.html
aleventhal: Good question. I would guess that it makes no difference and can be removed, but I think it should be tested with at least one screen reader to be completely sure.

FWIW, I find it easier to understand the examples when the roles are explicitly stated.  When we remove "row" from the markup, should we also make sure it's clear in the docs that <tr> can be replaced with something like <div role="row"> ?

mcking65:
OK ... Editorially in reasonable shape and the only errors left are the posinset and setsize errors, which are being addressed by w3c/aria#558. So, merging to master.

PR #766, opened 2018-07-13
Changes to example page html for listbox-combo.html and LayoutGrid.html

spectranaut:
Hi @mcking65 -- per our discussion, here is a PR to master with the changes to the example pages themselves. This commit only add "data-test-id" to the appropriate tables.

### DIFF
--- a/examples/combobox/aria1.1pattern/listbox-combo.html
+++ b/examples/combobox/aria1.1pattern/listbox-combo.html
@@ -182,7 +182,7 @@
           </tr>
         </thead>
         <tbody>
-          <tr data-test-id="key-down-arrow">
+          <tr data-test-id="textbox-key-down-arrow">
             <th><kbd>Down Arrow</kbd></th>
             <td>
               <ul>
@@ -196,7 +196,7 @@
               </ul>
             </td>
           </tr>
-          <tr data-test-id="key-up-arrow">
+          <tr data-test-id="textbox-key-up-arrow">
             <th><kbd>Up Arrow</kbd></th>
             <td>
               <ul>
@@ -219,7 +219,7 @@
               </ul>
             </td>
           </tr>
-          <tr data-test-id="key-escape">
+          <tr data-test-id="textbox-key-escape">
             <th><kbd>Escape</kbd></th>
             <td>
               <ul>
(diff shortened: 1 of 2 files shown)


examples/combobox/aria1.1pattern/listbox-combo.html
```diff
@@ -274,7 +274,7 @@ <h3 id="kbd_label_listbox">Listbox Popup</h3>
               </ul>
             </td>
           </tr>
-          <tr>
+          <tr data-test-id="key-down-arrow">
```
mcking65: This is same ID as line 185, but won't the tests be different?

examples/combobox/aria1.1pattern/listbox-combo.html
```diff
@@ -264,7 +264,7 @@ <h3 id="kbd_label_listbox">Listbox Popup</h3>
               </ul>
             </td>
           </tr>
-          <tr>
+          <tr data-test-id="key-escape">
```
mcking65: This is same ID as line 222, but won't tests be different?

examples/combobox/aria1.1pattern/listbox-combo.html
```diff
@@ -284,7 +284,7 @@ <h3 id="kbd_label_listbox">Listbox Popup</h3>
               </ul>
             </td>
           </tr>
-          <tr>
+          <tr data-test-id="key-up-arrow">
```
mcking65: This is same ID as line 199, but won't tests be different?

examples/combobox/aria1.1pattern/listbox-combo.html
spectranaut: you are right, and it lead to me to not test this row of this table. Changing this to: `listbox-key-down-arrow` and the other to `textbox-key-down-arrow`

examples/combobox/aria1.1pattern/listbox-combo.html
spectranaut: Changing this to: `listbox-key-escape` and the other to `textbox-key-escape`

examples/combobox/aria1.1pattern/listbox-combo.html
spectranaut: Changing this to: `listbox-key-up-arrow` and the other to `textbox-key-up-arrow`

spectranaut:
Hey @mcking65 addressed your feedback. I'll have to add a few more tests to listbox-combo.js -- so good catch. I'll add that to the backlog issue #780

mcking65:
looks good

PR #791, opened 2018-07-19
Add data-test-ids for next set of examples

spectranaut:
Hi @mcking65 -- thought I'd get a week ahead on the data-test-ids as you will be out all of next week :) If you could review/merge before end of day Friday that would be great.


mcking65:
Looks good.

Issue #794, opened 2018-07-22
single-select listbox is not setting `aria-selected`


tryenc:
None of the single-selection listboxes are setting `aria-selected` on an `option` when it receives focus.

mcking65:
Resolved with commit 66229af.

PR #795, opened 2018-07-22
Single-selection listbox example: Fix aria-selected bug

tryenc:
For issue #794, update aria-selected for single-selection listbox options on focus change in listbox/js/listbox.js:
* When not multi-selection, toggling aria-selected when option receives or loses focus.

### DIFF
--- a/examples/listbox/css/listbox.css
+++ b/examples/listbox/css/listbox.css
@@ -35,17 +35,13 @@
 
 [role="option"] {
 	display: block;
-	padding: 0 1em;
+	padding: 0 1em 0 1.5em;
 	position: relative;
 	line-height: 1.8em;
 }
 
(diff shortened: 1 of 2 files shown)


mcking65:
@tryenc, thank you for catching this and submitting a fix.

I also agree with the change to use the checkmark to indicate the selected option in the siggle selects. If I correctly understand the styling, previously the selected item was being identified only by color.

Since there are visual CSS changes, I'd like someone who can see to review. @sh0ji, can you have a look at the CSS changes in this?

mcking65:
Thank you @sh0ji, that css change makes sense to me now.

mcking65:
@spectranaut, you will want to merge this into bocoup before doing listbox tests.

PR #801, opened 2018-07-24
Add tests for treeview/treeview-1/treeview-1a.html

spectranaut:


test/tests/treeview-1a.js
```diff
@@ -0,0 +1,675 @@
...
+    return el.parentElement.getAttribute('role') == 'tree';
+  }, el);
+};
+
+const isFolderTreeitem = async function (el) {
+  return !(await el.getAttribute('class')).includes('doc');
+};
+
+const isOpenedFolderTreeitem =  async function (el) {
+  return await el.getAttribute('aria-expanded') === 'true';
```
sh0ji: Async functions are already wrapped in a `Promise.resolve`, so there's no need to `return await` here (or similar). And since I've never actually tested this, it looks like `return await` actually adds some overhead:

```javascript
const first = async () => await 'first';
const second = async () => 'second';
first().then(console.log);
second().then(console.log);
// second
// first
```

test/tests/treeview-1a.js
```diff
@@ -0,0 +1,675 @@
...
+  docSelector: '#ex1 .doc[role="treeitem"]',
+  textboxSelector: '#ex1 #last_action'
+};
+
+const openAllFolders = async function (t) {
+  const closedFoldersSelector = ex.treeitemSelector + '[aria-expanded="false"]';
+  let folders = await t.context.session.findElements(By.css(closedFoldersSelector));
+
+  // While their are closed folders
+  while (folders.length != 0) {
```
jessebeach: `>`, just so we don't get any weird falsey comparisons in the future. For example, `undefined != 0` is `true`, which will yield a lot of looping :)

test/tests/treeview-1a.js
```diff
@@ -0,0 +1,675 @@
...
+
+    // Find the first visible closed folder and click it
+    for (let folder of folders) {
```
jessebeach: This loop in a loop is a little difficult to parse, especially given the original assignment of folders and the subsequent re-assignment in the loop is the same expression:

```
await t.context.session.findElements(By.css(closedFoldersSelector));
await t.context.session.findElements(By.css(closedFoldersSelector));
```

Perhaps I just don't understand the intent of the logic, so either a comment would elucidate or maybe a stack approach would be more clear to avoid the loop in a loop?

test/tests/treeview-1a.js
```diff
@@ -0,0 +1,675 @@
...
+        break;
+      }
+    }
+
+    folders = await t.context.session.findElements(By.css(closedFoldersSelector));
+  }
+};
+
+const checkFocus = async function (t, selector, index) {
+  return await t.context.session.executeScript(function (/* selector, index*/) {
```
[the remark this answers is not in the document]
jessebeach: @sh0ji, if a function contains an `await`, it itself must be async.

test/tests/treeview-1a.js
```diff
@@ -0,0 +1,675 @@
...
+  }, selector, index);
+};
+
+const checkFocusOnParentFolder = async function (t, el) {
+  return await t.context.session.executeScript(function () {
+    const el = arguments[0];
+
+    // the element is a folder
+    if (el.hasAttribute('aria-expanded')) {
+      return document.activeElement == el.parentElement.closest('[role="treeitem"][aria-expanded]');
```
jessebeach: `===` strict equals everywhere so we don't inadvertently introduce truthy comparisons of existence rather than reference equivalence.

test/tests/treeview-1a.js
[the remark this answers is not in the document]
sh0ji: Right, I meant that the whole async/await pattern isn't necessary if its sole purpose is to make `checkFocus` thenable (async).

This should be equivalent unless I'm missing something (entirely possible):
```javascript
const checkFocus = function (t, selector, index) {
  return t.context.session.executeScript(/*... */) ;
}
// resolves to the Promise returned by `executeScript`
```

sh0ji:
Let's address the comments in a follow up.

test/tests/treeview-1a.js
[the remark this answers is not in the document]
spectranaut: good catch. it was an oversight!

test/tests/treeview-1a.js
spectranaut: This logic is unnecessarily complex. I can pair it down a lot, thanks for the catch :)

PR #804, opened 2018-07-25
Address feedback for test/tests/treeview-1a.js

spectranaut:
Addressed feedback from #801 -- @jessebeach @sh0ji

### DIFF
--- a/test/tests/treeview-1a.js
+++ b/test/tests/treeview-1a.js
@@ -23,15 +23,10 @@ const openAllFolders = async function (t) {
   const closedFoldersSelector = ex.treeitemSelector + '[aria-expanded="false"]';
   let closedFolders = await t.context.session.findElements(By.css(closedFoldersSelector));
 
-  // While there are closed folders
-  while (closedFolders.length > 0) {
-
-    // Find the first visible closed folder and click it
-    await closedFolders[0].click();
-
-    // Get the updated list of closed folders (as opening one folder may have revealled
-    // more folders)
-    closedFolders = await t.context.session.findElements(By.css(closedFoldersSelector));
+  // Going through all closed folder elements in dom order will open parent
+  // folders first, therefore all child folders will be visible before clicked
+  for (let folder of closedFolders) {
+    await folder.click();
   }
 };
 
--- a/test/tests/treeview-1a.js
+++ b/test/tests/treeview-1a.js
@@ -21,25 +21,22 @@ const ex = {
 
 const openAllFolders = async function (t) {
   const closedFoldersSelector = ex.treeitemSelector + '[aria-expanded="false"]';
-  let folders = await t.context.session.findElements(By.css(closedFoldersSelector));
+  let closedFolders = await t.context.session.findElements(By.css(closedFoldersSelector));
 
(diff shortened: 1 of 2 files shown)


test/tests/treeview-1a.js
```diff
@@ -21,50 +21,47 @@ const ex = {
...
-        await folder.click();
-        break;
-      }
-    }
+    await closedFolders[0].click();
 
-    folders = await t.context.session.findElements(By.css(closedFoldersSelector));
+    // Get the updated list of closed folders (as opening one folder may have revealled
+    // more folders)
+    closedFolders = await t.context.session.findElements(By.css(closedFoldersSelector));
```
jessebeach: You should be able to do this:

```
const childClosedFolders = await folder.findElements(By.css(closedFoldersSelector));
closedFolders = closedFolders.concat(childClosedFolders);
```

test/tests/treeview-1a.js
```diff
@@ -21,50 +21,47 @@ const ex = {
...
+  while (closedFolders.length > 0) {
 
     // Find the first visible closed folder and click it
-    for (let folder of folders) {
-      if (await folder.isDisplayed()) {
-        await folder.click();
-        break;
-      }
-    }
+    await closedFolders[0].click();
```
jessebeach: If `folder.findElements` works, then try this, combined with my comment below:

```
const folder = closedFolders.pop();
await folder.click();
```

test/tests/treeview-1a.js
jessebeach: This way, you're just finding new closed folders inside the one you just opened, rather than gathering the entire list of closed folders on the page in each iteration and replacing the existing list.

sh0ji:
@spectranaut, do you prefer to address this kind of feedback in a follow-up PR like we've been doing, or would you prefer to address things like this here? It doesn't look like this is blocking any slated work, so I'll wait for your response.

spectranaut:
It's not blocking @sh0ji thanks for checking, I'll follow up with the feedback on this PR!

test/tests/treeview-1a.js
spectranaut: Actually, it's a bit foolish -- but I don't need to and can't appended, the selector used to find closed folders finds all closed folder dom elements regardless of whether they are visible. Which makes me realize.. the whole thing should be a lot simpler :/

test/tests/treeview-1a.js
```diff
@@ -23,15 +23,10 @@ const openAllFolders = async function (t) {
...
-    await closedFolders[0].click();
-
-    // Get the updated list of closed folders (as opening one folder may have revealled
-    // more folders)
-    closedFolders = await t.context.session.findElements(By.css(closedFoldersSelector));
+  // Going through all closed folder elements in dom order will open parent
+  // folders first, therefore all child folders will be visible before clicked
+  for (let folder of closedFolders) {
+    await folder.click();
   }
```
spectranaut: tada :)

jessebeach: @sh0ji ship it.

PR #806, opened 2018-07-26
tests for dialog-modal/dialog.html

spectranaut:


spectranaut:
Ready for review, @sh0ji and @jessebeach 

Aside -- this one took me longer than I predicted, I gave it an "easy" (half day) at first glance, it took over a full day. The first way I solve it was too complicated and too hard to read so I scrapped it. I hope this is good!

sh0ji:
Fixed a quotes conflict, but it otherwise looks good! Merging.

Issue #813, opened 2018-07-31
Bug in example listbox/listbox-scrollable.html: "aria-selected" never set on options


spectranaut:
The documentation of the [listbox-scrollable](https://w3c.github.io/aria-practices/examples/listbox/listbox-scrollable.html) example claims the `aria-selected` attribute should be set on the option that is visually selected, but it is never set.

spectranaut:
This bug was actually already reported and fixed by https://github.com/w3c/aria-practices/pull/795 -- thanks @tryenc !

PR #818, opened 2018-08-01
Updates to aria-activedescendant util function

spectranaut:
When noticing #813 in the example `listbox/listbox-scrollable.html` example, I realized that setting the "aria-selected" attribute on the appropriate option is not strictly a part of the aria-activedescendant pattern. I renamed this util function to specify that it is checking for both the aria-activedescendant pattern and a corresponding aria-selected attribution.

### DIFF
--- a/test/tests/listbox-combo.js
+++ b/test/tests/listbox-combo.js
@@ -4,7 +4,7 @@ const { ariaTest } = require('..');
 const { By, Key } = require('selenium-webdriver');
 const assertAttributeValues = require('../util/assertAttributeValues');
 const assertAriaLabelledby = require('../util/assertAriaLabelledby');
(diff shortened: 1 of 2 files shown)


spectranaut:
should be straightforward review @tatermelon

mcking65:
@spectranaut, I just want to confirm that testing for selected is or can be independent of testing for activedescendant? If we have an example that uses activedescendant and it implements a pattern where selection does not follow focus, that would be important. For instance, if the tabs example where tabs are manually activated were implemented with activedescendant instead of roving tabindex, we would need that.

spectranaut:
hi @mcking65 right -- this util is only for testing examples which use both aria-selected and aria-activedescendent, and the purpose of this PR is to make that more clear. So far I haven't encountered an activedescendent (without selected) example but I'll make probably make a util function just for testing that once I find one :)

PR #779, opened 2018-07-17
Expand VNU validator file pattern

nschonni:
Realized that the original wasn't catching all subfolders. Fixed up the issues that I understood, but there are a few combobox examples using `rel` on `ul` and `li` elements that I don't understand enough to fix. EX: 
```
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":57.11-57.63: error: Attribute “rel” not allowed on element “ul” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":67.13-67.77: error: Attribute “rel” not allowed on element “li” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":68.13-68.79: error: Attribute “rel” not allowed on element “li” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":71.15-71.72: error: Attribute “rel” not allowed on element “ul” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":79.15-79.82: error: Attribute “rel” not allowed on element “ul” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":90.11-90.67: error: Attribute “rel” not allowed on element “ul” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":101.13-101.73: error: Attribute “rel” not allowed on element “li” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":102.13-102.73: error: Attribute “rel” not allowed on element “li” at this point.
"file:/Users/nschonni/Documents/github/aria-practices/examples/menubar/menubar-2/menubar-2.html":105.15-105.71: error: Attribute “rel” not allowed on element “ul” at this point.
```
These can be ignored in the `.vnurc`, but I wanted to see if it was intentional first

PR #820, opened 2018-08-01
Fix vnu examples markup

nschonni:
Split the markup fixes from #779 per @mcking65


mcking65:
Many, many  thanks @nschonni!

Issue #704, opened 2018-06-05
Spinbutton: What should aria-valuenow be when a user gives invalid input, when should it update?


0ddfell0w:
We're developing an example ARIA spin button, and we're hoping for a clarification around aria-valuenow.

From the [APG](https://www.w3.org/TR/wai-aria-practices-1.1/#spinbutton):
> "The slider element has the aria-valuenow property set to a decimal value representing the current value of the slider."

What should the value of aria-valuenow should be when a user (for example) pastes in some invalid input. Should aria-valuenow simply not change? Should it be set back to some default when the spinbutton loses focus? Should the paste be undone by reverting to the previous value?

Thanks!

0ddfell0w:
Addendum: What should aria-valuenow be when the user deletes the text in the spinbutton? If it should be the empty string, can we adjust the following language to include a reference to that case? 
>  "The spinbutton element has the aria-valuenow property set to a decimal value representing the current value of the spinbutton."

jnurthen:
(this is not a WG response - just some additional information)

The native spinbutton in Firefox does not use valuenow but it does expose valuetext.
When an invalid value is specified valuetext is set to ""
When the value is deleted from the spinbutton valuetext also gets set to ""

IMO I would envision valuenow working in the same way to mozilla's valuetext implementation.

0ddfell0w:
Firefox also adds an invalid state when the input has either
+ a non-numerical value
+ a value outside of the range specified by min and max.

Chrome also adds an invalid state when the value is outside the range. However, Chrome does more to prevent non-numerical input. The input filters out the non-numerical characters and writes the numerical characters (including hyphens, periods, and even the letter e for scientific notation)

Would the WG consider adding instructions to set aria-invalid on spinbuttons when the value is invalid, either because it's nonsense or because it's outside of the specified range?

mcking65:
@0ddfell0w, The APG TF discussed this in [today's teleconference](https://www.w3.org/2018/08/06-aria-apg-minutes.html).

The aria-valuenow, and if used, aria-valuetext, attribute should represent the current state of the control. That is, whatever is displayed should be reflected by the aria value attributes even if the displayed value is invalid. If the displayed value is invalid, then aria-invalid should be set.

Note that some spinbuttons are implemented so that inputting invalid values is not possible. For instance, numeric inputs will not allow letters to be typed.

So, the answers to your questions:

[...]

PR #825, opened 2018-08-06
Sppinbutton Pattern: Add aria-invalid guidance for values outside allowed range

mcking65:
In response to issue #704, modified the roles, states, and properties section of the spinbutton pattern in aria-practices.html.

Adds a bullet to the list of roles, states, and properties stating that aria-invalid is set true if it is possible for the textbox to contain a value that is not allowed.

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -2269,6 +2269,10 @@
             Otherwise, the spinbutton element has a label provided by
             <a href="#aria-label" class="property-reference">aria-label</a>.
           </li>
+          <li>
+            If it is possible to modify the value of the textbox such that the value is outside of the allowed set of values, the spinbutton element has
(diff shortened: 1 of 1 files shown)


0ddfell0w:
Thanks for the change, my only note is that the wording could be misconstrued (with some effort) to be saying "if an invalid state is possible then set aria-invalid to true" as opposed to "For spinbuttons that allow users to enter invalid inputs, set aria-invalid to true when the value is invalid."

charmarkk:
This looks good to me Matt! -Mark

MichielBijl:
How about something like this:

Most spinbutton implementations will prevent the input of invalid values. But, in some cases blocking invalid input might not be practical. In such scenarios aria-invalid is set to true if the value is outside of the allowed range.

This is pretty close to how I’d write it in an article or some such. I think I kept the information the same and only changed the tone slightly.

PR #832, opened 2018-08-14
Tests for listbox/listbox-scrollable.html

spectranaut:
I wrote these tests two weeks ago but they were failing because of bug now fixed by #795 -- nice to see them passing now :)

Ready for review -- @mcking65 which reviewer is currently not on vacation?


nschonni:
#445 landed so if you restart the build you should get CI builds now

mcking65:
@tatermelon, @jessebeach, @sh0ji, are one of you available fore review?

PR #834, opened 2018-08-15
Spelling fixes

nschonni:
Ran cspell to fix up some typos. Not entirely done, but I'll try and add it to the build with the config separately

### DIFF
--- a/examples/menubar/menubar-1/mb-outreach.html
+++ b/examples/menubar/menubar-1/mb-outreach.html
@@ -18,7 +18,7 @@
         <h2 id="comm-engage">Community Engagement</h2>
         <p><a href="menubar-1.html#code-ex-1">Back to <code>menubar</code> example</a></p>
 
-        <h2 id="copr-rel">Corporate Relations</h2>
+        <h2 id="corp-rel">Corporate Relations</h2>
         <p><a href="menubar-1.html#code-ex-1">Back to <code>menubar</code> example</a></p>
 
         <h2 id="extension">Extension</h2>
@@ -30,4 +30,4 @@
     </main>
 </body>
 
-</html> 
+</html>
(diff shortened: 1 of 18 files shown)


examples/menubar/menubar-1/mb-outreach.html
```diff
@@ -18,7 +18,7 @@ <h1 id="outreach">Outreach</h1>
         <h2 id="comm-engage">Community Engagement</h2>
         <p><a href="menubar-1.html#code-ex-1">Back to <code>menubar</code> example</a></p>
 
-        <h2 id="copr-rel">Corporate Relations</h2>
+        <h2 id="corp-rel">Corporate Relations</h2>
```
jnurthen: I'm not sure idrefs should be spellchecked.

nschonni: They can be indvidually ignored, but in this case there are no refs https://github.com/w3c/aria-practices/search?q=copr&unscoped_q=copr and it looked like a typo

mcking65: looks like we have some unnecessary files left over from old content in the menubar-1 folder.

mcking65:
@nschonni, thank you for all these fixes!

Having integrated spell checking would be great. I don't know anything about cspell. Having to ignore ids individually sounds like serious pain ... makes me wonder if cspell is a good solution for our purposes. Seems like we would not want to spell check the values of any attributes except for perhaps aria-label. The validator will catch most any other attribute mistake.

examples/menubar/menubar-1/mb-outreach.html
jnurthen: Agree - looks like we can remove mb-research.html, mb-arts.html and mb-outreach.html

mcking65: Done in commit c57878b

PR #837, opened 2018-08-16
Fix race conditions and clean up code in regression tests

spectranaut:
Test the CI

PR #846, opened 2018-08-20
Add tests for disclosure/disclosure-faq.html

spectranaut:
Tests written, only blocking on acceptance of race condition PR: #837

### DIFF
--- a/test/tests/disclosure-faq.js
+++ b/test/tests/disclosure-faq.js
@@ -1,0 +1,151 @@
+'use strict';
+
+const { ariaTest } = require('..');
+const { By, Key } = require('selenium-webdriver');
+const assertAriaControls = require('../util/assertAriaControls');
+const assertAttributeValues = require('../util/assertAttributeValues');
+const assertTabOrder = require('../util/assertTabOrder');
+
+const exampleFile = 'disclosure/disclosure-faq.html';
+
+const ex = {
+  buttonSelector: '#ex1 button',
+  answerSelector: '#ex1 p.desc',
+  buttonSelectors: [
+    '#ex1 dt:nth-of-type(1) button',
+    '#ex1 dt:nth-of-type(2) button',
+    '#ex1 dt:nth-of-type(3) button',
+    '#ex1 dt:nth-of-type(4) button'
+  ],
+  answerSelectors: [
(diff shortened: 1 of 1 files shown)


mcking65:
@tatermelon, @jessebeach, @sh0ji, this is ready for review.

mcking65:
@spectranaut, looks like we are getting the session errors. I'm restarting the build another time to see if it happens again.

test/tests/disclosure-faq.js
```diff
@@ -0,0 +1,151 @@
...
+});
+
+ariaTest('"aria-expanded" attribute on button', exampleFile, 'button-aria-expanded', async (t) => {
+  t.plan(6);
+
+  await assertAttributeValues(t, ex.buttonSelector, 'aria-expanded', 'false');
+
+  let buttons = await t.context.session.findElements(By.css(ex.buttonSelector));
+  for (let button of buttons) {
+    await button.click();
```
jessebeach: Huh, I thought having an await in a for..of loop would cause a problem, but the Iteration Protocol handles it automatically. Very readable, very nice.

test/tests/disclosure-faq.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  let answers = await t.context.session.findElements(By.css(ex.buttonSelector));
+  for (let answer of answers) {
+    t.true(
+      await answer.isDisplayed(),
```
jessebeach: This might fail if `true` isn't asynchronous. Maybe pull the `await` to the line above, store the result and then call `true` with the variable as the first argument. If this pattern is prevalent, it might be a source of your timing issues.

test/tests/disclosure-faq.js
```diff
@@ -0,0 +1,151 @@
...
+
+  for (let index = 0; index < ex.buttonSelectors.length; index++) {
+    let buttonSelector = ex.buttonSelectors[index];
+    let answerSelector = ex.answerSelectors[index];
+    let button = await t.context.session.findElement(By.css(buttonSelector));
+
+    await button.sendKeys(Key.ENTER);
+
+    t.true(
+      await waitAndCheckExpandedTrue(t, button),
```
jessebeach: ditto to above and for all the rest below.

PR #849, opened 2018-08-21
Add tests for button/button.html

spectranaut:


mcking65:
@jessebeach, @tatermelon, @sh0ji, even though travis is failing, this is ready for review. The failures are related to a race condition that is not specific to these tests. @spectranaut is debugging the failures separately.

PR #857, opened 2018-08-27
aria1.0 combobox with autocomplete=both example: Make Up Arrow set activedescendant to last match in list

jongund:
I fixed the bug in the up arrow key not properly setting aria-activedescendant


jongund:
I removed the console.log statement from the javascript file

PR #858, opened 2018-08-27
Activate/toggle buttons with space on keyup

### LINKED ISSUE #610: Button: Space should only activate on keyup
The synthetic click event that happens on native `<button>` elements occurs on keydown for Enter and keyup for Space, but the button pattern doesn't describe this behavior and the example does not exhibit it. For instance, holding down the space bar on the "Mute" button toggles it off and on until you release the space bar--it should only trigger once on key release.

I tested this behavior using [this codepen example](https://codepen.io/sh0ji/pen/vdXraZ) on Firefox, Chrome, and Safari on Mac to be certain about it. Holding down space bar on the "real button" only triggers the click even on release, whereas holding down enter triggers it repeatedly until release. I added a "fake button" to demonstrate how this behavior can be replicated in JavaScript.

kleinfreund:
Implements #610.

Both action and toggle buttons are now activated with space on the keyup event and no longer on the keydown event. The behavior for pressing enter has not changed.

As brought up in #610, I would like to remove the safety check on lines 66 to 69 (in the new file). What are your thoughts on that?

Also, do you want me to expand the description in the design pattern to reflect this behavior?

### DIFF
--- a/examples/button/js/button.js
+++ b/examples/button/js/button.js
@@ -40,13 +40,12 @@ function actionButtonKeydownHandler (event) {
 }
 
 /**
- * Activates the action button with the enter or space key.
+ * Activates the action button with the space key.
  *
  * @param {KeyboardEvent} event
  */
 function actionButtonKeyupHandler (event) {
-  // If either enter or space is pressed, activate the button
-  if (event.keyCode === 13 || event.keyCode === 32) {
+  if (event.keyCode === 32) {
     event.preventDefault();
     activateActionButton();
   }
@@ -87,12 +86,12 @@ function toggleButtonKeydownHandler (event) {
 }
 
 /**
- * Toggles the toggle button’s state with the enter or space key.
+ * Toggles the toggle button’s state with space key.
  *
  * @param {KeyboardEvent} event
  */
 function toggleButtonKeyupHandler (event) {
-  if (event.keyCode === 13 || event.keyCode === 32) {
+  if (event.keyCode === 32) {
     event.preventDefault();
     toggleButtonState(event.currentTarget);
   }
--- a/examples/button/js/button.js
+++ b/examples/button/js/button.js
@@ -2,84 +2,114 @@
 *   This content is licensed according to the W3C Software License at
 *   https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
 *
-*   File:   button.js
-*
-*   Desc:   JS code for Button Design Pattersn
+*   JS code for the button design pattern
 */
 
-var ICON_MUTE_URL  = 'images/mute.svg#icon-mute';
+var ICON_MUTE_URL = 'images/mute.svg#icon-mute';
 var ICON_SOUND_URL = 'images/mute.svg#icon-sound';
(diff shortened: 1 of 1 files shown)


examples/button/js/button.js
```diff
@@ -2,84 +2,114 @@
...
+}
+
+/**
+ * Activates the action button with the enter or space key.
+ *
+ * @param {KeyboardEvent} event
+ */
+function actionButtonKeyupHandler (event) {
+  // If either enter or space is pressed, activate the button
+  if (event.keyCode === 13 || event.keyCode === 32) {
```
sh0ji: The <kbd>Enter</kbd> check should be removed here. It's causing `activateActionButton` to trigger twice on <kbd>Enter</kbd> keypresses, effectively making the toggle button pointless.

examples/button/js/button.js
```diff
@@ -2,84 +2,114 @@
...
-  var button = event.target;
-  var currentState = button.getAttribute('aria-pressed');
-  var newState = 'true';
+/**
+ * Toggles the toggle button’s state with the enter or space key.
+ *
+ * @param {KeyboardEvent} event
+ */
+function toggleButtonKeyupHandler (event) {
+  if (event.keyCode === 13 || event.keyCode === 32) {
```
sh0ji: Same thing as line 49 here.

examples/button/js/button.js
```diff
@@ -2,84 +2,114 @@
...
-    newState = 'false';
-    newIconState = ICON_SOUND_URL;
-  }
+  button.setAttribute('aria-pressed', isAriaPressed ? 'false' : 'true');
 
-  // Set the new aria-pressed state on the button
-  button.setAttribute('aria-pressed', newState);
-  icon.setAttribute('xlink:href', newIconState);
+  var icon = button.querySelector('use');
+  icon.setAttribute('xlink:href', isAriaPressed ? ICON_SOUND_URL : ICON_MUTE_URL);
```
sh0ji: 👍

examples/button/js/button.js
kleinfreund: Oh, what a blunder. Of course.

kleinfreund:
I updated the pull request according to @sh0ji’s review.

kleinfreund:
Not sure if that Travis CI check is supposed to complete. It seems like it got stuck?

caasi:
My friend is working on her own ARIA practices. We are wondering if there is any reference about this keyup behavior other than the "Change History" section of [WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#change_log_2019_january). Can anyone point a direction?

caasi:
> Is it open source?

Currently it's just a [code sandbox](https://twitter.com/miau715/status/1184858504135245824).

> This was done to better match what the HTML button element does on native platforms.

You are right. Another friend @othree found an answer about this behavior here: https://stackoverflow.com/questions/16090578/why-do-enter-and-space-keys-behave-differently-for-buttons

PR #863, opened 2018-08-27
Add data-test-ids for regression tests

spectranaut:
Hi @mcking65 -- if you could review and merge this soon, that would be great!

### DIFF
--- a/examples/alert/alert.html
+++ b/examples/alert/alert.html
@@ -79,13 +79,13 @@
           </tr>
         </thead>
         <tbody>
-          <tr>
+          <tr data-test-id="alert-role">
             <th scope="row"><code>alert</code></th>
             <td></td>
             <td><code>div</code></td>
             <td>Identifies the element as the container where alert content will be added or updated.</td>
           </tr>
-          <tr>
+          <tr data-test-id="alert-aria-live">
             <td></td>
             <th scope="row"><code>aria-live=<q>assertive</q></code></th>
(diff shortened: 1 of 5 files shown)


examples/alert/alert.html
```diff
@@ -79,13 +79,13 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
...
             <th scope="row"><code>alert</code></th>
             <td></td>
             <td><code>div</code></td>
             <td>Identifies the element as the container where alert content will be added or updated.</td>
           </tr>
-          <tr>
+          <tr data-test-id="alert-aria-live">
             <td></td>
             <th scope="row"><code>aria-live=<q>assertive</q></code></th>
             <td>Implicit on <code>div</code></td>
```
mcking65: We do not need a test for implicit property; not sure that this row should even be in the table. Oddly with this example, the thing that needs to be tested, pressing the button triggers the alert, is not really documented in the table because it is not part of the aria markup.

examples/alert/alert.html
```diff
@@ -96,7 +96,7 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
               </ul>
             </td>
           </tr>
-          <tr>
+          <tr data-test-id="alert-aria-atomic">
```
mcking65: similar comment here; this implicit so nothing to test.

PR #867, opened 2018-08-30
typos: Double words

nschonni:
Just ran Vale to find duplicate words. There were a few others, but they seemed intentional

***
<a href="https://pr-preview.s3.amazonaws.com/nschonni/aria-practices/pull/867.html" title="Last updated on Aug 30, 2018, 5:35 AM GMT (48966e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/867/1379511...nschonni:48966e8.html" title="Last updated on Aug 30, 2018, 5:35 AM GMT (48966e8)">Diff</a>

### DIFF
--- a/aria-practices-DeletedSectionsArchive.html
+++ b/aria-practices-DeletedSectionsArchive.html
@@ -3305,7 +3305,7 @@ field. </p>
       extensions for Mozilla/Firefox</a> from the University of Illinois at
       Urbana-Champaign to render the document landmarks. This picture
       shows the Firefox browser rendering the University of Illinois
-      Career Career Center home page. In this example. The "List of Navigation Bars" viewer is shown, launched from the extension on the
+      Career Center home page. In this example. The "List of Navigation Bars" viewer is shown, launched from the extension on the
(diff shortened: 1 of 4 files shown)


mcking65:
Wow, @nschonni, you're a never-ending source of quality checking tools!

This is an interesting one for me ... I discovered that I can't even hear the difference between " the the " and "the " at my normal listening speed in most cases. It depends a bit on the preceding and following words. In some cases, the only way I could find the dupe word without navigating word by word was to slow speech down to a crazy slow speed. I have long fancied my concentration levels were better than that ... kind of sucks to learn just how limited they are.

Great catches!

Is Vale something we my want to integrate? Maybe as a check that is only informational, e.g., doesn't block builds? Or, would it be better to just run manually on occasion?

nschonni:
It could probably be added to the build, but I'm not sure about the Warning vs. Error part of setting it up.
The project is here https://github.com/errata-ai/vale and has distributions for most platforms since it is a GoLang project.
If you're interested in trying that, just create an issue and ping me.

PR #851, opened 2018-08-21
chore: Use Travis Build stages

nschonni:
Related to #848
This will add 2 separate builds to separate out the linting and HTML validation so there are separate results reported

PR #872, opened 2018-08-31
Add regression test run to travis CI build stages

spectranaut:
Before #851, travis.ci ran "npm test", which runs the linters and the regression tests all in one. Now the tests are separated out and the regression tests needed to be added explicitly back in.

This PR also sets "concurrency=1" for the ava tests. This will element the bug which causes firefox to crash at least once (sometimes twice) during the regression test. We reported the crashing of firefox to geckodriver in this [bug](https://github.com/mozilla/geckodriver/issues/1135#issuecomment-417337821).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ stages:
 jobs:
   include:
     - stage: Test
+      script: ava -c 1 test/tests
+    - stage: Test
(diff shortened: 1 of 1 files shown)


mcking65:
Looks good  ... let's give it a spin!

PR #876, opened 2018-09-05
Regression test: expanded error messages, rare race condition fixes

spectranaut:
I was reviewing some old code and found a few things to fix:
- add useful timeout error messages to all "wait" calls (waiting for state to change after sending interaction to browser)
- use "t.context.url" (set during test set up) instead of getting the url from the browsers
- while testing I found I could regularly locally reproduce some race conditions in the combobox tests, so I fixed those as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,6 @@ jobs:
         script: /tmp/deploy/travis-deploy.sh
         on:
           branch: master
+
+env:
+  - TEST_WAIT_TIME=1000
--- a/test/index.js
+++ b/test/index.js
@@ -9,6 +9,7 @@ const startGeckodriver = require('./util/start-geckodriver');
 
 let session, geckodriver;
 const firefoxArgs = process.env.CI ? [ '-headless' ] : [];
+const testWaitTime = process.env.TEST_WAIT_TIME || 500;
 
 test.before(async (t) => {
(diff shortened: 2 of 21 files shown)


mcking65:
@jessebeach, @tatermelon, @sh0ji, this is ready for review.

@spectranaut, Is the failure in the treeview test expected?

mcking65:
I'm guessing that increasing wait time on that test is better if it works; it would be overall less runtime if there is slow response.

Makes me wonder if I should work to eliminate dependencies on external networks/servers. If those links all pointed to the same local dummy file like we do in the navigation menubar instead of pointing to wikipedia, the tests may run faster and be more reliable. Thoughts?

sh0ji:
Another option: make the wait time a variable that uses an environment variable and falls back to some default, and then tweak that environment variable in the Travis config. Locally-running tests won't have it set, so they'll use the default.

For example:
```javascript
const waitTime = process.env.TEST_WAIT_TIME || 500;
```

```yaml
# .travis.yml
env:
  - TEST_WAIT_TIME=1000
```

spectranaut:
I think you are right @mcking65. Probably it's best for the example pages to be isolated.

In the meantime, I added the waitTime you suggested @sh0ji, thanks! :)

spectranaut:
@mcking65 @sh0ji can this be merged? I touch a lot of files so I want to rebase my work on top of it.

PR #881, opened 2018-09-11
Add data-test-id to example pages for regression

spectranaut:
@mcking65, please review and merge when convenient :)

### DIFF
--- a/examples/menu-button/menu-button-links.html
+++ b/examples/menu-button/menu-button-links.html
@@ -118,11 +118,11 @@
             </tr>
           </thead>
           <tbody>
-            <tr>
+            <tr data-test-id="button-down-arrow">
               <th><kbd>Down Arrow</kbd><br><kbd>Space</kbd><br><kbd>Enter</kbd></th>
               <td>Opens <code>menu</code> and moves focus to first <code>menuitem</code></td>
             </tr>
-            <tr>
+            <tr data-test-id="button-up-arrow">
               <th><kbd>Up Arrow</kbd></th>
               <td>Opens <code>menu</code> and moves focus to last <code>menuitem</code></td>
             </tr>
@@ -137,7 +137,7 @@
             </tr>
           </thead>
           <tbody>
-            <tr>
+            <tr data-test-id="menu-enter">
               <th><kbd>Enter</kbd></th>
               <td>
                 <ul>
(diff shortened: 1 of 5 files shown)


examples/menu-button/menu-button-links.html
```diff
@@ -118,11 +118,11 @@ <h3 id="kbd1_label">Menu Button</h3>
             </tr>
           </thead>
           <tbody>
-            <tr>
+            <tr data-test-id="button-down-arrow">
```
mcking65: There are 3 keys to test here. Should that somehow be represented with the testid value? Multiple values? Multiple IDs?

examples/menu-button/menu-button-links.html
```diff
@@ -174,15 +174,15 @@ <h3 id="kbd2_label" class="widget">Menu</h3>
...
               <th><kbd>Home</kbd></th>
               <td>Moves focus to the first menu item.</td>
             </tr>
-            <tr>
+            <tr data-test-id="menu-end">
               <th><kbd>End</kbd></th>
               <td>Moves focus to the last menu item.</td>
             </tr>
-            <tr>
+            <tr data-test-id="menu-character">
```
mcking65: I see that we do not have tab and shift+tab documented for the menu, but we have scripting to close the menu without taking action and to go to next/prior element when the user tabs. After merge, I'll add that documentation with data-test-id attribs.

examples/menu-button/menu-button-links.html
spectranaut: in previous instances I used values like `key-enter-or-space`. This scenario, I simply had two tests for this data-test-id, one that tested `enter` and one that tested `space`. The value `button-down-arrow-or-space-or-enter` is a little long, but I could continue that convention (I would have but I missed the multiple entries in this row).

Alternatively, the data-test-ids could be space delimited values. This would require going back and changing a few data-test-ids and test files. I would prefer to do this after you merge to master so I don't have to update things that depend on each other in two different branches? For now I'll use the old convention, if you don't mind!

examples/menu-button/menu-button-links.html
spectranaut: ok!

examples/menu-button/menu-button-links.html
mcking65: I'm fine with the old convention for now but think multiple space-delimited tokens is probably a better  long term approach. It would be good for each test to have its own token so that if we change a key, e.g., split to a different row, we'd only have to change the test for that key and it wouldn't effect testing the other key presses that didn't change.

Issue #541, opened 2017-11-26
Toolbar Example: Use button elements and add other types of elements, such as link and checkbox


mcking65:
In the [toolbar example](http://w3c.github.io/aria-practices/examples/toolbar/toolbar.html),
make the following enhancements:

- [x] replace some of the buttons with other types of elements, such as link, checkbox, toggle button, and switch.
- [x] For the link and buttons, provide some indication of activation when activated, e.g., popup a window alert.
- [x] Change `<div role="button">` to HTML `<button>` elements.

mcking65:
@gerardkcohen is working on this issue.

gerardkcohen:
Yes, I still intend in finishing this. It is very close to being done.

mcking65:
With commit 599d2de from pr #887, the issue541-toolbar-redesign feature branch now contains all changes needed to resolve this issue. Further review will be managed in issue #539.

Issue #847, opened 2018-08-20
Toolbar Example: Bugs with keyboard support, aria-disabled, and aria-label


mcking65:
As noted in issue #539, the [toolbar example](http://w3c.github.io/aria-practices/examples/toolbar/toolbar.html) has the following bugs:

- [x] Missing `aria-label="Example Toolbar"` on `div` with `role="`toolbar"`
- [x] Missing `aria-disabled="true"` and appropriate greyed out styling on one of the button elements (add to second button element in toolbar). As documented in accessibility features section, the disabled button should remain focusable.
- [x] `Home` key does not move focus to first element.
- [x] `End` key does not move focus to last element.
- [x] `Right Arrow` key does not wrap focus from last to first element.
- [x] `Left Arrow` key does not wrap focus from first to last element.
- [x] Elements in toolbar do not implement their respective patterns, e.g., enter and space do not activate buttons. Each element should execute an action, e.g., buttons and links open alert dialog. This let's users know the element was activated and prevents default action, e.g., scrolling with space.

mcking65:
@tatermelon is planning to submit a PR for this issue.

jongund:
I am planning on submitting a pull request on an updated toolbar example later this week.  See the following link:
https://rawgit.com/jongund/aria-practices/toolbar/examples/toolbar/toolbar.html

Mostly need to clean up some CSS best practices and focus styling issues with my student.

mcking65:
@jongund, the WIP you have looks like a more realistic toolbar. I didn't know you were working on that. With a few changes, it could resolve issue #541. That is something we should discuss. In fact, that would be a good way to frame the PR.

For project management reasons, I don't want this issue closed until we have the resolution committed. So, I am re-opening it.

The contract with Bocoup referenced the current toolbar; this one is significantly more complex. The tests for the current toolbar are merged into the Bocoup branch, The tests that fail due to the bugs in the current example are commented out. This example would drive rework on that set of tests, which @spectranaut may not be able to contain.

Timing will also be a factor. I see quite a few issues that need resolution and possibly discussion in this new example. So, I am not completely confident it can be done before the Bocoup contract ends. The current target completion date for test development is sep 24.

[...]

PR #887, opened 2018-09-13
Toolbar example for simple text editor

jongund:
This updated toolbar is intended to resolve issues #541 and #847.

[Preview the new toolbar in Jon's repo](https://rawgit.com/jongund/aria-practices/toolbar/examples/toolbar/toolbar.html)


.gitconfig
```diff
@@ -0,0 +1,3 @@
+[branch]
```
mcking65: Changes to this file should not be in the PR

examples/menubar/menubar-2/js/MenubarAction.js
```diff
@@ -65,17 +65,18 @@ MenubarAction.prototype.init = function (actionManager) {
   var menubarItem, childElement, menuElement, textContent, numItems;
```
mcking65: Changes to this file should not be included in this pull request

examples/menubar/menubar-2/js/MenubarItemAction.js
```diff
@@ -58,7 +58,7 @@ var MenubarItemAction = function (domNode, menuObj) {
```
mcking65: Changes to this file should not be in this PR.

examples/toolbar/toolbar.html
```diff
@@ -44,32 +46,63 @@ <h1>Toolbar Example</h1>
     <h2 id="ex_label">Example</h2>
     <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
     <div id="ex1">
-        <div class="toolbar" role="toolbar">
-            <div class="toolbar-item selected" role="button" tabindex="0">
-                Tool 1
+        <div class="toolbar" role="toolbar" aria-label="Toolbar">
+          <div class="group characteristics">
+            <div role="button" class="toolbar-item bold" aria-pressed="false" value="bold" tabindex="0">
+              <i class="fas fa-bold"></i>
```
mcking65: This content does not provide an accessible label so none of the buttons in this toolbar are accessible.

examples/toolbar/toolbar.html
```diff
@@ -44,32 +46,63 @@ <h1>Toolbar Example</h1>
...
+            <div role="button" class="toolbar-item italic" aria-pressed="false" value="italic" tabindex="1">
+              <i class="fas fa-italic"></i>
+            </div>
+            <div role="button" class="toolbar-item underline" aria-pressed="false" value="underline" tabindex="1">
+              <i class="fas fa-underline"></i>
+            </div>
+          </div>
+          <div class="group align">
+            <div role="button" class="toolbar-item textAlign" aria-pressed="true" value="left" tabindex="1">
+              <i class="fas fa-align-left"></i>
```
mcking65: These alignment buttons work like radio so cannot be toggles. Use a radio group or alternatively, use a popup menu with menuitem radios.

jongund:
Matt,

I am working on getting these problems fixed by tonight.  I should have reviewed the students code more carefully before submitting the pull request.

jongund:
Matt,

I checked in the latest changes and I think I have got all the problems fixed and improved the code readability

jongund:
In the short term we will remove the font size buttons and replace them with cut. copy and paste buttons.   It is nuclear to me why you would have links or checkboxes in a toolbar, so I need to learn more about why these are important and maybe we need a different toolbar example for those cases.

examples/toolbar/toolbar.html
```diff
@@ -44,32 +51,123 @@ <h1>Toolbar Example</h1>
...
-        <div class="toolbar" role="toolbar">
-            <div class="toolbar-item selected" role="button" tabindex="0">
-                Tool 1
+      <div class="format" role="toolbar" aria-label="Text Formatting" aria-controls="textarea1">
+          <div class="group characteristics">
+            <div role="button"
+                class="item bold"
+                aria-pressed="false"
+                aria-label="bold"
+                value="bold"
```
mcking65: This is not an html button. From an HTML perspective, this is a div. therefore yu can't use the value attribute. You could instead use an HTML button. That would be better anyway because we should use as much native HTML as possible. Only use the ARIA necessary to demonstrate ARIA toolbars and the ariety of elements  they can contain.

examples/toolbar/toolbar.html
```diff
@@ -44,32 +51,123 @@ <h1>Toolbar Example</h1>
     <h2 id="ex_label">Example</h2>
     <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
     <div id="ex1">
-        <div class="toolbar" role="toolbar">
-            <div class="toolbar-item selected" role="button" tabindex="0">
-                Tool 1
+      <div class="format" role="toolbar" aria-label="Text Formatting" aria-controls="textarea1">
```
mcking65: The way that most screen readers render toolbars, I don't know how it be possible for them to exploit aria-controls on a toolbar container. I think we should not have it here.

mcking65: Actually, on further investigation, at least JAWS is seeing aria-controls on each button, and it enables movement to the text area from the button if in eading mode. I didn't expect that specifying aria-controls on the parent container would do that -- very interesting. So, at least there is some support for aria-controls in this scenario. The question is whether we want this to be part of the pattern.

jongund:
I updated the toolbar example and it can be viewed here:
https://rawgit.com/jongund/aria-practices/toolbar/examples/toolbar/toolbar.html

jongund:
I updated pull request to add the design pattern references to the documentation and to remove left and right keyboard behaviors from the pull down menu items.   Still working on the test cases.

carmacleod:
Finally had a chance to look at this quickly and write down a few suggestions:

- Please don't use `<i>` for icons. I know everybody does it, but the semantics of `<i>` are "my content is in italics", not "my content is an icon". We should be using a span instead.

- The "Text Alignment" group should be a "radiogroup" with "radio" children, not a role="menu".

- Consider giving the div's with `class="group"` a `role="group"` and an aria-label. (Not sure if this would be too verbose - might check with @mcking65).

- A more common ordering of this type of toolbar is font name first, then font size, then font emphasis like bold, italic, underline, then the paragraph alignment.

- Cut, Copy, and Paste have a title (i.e. tooltip; note misspelled "from" in Cut/Copy), so for consistency (and for sighted users) Bold, Italics, Underline, Align left/center/right, Font type, and size increment/decrement should have a title also. (Or even better, all of the tools should have a custom tooltip that comes up on both hover and focus, or only on hover but user can type Shift+F1/Cmd+? to open tooltip when focused).

- Would be nice if the tools supported the common shortcut keys, i.e. Ctrl/Cmd + B for Bold, Ctrl/Cmd + I for Italics, Ctrl/Cmd + U for Underline, Ctrl/Cmd + L/E/R for Left/Center/Right alignment; and if so, then all tooltips should include the appropriate shortcut key.

- Would be nice if all of the tools supported a visual change on hover.

- The visual difference between pressed/not pressed toggle buttons and checked/unchecked radios is too subtle. It would not pass the new [WCAG 2.1 SC 1.4.11 Non-text Contrast](https://w3c.github.io/wcag21/guidelines/#non-text-contrast), i.e.
  - Unpressed toggle white background:#FFFFFF in toolbar background:#EDEDEB contrast ratio is: 1.2:1
  - Pressed toggle 1-pixel outline:#F5F5F5 beside button background:#A9A9A9 contrast ratio is: 2.2:1
  - Downward-pointing triangle in menu button:#878787 on white background:#FFFFFF contrast ratio is: 3.6:1
  - Toolbar background:#EDEDEB to pressed toggle background:#F5F5F5 contrast ratio is: 1.1:1
  - Pressed toggle background:#F5F5F5 to unpressed toggle white background:#FFFFFF contrast ratio is: 1.1:1

jongund:
Carolyn,

Thank you for the feedback and comments.


  1.  I will fix the problem with the <I> element.
  2.  The standard radio button pattern does not work since selection (e.g. checked item) in the radio group pattern moves when focus moves, so we need to have a select one of many pattern, where selection must be explicitly  set.  So there best option right now is menuitemradio.   Documentation on this usage needs to be improved in the example, I will add the explanation.
  3.  I will fix the tooltip issue.
  4.  Right now the toolbar pattern does not discuss the use of shortcut keys, so created an issue for the working group:
https://github.com/w3c/aria-practices/issues/948
  5.  I can add some hover styling to the buttons
  6.  The pressed buttons include changing the border to have more than color, but I will also update to your color recommendations.

Jon

jongund:
Updated pull request to include changes from Carolyn.

carmacleod:
Wow, that was fast - thanks, Jon!

Regarding:
> 2.  The standard radio button pattern does not work since selection (e.g. checked item) in the radio group pattern moves when focus moves

Ah, I see. So we need to fix the pattern to allow for radiogroups in toolbars. I've opened https://github.com/w3c/aria-practices/issues/949 to address this.

In the meantime, you might want to add `aria-orientation="horizontal"` to the "menu" to indicate that left/right arrow keys are expected (although I see that it works with up/down arrows as well, so not critical).

I like the new hover and focus styles - they are beautifully clear! It's still hard to tell the difference between checked and not-checked toggles and radios, though.

Are you still planning to add a title/tooltip to the Bold, Italics, Underline, Align left/center/right, Font type, and size increment/decrement tools (for consistency with Cut/Copy/Paste, and in case any sighted folks need help understanding the icons)?

Thanks again for the super-quick updates and explanations!

mcking65:
@jongund, during tomorrow's meeting, we'll discuss how to divide remaining work on this.

annabbott:
I can now visually tell the difference between checked and not-checked toggles and radios with ease. Looks good to me.

Issue #790, opened 2018-07-19
Review Treegrid Example


mcking65:
The [editor's draft of the treegrid example](http://w3c.github.io/aria-practices/examples/treegrid/treegrid-1.html)
developed for issue #132 is complete.
Because of the complexity of this example, the APG task force is seeking broad community feedback with its inclusion in APG 1.1 Release 2 and encourages anyone who may reference this example to follow this issue and any subsequent updates made to the editor's draft of the example page referenced in this issue.

Please provide feedback about the functionality or documentation of this example by commenting on this issue.

#### Known issues

[ ] Issue #789: Some focus rectangles overlap neighboring cells.

devarshipant:
- NVDA 2018.1 and Chrome 68:

1. There is no association with the table header when focus is on a gridcell.
2. Speaks the level number and expanded state.
3. Missing instructions on using arrow keys to navigate.
4. Should the treegrid with table headers be navigable using table quick key?


- JAWS 2018 and Chrome 68:

1. There is association with the table header when focus is on a gridcell, except the first header Subject.
2. Speaks expanded state, not the level.
3. Missing instructions on using arrow keys to navigate.
4. JAWS recognizes the table and allows quick key navigation (T).
5. Changes cursor mode (to Virtual) when JAWS encounters the link throwing off arrow navigation.

spectranaut:
I'm almost writing regression tests for this example. This is really good work! :)

There are only two places where the documentation is unclear. Both are for `tabindex` in the "Role, Property, State, and Tabindex Attributes" table:
* in the `tabindex="-1"` row 
  * It says: "All row and gridcell elements are focusable, but only one is included in the tab sequence." 
  * This is not exactly true, gridcells do not have `tabindex="-1"` but rather no tabindex set. When there is an interaction with the treegrid that should move focus to that gridcell, the gridcell will get `tabindex='0'`. If focus moves from the gridcell, it will go back to having no tabindex. In a semantic way, all gridcells are "focusable", but I think that using "focusable" semantically might be misleading (because you cannot call focus on them nor send keys to them unless they are already in focus).
* in the `tabindex="0"` row
  * It says: "Only one row or gridcell in the treegrid has tabindex="0".
[...]

PR #889, opened 2018-09-17
Add remaining test for treegrid/treegrid-1.html

spectranaut:
This completes all tests for the treegrid example. This is an odd PR, because the only test added is also commented out due to a bug. This is our convention for handling bugs in the examples so far.

You can read about the bug in the open issue on treegrid [here](https://github.com/w3c/aria-practices/issues/790#issuecomment-422079276).

### DIFF
--- a/test/tests/treegrid-1.js
+++ b/test/tests/treegrid-1.js
@@ -678,8 +678,52 @@ ariaTest('CTRL+END moves focus', exampleFile, 'key-control-end', async (t) => {
   );
 });
 
-// TODO: backlog issue 878
+// This test fails due to: https://github.com/w3c/aria-practices/issues/790#issuecomment-422079276
+// When the bug is fixed this test should be uncommented.
 
-// ariaTest('ENTER actives item', exampleFile, 'key-enter', async (t) => {
-//   t.pass();
+// ariaTest('ENTER actives interative items item', exampleFile, 'key-enter', async (t) => {
+//   t.plan(16);
+
+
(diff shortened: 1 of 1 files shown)


mcking65:
@jessebeach, @tatermelon, @sh0ji, this PR is ready for review.

Side note: we've discussed finding a better way to control whether or not a test runs for a known bug. As @spectranaut mentions, our current approach is to comment out the test. That's a stopgap in my mind. I'd like something that does not involve editing the test code and is easier to track. We have not settled on an approach and are open to suggestions.

nschonni:
Maybe using the `skip` function https://github.com/avajs/ava#skipping-assertions

mcking65:
> Maybe using the `skip` function https://github.com/avajs/ava#skipping-assertions


Hmm, good suggestion. I also see an option called `Failing tests`.:

>You can use the .failing modifier to document issues with your code that need to be fixed. Failing tests are run just like normal ones, but they are expected to fail, and will not break your build when they do. If a test marked as failing actually passes, it will be reported as an error and fail the build with a helpful message instructing you to remove the .failingmodifier.
>
>This allows you to merge .failing tests before a fix is implemented without breaking CI. This is a great way to recognize good bug report PR's with a commit credit, even if the reporter is unable to actually fix the problem.

@spectranaut, what do you think?

nschonni:
Ah, didn't see that option https://github.com/avajs/ava#failing-tests it does seem better than skip (as long as the test steps are OK to run without affecting others.

PR #898, opened 2018-09-21
Use ava '.failing' to skip known failures

spectranaut:
This commit:
- adds the `.failing` option to the ariaTest wrapper around ava's test
- uses `.failing` in all places where failing tests were previously commented out


spectranaut:
The CI is getting a timeout... it seems like something is crashing when trying to run a failing test, you can see it in the last two CI runs. I can't replicate locally.  I'm going drop the test that seems to be causing a failure from this commit and address it in a different issue.

PR #901, opened 2018-09-24
Regression test clean up

spectranaut:
* Using .failing for treegrid test (this one test got lost in the sea of PRs last week)
* Remove t.log from datagrid test

PR #909, opened 2018-09-27
Regression test coverage report

spectranaut:


.travis.yml
```diff
@@ -17,6 +17,9 @@ jobs:
     - stage: Test
       name: AVA Regression Tests
       script: ava --tap -c 1 test/tests
+    - stage: Test Coverage
```
nschonni: Should the regression and lint jobs be done in separate stages too?

spectranaut: hmmm actually I don't think this should be a separate stage. I thought perhaps it should be in a different category -- because it is a coverage report -- but I don't think the stages should be used like that. It seems the next stage doesn't run if the first one fails.

spectranaut:
To see the results of the report in the CI, look here: https://travis-ci.com/w3c/aria-practices/jobs/148463268

mcking65:
@jessebeach, @tatermelon, @sh0ji, this is ready for review.

jessebeach:
The report wants more tests:

```
ERROR:  Please write missing tests for this report to pass.
```

spectranaut:
The report will no longer fail if this PR is accepted: https://github.com/w3c/aria-practices/pull/901
And I rebase this branch on top of bocoup :)

PR #916, opened 2018-10-01
Change regression test output from TAP to ava default

spectranaut:


jessebeach:
What is TAP output?

nschonni:
> What is TAP output?

https://en.wikipedia.org/wiki/Test_Anything_Protocol kind of a flat file/ini looking thing

mcking65:
We used it for a while because it is easier for me to read ... but then I decided to go back and deal with default because the TAP didn't handle aria.failing well (at all). The default creates some pretty red formatting that you sightees can't miss. Making output that is awesome for everyone is a future project idea.

PR #922, opened 2018-10-03
Add wait to dialog to account for empty div 

spectranaut:
As we just discussed, @mcking65! Can you take a look @jessebeach?

This is a PR on master because I'm not sure when you will get to review/merge bocoup into master for good :)

### DIFF
--- a/test/tests/dialog.js
+++ b/test/tests/dialog.js
@@ -127,11 +127,14 @@ const reload = async (t) => {
 };
 
 const checkFocus = async function (t, selector) {
-  return t.context.session.executeScript(function () {
-    const [ selector ] = arguments;
-    const items = document.querySelector(selector);
-    return items === document.activeElement;
-  }, selector);
+  return t.context.session.wait(
+    t.context.session.executeScript(function () {
+      const [ selector ] = arguments;
+      const items = document.querySelector(selector);
+      return items === document.activeElement;
+    }, selector),
+    t.context.waitTime
+  );
 };
 
 const sendTabToSelector = async function (t, selector) {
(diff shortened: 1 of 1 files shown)


test/tests/dialog.js
```diff
@@ -246,7 +249,7 @@ ariaTest('tab changes focus within dialog', exampleFile, 'key-tab', async (t) =>
     t.true(
       await checkFocus(t, ex.dialog2FocusableEls[i]),
       'Focus should be on "' + ex.dialog2FocusableEls[i] + '" after ' +
-        (i + 1) + ' tabs have been sent to dialog 2'
+        (i + 2) + ' tabs have been sent to dialog 2'
```
jessebeach: What does `2` represent here? A comment to explain these offsets will be helpful.

spectranaut: it's the number of times tab has been sent to the browser. The for-loop is index 0, but we say that we sent "1" tab to the browser, not "0" tabs. in the case where there is 2, there is a tab send before the for loop starts.

Lemme think about how to make this more clear..

spectranaut: oh wow! I thought I had the logic wrong the FIRST time but I actually had it wrong THIS time.

spectranaut:
@jessebeach -- do you still think a comment should be added to make this clearer? I feel like the test text doubles as a comment..

jessebeach:
> @jessebeach -- do you still think a comment should be added to make this clearer? I feel like the test text doubles as a comment.

This comment was hidden in a code fold:

> // Loop through the focusable elements
   // focus is not first focusable element on popup -- send tab to start

I see your point. This comment is sufficient. No update needed here.

PR #931, opened 2018-10-18
Apply role="none" to <li> elements in menubar

### LINKED ISSUE #930: menubar example bug: missing role="none" on some <li>
The menubar-1.html example says that the `<li>` elements under the `role="menubar"` should have `role="none"`. The example states this under the "Role, Property, State, and Tabindex Attributes" table.

However, the [actual example does not include `role="none"` on the `<li>` elements](https://github.com/w3c/aria-practices/blob/master/examples/menubar/menubar-1/menubar-1.html#L53).

I think this might be causing NVDA 2018.3.2 + Firefox 62 to announce each menuitem has “1 of 1” instead of "1 of 3".

I'm happy to submit a PR to fix this if you want, but I thought I'd confirm before I create a PR.

mfairchild365:
fixes #930

### DIFF
--- a/examples/menubar/menubar-1/menubar-1.html
+++ b/examples/menubar/menubar-1/menubar-1.html
@@ -50,7 +50,7 @@
     <div id="ex1">
       <nav aria-label="Mythical University">
         <ul id="menubar1" role="menubar" aria-label="Mythical University">
-          <li>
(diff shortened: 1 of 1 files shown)


jnurthen:
Thanks

PR #941, opened 2018-11-06
Grid Examples: Fix event listener cleanup

gilmoreorless:
First of all, thank you very much for the existence of this project. It has been the perfect help for me while implementing a grid-like component.

However, while looking at the example JS code for layout grids, I noticed that event listeners were not being cleaned up properly. This resulted in multiple instances of the same listener being attached to an element.

Calling `.bind()` on a function creates a new function each time, so this code doesn't quite do what it looks like:

```js
// Creates a new function with a specific binding, then adds it as a listener.
elem.addEventListener('focus', this.showKeysIndicator.bind(this));

// Creates a new function with a specific binding, then removes it as a listener.
// This does NOT remove the original listener created above.
elem.removeEventListener('focus', this.showKeysIndicator.bind(this));
```

To fix this, I added a utility method inspired by `_.bindAll()` from Underscore/Lodash and removed all the explicit `.bind()` calls in `dataGrid.js`.

### DIFF
--- a/examples/js/utils.js
+++ b/examples/js/utils.js
@@ -117,3 +117,10 @@ aria.Utils.removeClass = function (element, className) {
   var classRegex = new RegExp('(\\s|^)' + className + '(\\s|$)');
   element.className = element.className.replace(classRegex, ' ').trim();
 };
+
+aria.Utils.bindMethods = function (object /* , ...methodNames */) {
+  var methodNames = Array.prototype.slice.call(arguments, 1);
+  methodNames.forEach(function (method) {
+    object[method] = object[method].bind(object);
+  });
+};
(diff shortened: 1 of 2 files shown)


mcking65:
@gilmoreorless, thank you very, very much for this PR. Sorry for the long delay getting it into review.

@tatermelon, can you please review?

tatermelon:
This looks good to me!

mcking65:
@gilmoreorless, thank you again for this fix!!

examples/js/utils.js
```diff
@@ -117,3 +117,10 @@ aria.Utils.removeClass = function (element, className) {
   var classRegex = new RegExp('(\\s|^)' + className + '(\\s|$)');
   element.className = element.className.replace(classRegex, ' ').trim();
 };
+
+aria.Utils.bindMethods = function (object /* , ...methodNames */) {
+  var methodNames = Array.prototype.slice.call(arguments, 1);
```
bluSCALE4, line 122: This is clever. Didn't know you could write such a function.

PR #943, opened 2018-11-12
929 fix

### LINKED ISSUE #929: IE11 Error
IE11 throws the following error when trying to open a dialog on line if (this.dialogNode === null || !['dialog', 'alertdialog'].includes(role)) of dialog.js in the constructor function:
	"Object doesn't support property or method 'includes'"

This doesn't happen in Chrome or Firefox.

sh0ji:
Fixes #929.

I went with [Array.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) to do exact role matching for either `dialog` or `alertdialog`. It's well supported across browsers, and allows us to look through the tokenized list of roles, which should address [James' point](https://github.com/w3c/aria-practices/issues/929#issuecomment-430410032).

@RSugelSr's proposed [String.prototype.indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) check was more prone to false positives like `role="invalid-dialog"`, which is why I didn't use it.


examples/dialog-modal/js/dialog.js
```diff
@@ -127,8 +127,11 @@ aria.Utils = aria.Utils || {};
   aria.Dialog = function (dialogId, focusAfterClosed, focusFirst) {
     this.dialogNode = document.getElementById(dialogId);
     var role = this.dialogNode.getAttribute('role');
+    var isDialog = role.split(' ').some(function (r) {
```
jkva: I'd suggest `trim()`ing `r` as well.

Does this code need to check for `this.dialogNode === null`? Wouldn't the `getAttribute('role')` call at line 129 already fail?

sh0ji: Good call! I added a fix.

examples/dialog-modal/js/dialog.js
```diff
@@ -126,12 +126,20 @@ aria.Utils = aria.Utils || {};
    */
   aria.Dialog = function (dialogId, focusAfterClosed, focusFirst) {
     this.dialogNode = document.getElementById(dialogId);
-    var role = this.dialogNode.getAttribute('role');
-    var isDialog = role.split(' ').some(function (r) {
-      return r === 'dialog' || r === 'alertdialog';
-    });
+    if (this.idalogNode === null) {
```
jkva: `idalogNode` looks like a typo.

examples/dialog-modal/js/dialog.js
```diff
@@ -126,12 +126,20 @@ aria.Utils = aria.Utils || {};
...
 
-    if (this.dialogNode === null || !isDialog) {
+    var validRoles = ['dialog', 'alertdialog'];
+    var isDialog = element.getAttribute('role')
+      .trim()
+      .split(/\s+/g)
+      .some(function (token) {
+        return validRoles.some(function (role) {
+          return token === role;
+        });
```
jkva: Travis JS linter build seems to break on this indentation size.

examples/dialog-modal/js/dialog.js
jkva: Potentially you could shorthand this into a one-liner
`(this.dialogNode = document.getElementById(dialogId)) || throw new Error('No element found with id="' + dialogId + '".');`

sh0ji: 🤦‍♂️ fixed that typo.

I like your error handler suggestion, but am going to leave it as-is for consistency's sake&mdash;most other errors are thrown with explicit if/else patterns.

examples/dialog-modal/js/dialog.js
```diff
@@ -126,9 +126,20 @@ aria.Utils = aria.Utils || {};
...
   aria.Dialog = function (dialogId, focusAfterClosed, focusFirst) {
     this.dialogNode = document.getElementById(dialogId);
-    var role = this.dialogNode.getAttribute('role');
+    if (this.dialogNode === null) {
+      throw new Error('No element found with id="' + dialogId + '".');
+    }
 
-    if (this.dialogNode === null || !['dialog', 'alertdialog'].includes(role)) {
+    var validRoles = ['dialog', 'alertdialog'];
+    var isDialog = (element.getAttribute('role') || '')
```
jkva: @sh0ji I think this is failing the build because you're referring to `element`, should that not be `this.dialogNode`?

sh0ji: Ha yes, thanks for the quick checks!

mcking65:
Thank you team for this bug fix and thorough review!

PR #947, opened 2018-11-25
Issue 752 example index

jongund:
Updated the example index as a separate document using ReSpec.
Updated the generation of the example index to not use Glob, only uses fs (FileSystem) for node.


examples/example-index.js
```diff
@@ -0,0 +1,423 @@
+var fs = require('fs');
+var i;
+
+var fileNameTemplate = 'index.template';
+var fileNameIndex = 'index.html';
+
+ariaRoles = [
```
nschonni: `var = ariaRoles` unless it's supposed  to be a global variable

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+  'rowgroup',
+  'rowheader',
+  'separator',
+  'table',
+  'term',
+  'toolbar',
+  'tooltip'
+];
+
+ariaPropertiesAndStates = [
```
nschonni: `var = ariaPropertiesAndStates` unless it's supposed  to be a global variable

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+  'aria-labelledby',
+  'aria-owns',
+  'aria-posinset',
+  'aria-rowcount',
+  'aria-rowindex',
+  'aria-rowspan',
+  'aria-setsize'
+];
+
+indexOfRoles = [];
```
nschonni: Global variable?

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+addLandmarkRole('region', true, 'Region Landmark', 'http://localhost/GitHub/aria-practices/examples/landmarks/region.html');
+addLandmarkRole('search', true, 'Search Landmark', 'http://localhost/GitHub/aria-practices/examples/landmarks/search.html');
+
+var exampleIndexFile = fs.readFileSync(fileNameTemplate, function (err) {
+  console.log('Error reading aria index:', err);
+});
+
+var sorted = [];
+
+for (role in indexOfRoles) {
```
nschonni: `sorted = indexOfRoles.slice();`

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+  }
+  html += '      </td>\n';
+  html += '    </tr>\n';
+}
+
+exampleIndexFile = replaceSection('examples_by_role_tbody', exampleIndexFile, html);
+
+sorted = [];
+
+for (prop in indexOfPropertiesAndStates) {
```
nschonni: `sorted = indexOfPropertiesAndStates.slice();`

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+  'aria-owns',
+  'aria-posinset',
+  'aria-rowcount',
+  'aria-rowindex',
+  'aria-rowspan',
+  'aria-setsize'
+];
+
+indexOfRoles = [];
+indexOfPropertiesAndStates = [];
```
nschonni: Global variable?

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+    console.log('Replacing at: ' + indexStart + ' .... ' + indexEnd);
+
+    content = content.slice(0, indexStart) + newContent + content.slice(indexEnd);
+  }
+
+  return content;
+}
+
+function getTitle (data) {
+  title = data.substring(data.indexOf('<title>') + 7, data.indexOf('</title>'));
```
nschonni: `var title = ...`

examples/example-index.js
```diff
@@ -0,0 +1,423 @@
...
+
+var sorted = [];
+
+for (role in indexOfRoles) {
+  sorted.push(role);
+}
+
+sorted.sort();
+
+html = '';
```
jkva: In places where HTML is generated via string concatenation, consider using a JS template literal with some utility functions, such as

```javascript
const html = sorted.reduce((role, set) => {
  const examples = indexOfRoles[role];

  let examplesHTML = '';
  if (examples.length === 1) {
    examplesHTML = `<a href="${examples[0].ref}">${examples[0].title}</a>`;
  else {
    const exampleListItem = item => `<li><a href="${item.ref}">${item.title}</a></li>`;
    examplesHTML = `<ul>${examples.map(exampleListItem).join('\n')}</ul>`;
  }

  return set + `
    <tr>
      <td><code>${role}</code></td>
      <td>${examplesHtml}</td>
    </tr>
  `;
}, '');

mcking65:
Wow @jongund, this index is super useful!!

One side effect of respec is that it changes the heading content, which then makes for some very confusing table captions:

>1. Examples by Role §
>2. Examples By Properties and States §

I was particularly confused by the word "section", which is the accessible name Chrome calculates for the permalink. It gets included in the table caption along with the section number. Somehow it caused me to overlook the presence of the table ... then I looked at the code and understood.

mcking65:
I see you are trimming the word "example". If the title starts with "example of ", like with the tabs examples, we should trim the " of " as well.

mcking65:
H@jongund, We probably want the example-index.js and index.template files in the root /scripts folder along with travis-deploy.sh, or perhaps in the root folder itself. I think it would be good to name the output file attribute-index.html instead of index.html. We have a plan to change the name of aria-practices.html to index.html.

@michael-n-cooper may have an opinion on this.

michael-n-cooper:
I'm afraid I can't tell what you're doing enough to have an opinion, I'm not a code whiz and there are a lot of commits in this pull request to try to figure out. I would have an opinion about the locations of files named index.html, the naming of files meant to serve as indices, etc. But I can't offer an opinion on this specific issue at this time.

mcking65:
Jon, We do not have to use respec for it to a file in the examples directory. Just like none of the example pages are respec docs, the attribute-index.html file does not have to be a respec doc.

mcking65:
I will ask Michael tomorrow if having the index in a separate file is better from a build and publication process perspective.

mcking65:
@jongund, that sounds great.

I did discuss this with Michael today. Here are his preferences:

1. Leave the name of the file that contains the index as index.html. Michael prefers that we have an index.html in each of the high-level directories. So, ignore what I said about changing its name.
2. Do not make the examples/index.html file into a respec doc. We will format it so it looks like it belongs to the rest of the document with an appropriate header and footer, and that is best done without using respec.
3. Put the js in the root /scripts folder. You can also put index.template there, even though it is not a script; we really do not have a better place for it now.

After you have made these changes and considered the other recommendations above, we'll look for advice from @nschonni wrt integrating into the travis build.

Thank you again for taking this one on!!!

jongund:
This pull request can be ignored, please see pull request 951 for the latest version of the index based on feedback from Matt King and Michael Cooper.

Issue #949, opened 2018-12-03
Need to allow for radiogroup inside toolbar


carmacleod:
In https://github.com/w3c/aria-practices/issues/847#issuecomment-416693253 @mcking65 said:
> The only real barrier to using the existing radio group pattern in a toolbar is that the checked state cannot follow focus like in normal radio groups because the arrow keys are needed to navigate to other elements in the toolbar. That means requiring the user to press space to check the radio. Maybe that is OK. 

Personally, I think that is absolutely ok. As a bonus, it's the way the old Windows desktop toolbars work. The arrow keys move the focus without changing the selection, and then space or enter selects the currently focused radio tool. Arrow keys do not wrap within the radiogroup - they simply move focus into the next part of the toolbar.

Testing google docs toolbars, I see that they behave the same as the old Windows desktop toolbars (except that they bogusly don't activate on space - only on enter. I've opened a [Google Docs issue](https://issuetracker.google.com/issues/120382641) for that).

So I'm splitting this out of https://github.com/w3c/aria-practices/issues/218 because I think the specific issue of radiogroups in toolbars needs to be addressed right away so that the [Toolbar example for simple text editor](https://github.com/w3c/aria-practices/pull/887) can be released containing a more correctly semantic "radiogroup" with "radio" children instead of the workaround of a "menu" with "menuitemradio" children.

I think this is not a difficult update to the [radiogroup pattern](http://w3c.github.io/aria-practices/#radiobutton), and there's already well established precedent for the behavior, so perhaps it is as simple as splitting the Keyboard Interaction section into 2 subsections, one with "For a radiogroup that is not inside a toolbar: (bulleted list here)", and the other with "For a radiogroup that is in a toolbar: (other bulleted list here)".

mcking65:
@carmacleod, thank you thank you for the examples of precedence. I was not aware of them.

I personally wanted to make the toolbar radios the way you suggest, but the rest of the task force thought that would be confusing because of the conflict with the radio pattern. But, if we specifically adjust the radio pattern for toolbars, as suggested by this issue, and if we have precedence to justify it, then I am 1000% on board. I would love to get rid of the menuitemradios in the toolbar; they were the only workaround I could dream up that is consistent with all other patterns.

This is going on tomorrow's agenda. Thanks!

jongund:
[...]

PR #952, opened 2018-12-07
Radio Group Pattern: Add new keyboard interaction for radios in toolbars

mcking65:
For issue #949, modified the radio group pattern section aria-practices.html:
* Divided the keyboard interaction subsection into two subsections: not in toolbar and in toolbar.
* Added interactions for the special case of radios contained in toolbars.

Go to [section 3.16 in the feature branch](https://raw.githack.com/w3c/aria-practices/issue949-radios-in-toolbars/aria-practices.html#radiobutton) to see the changes.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/952.html" title="Last updated on Dec 19, 2018, 4:46 AM UTC (8260814)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/952/a1a8d92...8260814.html" title="Last updated on Dec 19, 2018, 4:46 AM UTC (8260814)">Diff</a>

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -2019,8 +2019,7 @@
                   button in the group.</li>
               </ul>
             </li>
-            <li>
-  							<kbd>Space</kbd>: checks the focused radio button if it is not already checked. </li>
+            <li><kbd>Space</kbd>: checks the focused radio button if it is not already checked.</li>
             <li>
               <kbd>Right Arrow</kbd> and <kbd>Down Arrow</kbd>: move focus to the next radio button in
               the group, uncheck the previously focused button, and check the newly focused button. If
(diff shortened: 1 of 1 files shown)


jongund:
I think it looks good, but we I think we also need to define behavior of the up and down arrow keys in the toolbar case.   The up and down arrow keys could be defined only allowing navigation within the radio buttons in the current radio group and could also have the selected radio to follow focus.   The current toolbar example provides most of this up and down arrow functionality, it could be modified for the selected radio button to follow focus.    

Also do we need to provide any guidance on the use of the ENTER key, since it usually is the keyboard behavior for activating the default action of buttons?

PR #962, opened 2018-12-23
chore: Explicitly call git add for gh-pages

nschonni:
Found that the new generated example index wasn't showing up yet because of this https://stackoverflow.com/questions/8470547/git-commit-a-untracked-files
There are also a few files in the `common` folder that were missing https://travis-ci.com/w3c/aria-practices/jobs/166769502#L491-L498


scripts/travis-deploy.sh
```diff
@@ -9,5 +9,6 @@ echo 'Pushing built changes'
 git config --global user.email "cooper@w3.org"
 git config --global user.name "michael-n-cooper"
 git config remote.origin.url "https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git"
-git commit -am "Generated by TRAVIS-CI ${TRAVIS_COMMIT}" -m "${TRAVIS_COMMIT_MESSAGE}"
+git add --all
+git commit -m "Generated by TRAVIS-CI ${TRAVIS_COMMIT}" -m "${TRAVIS_COMMIT_MESSAGE}"
```
mcking65: I am 99% sure we still need the 'a' option here ... the git add command adds untracked files, but to commit all changes in all tracked files, we need '-a' on the commit as well.

nschonni: Think it should be fine without it, but I've added it back anyway

PR #963, opened 2018-12-24
Reference table nodify

nschonni:
- Added ESLint config for the script to allow Node const/let/string literals
- Use `glob` for picking/filtering files
- Use cheerio for content replacement
- Formatting of strings for output HTML

Probably need some windows testing to see if there is anything with the path resolves that might not translate properly from POSIX file paths

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -37,7 +37,7 @@
         <tbody id="examples_by_role_tbody">
           <tr>
             <td><code>article</code></td>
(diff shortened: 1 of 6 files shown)


spectranaut:
New PR to handle fixing merge commit: https://github.com/w3c/aria-practices/pull/1078

nschonni:
No problem, but if you pinged, i would have just rebased

spectranaut:
Whoops next time I'll ping first @nschonni :)

PR #946, opened 2018-11-22
Add Stylelint

sh0ji:
This adds [stylelint](https://stylelint.io/) to help us enforce the CSS style guide. Addresses #932.

Note that I chose to extend [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard) rather than implementing our own style guide. After going through [the CSS code guide](https://github.com/w3c/aria-practices/wiki/Code-Guide#css-1) to see what [stylelint rules](https://stylelint.io/user-guide/rules/) I would need to implement, it became apparent that a lot of them matched with config-standard's opinions. I also vastly prefer the idea of outsourcing code style to other configurations since it's unlikely that we'll ever spend a ton of time maintaining it.

If this seems like a reasonable approach, I'll run `stylelint --fix` to fix any existing issues in our stylesheets before merging. I haven't done that yet to avoid cluttering the list of changes.

***
<a href="https://pr-preview.s3.amazonaws.com/sh0ji/aria-practices/pull/946.html" title="Last updated on Jan 28, 2019, 11:55 PM UTC (c59129e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/946/0f32611...sh0ji:c59129e.html" title="Last updated on Jan 28, 2019, 11:55 PM UTC (c59129e)">Diff</a>

PR #981, opened 2019-02-04
Stylelint 2

sh0ji:
#946 got very messy thanks to my log of fork rebases, so I redid it mostly via copy/paste (exceptions are `stylelint --fix` & the manual fixes, which I wanted to make sure were done on master. This is the new PR for that work. Review #946 for details.

@mcking65, this should be good to go. Note that I accidentally pushed the updated branch directly to `w3c` instead of my own fork because I didn't realize I had permission to do that. It's probably a holdover from when I was managing PRs for regression testing.

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,0 +1,3 @@
+common
+bootstrap.css
+bootstrap-theme.css
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ['stylelint-config-standard']
+};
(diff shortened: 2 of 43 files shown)


mcking65:
@sh0ji commented:

> @mcking65, this should be good to go. Note that I accidentally pushed the updated branch directly to `w3c` instead of my own fork because I didn't realize I had permission to do that. It's probably a holdover from when I was managing PRs for regression testing.

Actually, now that we have protected branches, we prefer you push directly to w3c. That makes review/test much easier. All members of the task force can now push to unprotected branches on w3c and create new branches as well.

Issue #992, opened 2019-02-25
Deployment is canceled because Regression Tests Coverage Report job fails


zcorpan:
Master has not been deployed since https://github.com/w3c/aria-practices/commit/ca78f0a602a80ba904f7b8a83c1d9fc2e461998c (15 Jan) - the next commit https://github.com/w3c/aria-practices/commit/a161bc69c17bad429253cf3b7c99986eeda1fa67 started failing the Regression Tests Coverage Report job. I believe https://github.com/w3c/aria-practices/issues/965 tracks this. The result is that http://w3c.github.io/aria-practices/ is missing changes (I noticed this because I found a broken link, which [has been fixed in master](https://github.com/w3c/aria-practices/commit/e040bfd4980e47f52b7274e828d1142bfeb444dd)).

See https://github.com/w3c/aria-practices/runs/50007038 for the Travis CI result of the change that regressed the build status.

Do we want to mark this job as [allowed to fail](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail)?

Further, do we want to block merging PRs to master when checks have failed? This can be done in the project's settings (under Branches).

mcking65:
@zcorpan commented:

>Further, do we want to block merging PRs to master when checks have failed? This can be done in the project's settings (under Branches).

I have enough confidence in our process now, especially with the fix in #993 that I agree this is a good idea. I'll ask Michael to make the change.

zcorpan:
I believe this was fixed back in February.

PR #993, opened 2019-02-26
Add allow_failures to Travis

sh0ji:
Addresses #992.

This should allow jobs to fail when they match certain criteria, allowing us to make some jobs non-blocking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ git:
   depth: 3
 
 stages:
-- Lint
 - Test
 - Deploy
 
@@ -20,13 +19,13 @@ jobs:
     - env: ALLOW_FAILURE=true
 
   include:
-    - stage: Lint
+    - stage: Test
       name: CSS Linting
       script: npm run lint:css
-    - stage: Lint
+    - stage: Test
       name: JS Linting
       script: npm run lint:es
-    - stage: Lint
+    - stage: Test
       name: HTML Linting
       script: npm run vnu-jar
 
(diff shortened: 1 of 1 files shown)


nschonni:
wonder if it is possible to make it required for master PRs. That wouldn't fix the linked issue, but would mean that topic branches could keep going, but need to address the issues before merging into the mainline.

sh0ji:
From what I can tell, something like the following would allow jobs that aren't part of a pull request (i.e. branch jobs) to fail, but the PR jobs wouldn't be allowed to fail since they don't match envs perfectly.

```yaml
allow_failures:
  - env:
    - ALLOW_FAILURE=true
    - TRAVIS_PULL_REQUEST=false
```

I'd rather not implement that, though. The "allowed to fail" jobs still fail and maintainers can still ask contributors to resolve those issues if it is necessary.

And probably more importantly, we don't want regression coverage to be blocking, though. Getting the examples out is more important than ensuring 100% coverage.

sh0ji:
I split the linting jobs into their own stage, which I may revert once I see how long everything takes.

sh0ji:
Yep, splitting linting/testing jobs doesn't seem worth it. My original thinking was that it might be better to run the cheaper tests first (each linting job takes 40s-2 minutes), but the overall impact doesn't seem worth it, especially since there isn't a build stage.

@mcking65 this should be good to go once everything passes.

mcking65:
Thank you @sh0ji and @zcorpan; Looks great! Excellent improvement!

Issue #986, opened 2019-02-10
Toolbar Example: Add missing descriptive tooltips


mcking65:
Per [comment in issue 539](https://github.com/w3c/aria-practices/issues/539#issuecomment-454870090)
by @sh0ji about the [toolbar example](http://w3c.github.io/aria-practices/examples/toolbar/toolbar.html),
all the controls should have a descriptive title attribute so sighted users don't have to infer or discover their function.
Currently, some have tooltips and some do not.

lenoraporter:
Anyone working on this task?

ardnahcivar:
@lenoraporter , yes .

carmacleod:
@jongund Here's what I see on focus (popup label's "tail" is perfectly joined to the text bubble):
![image](https://user-images.githubusercontent.com/3331913/59712336-0177bb00-91db-11e9-9e11-6447341f89d4.png)

And here's what I see on hover (popup label's "tail" is slightly separated from the text bubble):
![image](https://user-images.githubusercontent.com/3331913/59712231-d1301c80-91da-11e9-9e11-b4b98b3c4691.png)

I'm on Windows 10, latest Chrome. I'll see if I can debug it.

carmacleod:
@jongund 
I can make the above problem go away by making any one of the following 3 CSS changes.
Your choice which one you like best.

**1.** change the 2px white border to be 1px in:
```css
[role="toolbar"] button .popup-label {
    ...
    border: 1px solid white;
    ...
}
```

**OR
2.** change the top/bottom padding from 2px to 1px in:
```css
[role="toolbar"] button .popup-label {
    ...
    padding: 1px 4px;
    ...
}
```

**OR** (this one is a bit strange... ;)
**3.** change the top from 100% to 99% in:
```css
[role="toolbar"] button .popup-label:after, [role="toolbar"] button .popup-label:before {
    top: 99%;
    ...
}
```

jongund:
@carmacleod Carolyn thanks for the fixes I like the first option the best, I will make the change.

jongund:
@carmacleod Do you want to check the updated example:
(Toolbar Example from GitHub branch)[https://raw.githack.com/w3c/aria-practices/issue986-toolbar-button-tooltips/examples/toolbar/toolbar.html]

carmacleod:
@jongund Works perfectly! Thank-you!

mcking65:
[...]

Issue #127, opened 2016-10-16
Develop example of tooltip design pattern


mcking65:
The tooltip design pattern is at:
http://w3c.github.io/aria-practices/#tooltip

Veyfeyken:
Any chance this will go live in the near future? I'm looking for an accessible example to point to.

Malvoz:
@Veyfeyken 

While you are waiting for the w3c developers to design a tooltip pattern, you might want to have a look at [Heydon Pickering](https://github.com/Heydon)'s [**tooltip component**](https://inclusive-components.design/tooltips-toggletips/).

MichielBijl:
Work on this is being done [on CodePen](https://codepen.io/Moiety/pen/LaPvWy). You can also view the [debug page](https://s.codepen.io/Moiety/debug/LaPvWy) which is a lot easier to navigate with AT.

Current open issues with the implementation:

- [ ] Doesn’t detect window bounds
- [ ] Doesn’t conform to [WCAG 2.1 SC 1.4.13 Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus)

Any help is greatly appreciated.

krixian:
@MichielBijl It's looking good! Is there any description on how overflowing tooltips should behave?

weboverhauls:
Yes, looking good! Would also be great to have a tooltip example on something other than a button such as a link/anchor, text input, and possibly regular text.

mfairchild365:
@MichielBijl that is a great example. One question immediately comes to mind: the example uses a `<button>` element to trigger the tooltip. Because the tooltip is automatically displayed when the button recieves both hover and focus events, activating the button element doesn't actually do anything. My hunch is that users might find a button that doesn't do anything a bit confusing and unexpected.

As indicated in #128 the presence of a button might indicate a 'dialog toolip' pattern rather than a traditional tooltip. In any case, I think it might be worth while to think about what the role of the of the element that triggers the tooltip should be. My gut says that no role would be sufficent, but perhaps I'm way off base.

MichielBijl:
> @MichielBijl It's looking good! Is there any description on how overflowing tooltips should behave?

Do you mean when there’s more text than fits ion the tooltip?

> Yes, looking good! Would also be great to have a tooltip example on something other than a button such as a link/anchor, text input, and possibly regular text.

That’s a good idea! I will add that.

[...]

PR #998, opened 2019-03-08
Toolbar Example: Add missing descriptive tooltips #986 fix

ardnahcivar:
added the descriptive title for the toolbar buttons.


mcking65:
I'll get a reviewer assigned today. Question about changes starting at line 213, are they white space changes only? The replacement line sounds the same as the original to me.

ardnahcivar:
> I'll get a reviewer assigned today. Question about changes starting at line 213, are they white space changes only? The replacement line sounds the same as the original to me.

yes at line 213 onwards,  it is white space changes only

jongund:
@ardnahcivar @mcking65 
I don't think we want to do it this way, it you want the tooltips with the title attribute we need to remove aria-label, otherwise we are setting up a situation where the accessible name will be the same as the accessible description.  This also needs to be documented in the example descriptions for use of  properties and states.

ardnahcivar:
> @ardnahcivar @mcking65
> I don't think we want to do it this way, it you want the tooltips with the title attribute we need to remove aria-label, otherwise we are setting up a situation where the accessible name will be the same as the accessible description. This also needs to be documented in the example descriptions for use of properties and states.

@jongund , any pointers, sugesstions  how to fix it

carmacleod:
Hi, @ardnahcivar! Thank-you so much for your pull request!

The group discussed this yesterday, and we decided that we prefer not to use the `title` attribute because sighted keyboard users cannot see browser tooltips, so `title` is not the best choice for accessible tooltips.

The W3C HTML spec for the [title attribute](http://w3c.github.io/html/dom.html#the-title-attribute) mentions this:
> **Warning!** Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device, such as a mouse, to cause a tooltip to appear excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

We believe the direction should be similar to @ZoeBijl's [example tooltip with aria-labelledby](https://github.com/w3c/aria-practices/issues/127#issuecomment-467152945) (in CodePen).

Do you want to have a go at writing this for the toolbar example?

jongund:
@mcking65 Is this pull request going to be merged soon?

mcking65:
@ardnahcivar, sorry we chose another path, which @jongund implemented in #1069. Closing this PR.

If you are available for other bugs, please comment in the relevant issue so we can talk about the approach to resolution. We don't want people to get discouraged from contributing because of situations like this where they offer some work and then we re-do it or go down another path.

PR #1001, opened 2019-03-17
Make clicking padding move focus into toolbar fixed

ardnahcivar:
toolbar gets focused on click.


ardnahcivar:
failed the test cases

Issue #34, opened 2016-04-30
Develop example of date picker design pattern using a combobox that opens a dialog


ckundo:
Add a mobile-compatible example of a date picker that employs the form of the [ARIA 1.1 combobox pattern](http://w3c.github.io/aria-practices/#combobox) that pops up a dialog.

Completing this issue will also complete issue #534.

Development is in the issue34-date-picker-examples branch. Use that branch as the base for pull requests related to this issue.

#### Preview link for feature branch

[Date picker combobox example in feature branch](https://raw.githack.com/w3c/aria-practices/issue34-date-picker-examples/examples/datepicker/datepicker-combobox.html)

Related to issue #57.

ckundo:
I'm taking a pass at this now.

ckundo:
FWIW, there's a solid example of this marked up already at http://www.oaa-accessibility.org/example/15/. Is it worth reaching out to the author to reuse their example here?

jnurthen:
@jongund @a11ydoer is the code at http://www.oaa-accessibility.org/example/15/ either of yours?

mcking65:
Here is the [initial cut of an example page]((https://raw.githack.com/w3c/aria-practices/issue34-date-picker-examples/examples/datepicker/datepicker.html)) by @jongund merged into the issue34-date-picker-examples branch with commit 5d16101.

Rather than immediately tackling issue #57, this example could work as an example for the combobox pattern. That pattern needs an example of a combobox that pops up a dialog.

jongund:
@mcking65 I have pushed a new version of the datepicker example. I had to rewrite most of the code, but it should be much easier to make modifications to the example with the new code base.

jongund:
Link to the  new example:
https://raw.githack.com/w3c/aria-practices/issue34-date-picker-examples/examples/datepicker/datepicker.html

mcking65:
Per discussion in today's meeting, issues that need further discussion:

1. Where should focus go after a date is chosen from the grid in the dialog? If it goes to the text field, that pops open mobile keyboards. If it goes to the down arrow icon, it is confusing for screen reader and keyboard users who moved focus into the dialog from the text field with down arrow.

2. When should the dialog be visually displayed? It is very common for calendars to display when the text field is focused or clicked. Does that mean the user has to move focus out of the input to get the calendar to disappear?

accdc:
Hi,
[...]

PR #1017, opened 2019-04-16
Add combobox date picker example for issue 34

jongund:
Adds the combobox datepicker example described in issue #34.

#### Preview Link

[Preview of date picker combobox example](https://raw.githack.com/w3c/aria-practices/issue34-date-picker-examples/examples/combobox/combobox-datepicker.html)

### Review checklist

* [ ] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @mcking65 @carmacleod 
* [ ] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @carmacleod
* [ ] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @curtbellew 
* [ ] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @zcorpan)
* [ ] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @zcorpan)


examples/datepicker/datepicker.html
```diff
@@ -0,0 +1,820 @@
...
+           aria-owns="myDatepickerDialog"
+           aria-label="Date">
+
+        <div class="date">
+          <label for="id-date-1">Date</label>
+
+          <input type="text"
+                placeholder="mm/dd/yyyy"
+                id="id-date-1"
+                aria-expanded="false"
```
mcking65: As specified in 1.1 combo pattern, please remove aria-expanded from this input and add aria-autocomplete="none"

examples/datepicker/datepicker-combobox.html
```diff
@@ -0,0 +1,945 @@
...
+          <span class="arrow down">
+            <svg width="18" height="18">
+              <polygon points="4,9 14,9 9,16" style="fill:gray;stroke:gray;stroke-width:1" />
+            </svg>
+          </span>
+
+          <button class="icon"
+            tabIndex="-1"
+            aria-label="Change Date"
+            aria-haspopup="true"
```
smhigley: Should this be `aria-haspopup="dialog"` to match the popup role?

examples/datepicker/datepicker-combobox.html
```diff
@@ -0,0 +1,945 @@
...
+          </div>
+
+          <table
+            id="myDatepickerGrid"
+            class="dates"
+            role="grid"
+            aria-labelledby="id-dialog-label">
+            <thead>
+              <tr>
+                <th scope="col" aria-label="Sunday">Su</th>
```
smhigley: Is aria-label on a column header expected to affect the calculated column header name for individual grid cells? E.g. should navigating a table in a screen reader's scan mode read "Sunday" or "Su" when moving to a cell in this column? I'm getting pretty different results when testing.

examples/datepicker/js/datepicker.js
```diff
@@ -0,0 +1,809 @@
...
+
+  if (flag) {
+    event.stopPropagation();
+    event.preventDefault();
+  }
+};
+
+DatePicker.prototype.moveToNextYear = function () {
+  this.year++;
+  this.updateGrid();
```
smhigley: Setting a single date cell to `tabindex="0"` should probably be handled in the `updateGrid` function instead of `updateFocus`, since currently moving to the next or previous year or month only calls `updateGrid`, which then makes the date grid inaccessible via the keyboard.

examples/datepicker/datepicker-combobox.html
carmacleod: It should read "Sunday" according to the [HTML-AAM section on tr, td, th elements](https://w3c.github.io/html-aam/#tr-td-th-elements-accessible-name-computation).
Which AT/browser pairing(s) are reading "Su" for you? Might need to open some issues...

jnurthen: HTML-AAM describes the accessible name for the header cell itself. I don't see anything in it which indicates what the header name is when reading an associated cell in the table. 
If the section in HTML-AAM is intended to cover that then it is missing the abbr attribute which should certainly be a part of that calculation.
Indeed it may be better to use abbr in this example so this would become
 ```
<th scope="col" abbr="Sunday">Su</th>
```

**Comment by carmacleod** [comment]
Regarding whether the Enter key should drop down the calendar, the Windows 10 UX guidelines says yes:
> The Enter key can perform a variety of common user interactions, depending on the control with focus:
> - Displays the picker UI for controls such as ComboBox and DatePicker. The Enter key also commits and closes the picker UI.

Old win32 doc also said that Enter should drop down the calendar.
That's only one platform, however.

I tried to figure out what happens on Mac, but they don't mention keyboard in the [Human Interface Guidelines for Date Picker](https://developer.apple.com/design/human-interface-guidelines/macos/selectors/date-pickers/). Maybe someone on a Mac can try it.

I also tried `<input type="date">` on Windows browsers; they work differently from desktop:
- Chrome has 3 spinbuttons and a dropdown button all in the tab order; user can type up/down on the spinbuttons; and enter on the button opens calendar
- Firefox just has the 3 spinbuttons in tab order; up/down work on spinbuttons; [but Firefox does not have a way to popup calendar with keyboard](https://bugzilla.mozilla.org/show_bug.cgi?id=260121)

File: examples/datepicker/datepicker-combobox.html (hunk 3 above)
Comment (carmacleod) [reply examples/datepicker/datepicker-combobox.html]: > I don't see anything in it which indicates what the header name is when reading an associated cell in the table.

Good point.

> `<th scope="col" abbr="Sunday">Su</th>`

Interesting - I did not know about abbr _attribute_ on th. It's exactly what is needed here.
From the [HTML spec for th](http://w3c.github.io/html/tabular-data.html#the-th-element):
> The th element may have an abbr content attribute specified. Its value must be an alternative label for the header cell, to be used when referencing the cell in other contexts (e.g., when describing the header cells that apply to a data cell). It is typically an abbreviated form of the full header cell, but can also be an expansion, or merely a different phrasing.

I see it in the [HTML-AAM Attribute Table for th attr](https://w3c.github.io/html-aam/#att-abbr), but I don't really understand what it is saying.

Comment (jnurthen) [reply examples/datepicker/datepicker-combobox.html]: @scottaohara Any idea what  https://w3c.github.io/html-aam/#att-abbr means?

File: examples/datepicker/js/datepicker.js (hunk 4 above)
Comment (jongund) [reply examples/datepicker/js/datepicker.js]: The same cell has focus after moving by month or year, so no need to update the focus or the `tabindex` value

File: examples/datepicker/datepicker-combobox.html (hunk 3 above)
Comment (jongund) [reply examples/datepicker/datepicker-combobox.html]: changed `<th scope="col" aria-label="` to `<th scope="col" abbr"`

**Comment by jongund** [comment]
@carmacleod Added feature so mouse clicks on "disabled" dates now changes the month and day selected.

**Comment by carmacleod** [comment]
Found a keyboard/focus bug. Here are the steps:
- type `Down Arrow` to focus a date in the calendar
- type `Shift+Tab` to focus on any of the "next/previous month/year" buttons at the top
- type `Space` to change the month or year

At this point, the calendar date area does not take focus any more (because the "current date" is no longer visible since it was in a different month or year), so a keyboard user can't choose a date in another month or year.

Not quite sure what the best fix is; maybe focus the 1st of the month?

**Comment by smhigley** [comment]
Just a small thing I noticed: it looks like the input steals focus when the calendar dropdown closes, even if it closes because the user clicked somewhere else on the page.

You can test by creating a sample button below the datepicker, opening the datepicker, then clicking on the sample button. The datepicker closes, but then focus moves pack to the input.

**Comment by carmacleod** [comment]
Thanks @jongund ! The next/previous year/month fix works great. I even tried some edge cases, like setting the date to May 31 and then switching to June (which only has 30 days), and the correct thing happened (i.e. same cell selected). Nice!

**Comment by mcking65** [comment]
This example is now moved to the directory for ARIA 1.1 combobox examples. I have updated the top comment with the new preview link.

The spinbutton and dialog datepicker examples have ben moved to separate branches. @jongund will be opening new draft pull requests for those examples.

**Comment by jongund** [comment]
@mcking65 I think this example should be modified to use the aria 1.0 combobox pattern, since it is closer to what you are proposing for the 1.2 pattern

**Comment by jongund** [comment]
@spectranaut 

Valerie I have been updating the documentation and regression tests for the combox date picker example.   I have imported the regression tests you created for date picker dialog example.   I am getting a failing test on line 177 of the regression test script related to dynamically changing `tabindex` values.   Can you look error messages and tell me what needs to be changed to fix the problem.   I would like to make the fix so I can learn more about the regression testing code.

**Comment by jongund** [comment]
@spectranaut 

Valerie,
I am on vacation the rest of this week, and I think this is the last week you will be working on the practices for awhile.   So if you have time can you complete the test for this combobox example, and I will look at the test when I get back.

**Comment by jongund** [comment]
@mcking65 

Updated pull request:
* Moved `combobox` role to `input` element
* Moved `aria-expanded` to `input` element
* Update `aria-controls` on `input` element
* Starting working on test cases

**Comment by jongund** [comment]
@mcking65 
The combobox date picker example needs a rewrite based on the changes in the menu burron datepicker dialogbox example, so please do not review the current code.   It also needs to be updated for the Combobox 1.2 pattern.

**Comment by mcking65** [comment]
@jongund, I'll change the title of this PR then to reflect that this is still work in progress. When we are ready for reviews, we'll change it back.

**Comment by jongund** [comment]
@mcking65 

I have rewritten the example to conform with the preferred coding practices (e.g. one  prototype/class) and the combobox 1.2 pattern.   I have also update the initial test cases, but I more work needs to be done on the tests.  

The updated example is ready for initial review.

**Comment by jongund** [comment]
@jnurthen @mcking65 

This example uses `aria-selected` on a `button` element, which fails validation and is not allowed in the aria spec.  What should happen to `aria-selected` attribute in this example to identify the selected date in the grid?

**Comment by jongund** [comment]
@carmacleod  @smhigley @mcking65 

I made the changes to remove the button from the `girdcell`, so the date picker is ready for review.  Seems to work well with Jaws and NVDA, and initial testing in high contrast looks alright.

Jon

File: examples/combobox/combobox-datepicker.html [hunk 5]
```diff
@@ -0,0 +1,835 @@
...
+              </tr>
+              <tr>
+                <td tabindex="-1" data-date="2020-02-30">30</td>
+                <td tabindex="-1" data-date="2020-02-31">31</td>
+                <td class="disabled" tabindex="-1"></td>
+                <td class="disabled" tabindex="-1"></td>
+                <td class="disabled" tabindex="-1"></td>
+                <td class="disabled" tabindex="-1"></td>
+                <td class="disabled" tabindex="-1"></td>
+              </tr>
```
Cerberooo: You are creating this dynamically in the `combobox-datepicker.js` so it needs no default in the html I think.

jongund: I believe having a HTML view is useful for people to see the structure of the data picker grid, not everyone using the practices will be able to read and understand Javascript.

jongund:
@mcking65 @carmacleod 

I fixed the bug with aria-select following datepicker focus.

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+  <main>
+    <h1>Date Picker Combobox Example</h1>
+    <p>
+      <strong>NOTE:</strong> This example page is work in progress.
+      Please provide feedback in
+      <a href="https://github.com/w3c/aria-practices/pull/1017">pull request 1017.</a>
+    </p>
+    <p>
+      The date picker is an example of the <a href="../../#combobox"><code>combobox</code> design pattern</a> that opens a dialog box.
+      The date picker dialog box in this example is opened when keyboard focus is moved to the text box and a <kbd>down arrow</kbd> key pressed or clicking on the choose date button.
```
carmacleod:
```suggestion
      The date picker dialog box in this example is opened by moving keyboard focus to the text box and pressing either the <kbd>enter</kbd> key, <kbd>down arrow</kbd> key or <kbd>alt + down arrow</kbd> keys or by clicking on the choose date button.
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+      <table aria-labelledby="kbd_label_1 kbd_label" class="def">
+        <thead>
+          <tr>
+            <th>Key</th>
+            <th>Function</th>
+          </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <th><kbd>Down Arrow</kbd>, <kbd>Enter</kbd></th>
```
carmacleod:
```suggestion
                <th><kbd>Down Arrow</kbd>, <kbd>ALT + Down Arrow</kbd>, <kbd>Enter</kbd></th>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+    <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
+  </section>
+
+  <section>
+    <h2>Accessibility Features</h2>
+    <p>There are three main featues of the combobox date picker example:</p>
+    <ul>
+      <li>Textbox: Holds the value of the selected date and can open the date picker dialog box.</li>
+      <li>Choose Date Button: Opens the datepicker dialog box.</li>
+      <li>Date Picker Dialog Box: Displays a grid of dates for the user to select from and provide additional buttons to change the month and year of dates shown in the grid.</li>
```
carmacleod:
```suggestion
      <li>Date Picker Dialog Box: Displays a grid of dates for the user to select from and provides additional buttons to change the month and year of dates shown in the grid.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+
+    <p>The <code>combobox</code>, <code>textbox</code> and <code>button</code> all have the same accessible name.</p>
+
+
+    <h3>Date Textbox</h3>
+
+    <ul>
+      <li>Contains the date value.</li>
+      <li>Opens and closes the date picker dialog box through keyboard, click and touch events.</li>
+      <li>A live region provides information on how to open the date picker dialog when textbox receives focus (e.g. using <kbd>down arrow</kbd> key).</li>
```
carmacleod: Is it really necessary to have a live region to say how to open the combobox? Screen readers know that it's a combobox, so they can choose to say the usual combobox keyboard hint. Enter and down arrow and alt+down arrow all work.

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+      <li>To support people who are sighted in understanding the <kbd>down arrow</kbd> key opens the date picker dialog box, a down arrow icon appears in the <code>textbox</code> when the <code>textbox</code> receives keyboard focus and the down arrow icon is removed when the <code>textbox</code> does not have keyboard focus.</li>
```
carmacleod: This "down arrow icon hint" doesn't appear to be working, plus, I don't think it's necessary because the combobox looks like a combobox, so sighted keyboard users will know how to operate it (enter works, as well as down arrow and alt+down arrow).

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+    </ul>
+
+    <h3>Choose Date Button</h3>
+
+    <ul>
+      <li>Opens and closes the date picker dialog box through click events.</li>
+      <li>The accessible name of the button is updated when a date is selected to include the current date.</li>
+      <li>Excluded from the <kbd>tab</kbd> sequence of the page (eg. <code>tabindex=-1</code>), since the user can use the <kbd>down arrow</kbd> key to open the date picker dialog box from the <code>textbox</code>.</li>
```
carmacleod:
```suggestion
      <li>Excluded from the <kbd>tab</kbd> sequence of the page (eg. <code>tabindex=-1</code>), since the user can use the <kbd>down arrow</kbd> or <kbd>enter</kbd> key to open the date picker dialog box from the <code>textbox</code>.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+    </ul>
+
+    <h3>Date Picker Dialog</h3>
+
+    <ul>
+      <li>The date picker dialog is a modal dialog box for selecting a date from a grid of dates for a particular month and year.</li>
+      <li>Additional buttons and keyboard commands are used to change month and year.</li>
+      <li>A live region announces changes in the month and year.</li>
+      <li>A live region announces the use of cursor keys when a date in the gird of dates gets focus.  The message is also visible at the bottom dialog box to provide the same information to keyboard users.</li>
```
carmacleod:
```suggestion
      <li>A live region announces the use of cursor keys when a date in the grid of dates gets focus.  The message is also visible at the bottom of the dialog box to provide the same information to keyboard users.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+      <li>The dialog is opened through keyboard commands in the textbox or clicking on the choose date button.</li>
```
carmacleod:
```suggestion
      <li>The dialog is opened through keyboard commands in the textbox or by clicking on the choose date button.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+    </ul>
+
+    <h3>Mobile Support</h3>
+
+    <ul>
+      <li>One issue with mobile browsers is that the onscreen keyboard is visible when the textbox has focus.</li>
+      <li>Adding <code>touchstart</code> event allows the user to close the onscreen keyboard when the textbox has focus.</li>
```
carmacleod:
```suggestion
      <li>Adding the <code>touchstart</code> event allows the user to close the onscreen keyboard when the textbox has focus.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+            <th>Key</th>
+            <th>Function</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr data-test-id="grid-space-return">
+            <th><kbd>Space</kbd>,<br><kbd>Enter</kbd></th>
+            <td>
+              <ul>
+                <li>Select the date, close the dialog, and move focus to the <q>Choose Date</q> button.</li>
```
carmacleod:
```suggestion
                <li>Select the date, close the dialog, and move focus to the textbox.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+                <li>Changes the grid of dates to the previous month.</li>
+                <li>Sets focus on the same day of the same week. If that day does not exist, then moves focus to the same day of the previous or next week.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="grid-shift-pageup">
+            <th><kbd>Shift</kbd>+<br> <kbd>PageUp</kbd></th>
+            <td>
+              <ul>
+                <li>Change the grid of dates to the previous Year.</li>
```
carmacleod:
```suggestion
                <li>Change the grid of dates to the previous year.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+                <li>Change the grid of dates to the next month.</li>
+                <li>The focus will be on the same day of that week, if it does not exist, then it will move the focus to the same day of previous or next week</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="grid-shift-pagedown">
+            <th><kbd>Shift</kbd>+<br> <kbd>PageDown</kbd></th>
+            <td>
+              <ul>
+                <li>Change the grid of dates to the next Year.</li>
```
carmacleod:
```suggestion
                <li>Change the grid of dates to the next year.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+                <li>The focus will be on the same day of that week, if it does not exist, then it will move the focus to the same day of previous or next week</li>
+              </ul>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </section>
+
+    <section>
+        <h3 id="kbd_label_6">Date Picker Buttons (OK and Cancel)</h3>
```
carmacleod:
```suggestion
        <h3 id="kbd_label_6">Date Picker Dialog: OK and Cancel Buttons</h3>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+                <th>Function</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr data-test-id="okay-cancel-button-space-return">
+              <th> <kbd>Space</kbd>,<br> <kbd>Return</kbd></th>
+              <td>
+                Activates the button:
+                <ul>
+                    <li><q>Cancel</q>: Closes the dialog, moves focus to <q>Date Textbox</q> button, does not update date in date textbox.</li>
```
carmacleod:
```suggestion
                    <li><q>Cancel</q>: Closes the dialog, moves focus to <q>Date</q> textbox, does not update date in date textbox.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+                    <li><q>OK</q>: Closes the dialog, moves focus to <q>Date Textbox</q> button, updates date in date textbox.</li>
```
carmacleod:
```suggestion
                    <li><q>OK</q>: Closes the dialog, moves focus to <q>Date</q> textbox, updates date in date textbox.</li>
```

mcking65:
I'm not sure what is happening with selected ... the first time I loaded, it seemed like selection was not following focus, but now it is.

I noticed that clicking the button does not move focus into the dialog.

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+            <td>
+              Identifes the <code>input</code> element as a combobox.
+            </td>
+          </tr>
+          <tr data-test-id="textbox-aria-haspopup">
+            <td></td>
+            <th scope="row"><code>aria-haspopup=<q>dialog</q></code></th>
+            <td><code>input</code></td>
+            <td>
+              Identifies the combobox opens a dialog box.
```
carmacleod:
```suggestion
              Identifies that the combobox opens a dialog box.
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+            <td></td>
+            <th scope="row"><code>aria-expanded=<q>true</q></code></th>
+            <td><code>input</code></td>
+            <td>
+              <code>aria-expanded=<q>true</q></code> when the date picker dialog box is open.
+            </td>
+          </tr>
+          <tr data-test-id="textbox-aria-autocomplete">
+            <td></td>
+            <th scope="row"><code>aria-autocomplete=<q>none</q></code></th>
```
carmacleod: Should we just not set aria-autocomplete at all, instead of setting to none?

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,835 @@
...
+          </thead>
+          <tbody>
+            <tr data-test-id="calendar-button-tabindex">
+              <td></td>
+              <th scope="row"><code>tabindex=<q>-1</q></code></th>
+              <td><code>button</code></td>
+              <td>
+                <ul>
+                  <li>Exclude the button from tab sequence of the page.</li>
+                  <li>The textbox supports opening the dialog through the Keyboard.</li>
```
carmacleod:
```suggestion
                  <li>The textbox supports opening the dialog through the keyboard.</li>
```

examples/combobox/combobox-datepicker.html
JAWS-test: The input instructions are automatically provided by the screen readers. Under no circumstances they should be repeated in a live region. This is unnecessary noise. I would only use a live region for input fields with autocomplete (without role=combobox), because otherwise it is not obvious when search suggestions are shown.

shirsha:
Recommend to provide date format (mm/dd/yyyy) in label name (ex: Date (mm/dd/yyyy)) instead of having it as placeholder.
The days are narrated as "Tu", "We" instead of Tuesday, Wednesday etc.
Date narration is not intuitive-For example, when focus is on 17th, NVDA narrates as 17 row four tu column three.
The empty day grid cell should be hidden from screen reader user.
Both the input text (role- combobox) and button have aria-expanded attribute and labels. This will confuse the screen reader users.
Button is available to screen reader users in read mode as it has tabindex="-1".

mcking65:
Siri, thank you for the review. I have some questions and comments.

@shirsha commented:

> Recommend to provide date format (mm/dd/yyyy) in label name (ex: Date (mm/dd/yyyy)) instead of having it as placeholder.

Date format is ordinarily considered an instruction, which should be in a description, not the accessible name per the guidance we have provide in the accessible naming section. Do you think there is an overriding reason not to follow that guidance here?

> The days are narrated as "Tu", "We" instead of Tuesday, Wednesday etc.

That is because the screen reader is not supporting the abbr attribute that is on the column headers. I have noticed that most screen readers support it in their reading mode with table reading keys, but then they fall apart when in their interaction mode, e.g., forms or focus mode. We will certainly cover that behavior with the assertions in the aria-at project.

> Date narration is not intuitive-For example, when focus is on 17th, NVDA narrates as 17 row four tu column three.

Like the issue with abbr, this is what NVDA chooses to do with the markup. If it reads the column headers correctly, that should resolve this.

> The empty day grid cell should be hidden from screen reader user.

What is the rationale? It would break the table structure and remove the information that those days belong to the prior or next month, which is conventional for calendars. If the table reading stopped in the middle of a row, that would be pretty confusing.

BTW, we had previously made the decision for those to be blank instead of putting in numbers for the previous or next month. Now that we removed the buttons, that decision is especially important.

> Both the input text (role- combobox) and button have aria-expanded attribute and labels. This will confuse the screen reader users.

I am not sure I understand. Can you explain how this would be confusing? We just added those attributes to the button for touch screen reader users. 

> Button is available to screen reader users in read mode as it has tabindex="-1".

Correct, we want touch screen reader users and reading cursor users to be able to operate the button, but do not want it in the tab sequence, which aligns with the authoring guidance in the combobox spec.

jongund:
@mcking65 
@a11ydoer 

I have updated the regression test to include:
1. opening dialog from the textbox
1. `aria-expanded='true'` on textbox and button when dialog is open
1. `aria-tabindex=-1` for the button to open the dialog

Let me know if there other test are needed.

zcorpan:
When first selecting a date, the button's `aria-label` is changed to include the selected date. (But @mcking65 questions above if that should be done at all.)

If I then press escape or backspace to clear the value, the button's `aria-label` is unchanged.

zcorpan:
I added another test, but these are still missing per `npm run regression-report`
```
combobox/combobox-datepicker.html:
    grid-space
    grid-return
    grid-up-arrow
    grid-down-arrow
    grid-right-arrow
    grid-left-arrow
    grid-home
    grid-end
    grid-pageup
    grid-shift-pageup
    grid-pagedown
    grid-shift-pagedown
    okay-cancel-button-space-return
    textbox-aria-autocomplete
    textbox-aria-live
    calendar-button-aria-haspopup
    calendar-button-aria-controls
```

If we think a row doesn't need a test, it should be annotated with `data-test-id="test-not-required"`.

jongund:
> 
> 
> When first selecting a date, the button's `aria-label` is changed to include the selected date. (But @mcking65 questions above if that should be done at all.)
> 
> If I then press escape or backspace to clear the value, the button's `aria-label` is unchanged.

@zcorpan 

Matt originally requested the button be updated to include the currently selected date.  But I don't think it needs or necessarily should be done for accessibility.  It should be an easy change.

jongund:
@zcorpan 
@mcking65 
I have removed the code to update the button with the current date.

zcorpan:
The tests were inconsistently passing and failing in CI. I think the reason may be that the code wasn't consistently using `await` -- it sometimes used a pattern like
```
  await t.context.session.findElement(By.css(ex.buttonSelector)).sendKeys(Key.ENTER);
```
and elsewhere
```
  let chooseDateButton = await t.context.session.findElement(By.css(ex.buttonSelector));
  await chooseDateButton.sendKeys(' ');
```
I think the latter is correct and the former would continue the script before the `sendKeys` promise had settled, resulting in possibly racy behavior.

Should be fixed in c7fcea4

jongund:
@zcorpan 

I changed the previous and next year test to use key.SPACE and it seems to be passing locally now, it was failing using the ` ` character.   So I hope this test passes on the server.

zcorpan:
@jongund thanks. I used space as a string as per https://github.com/w3c/aria-practices/issues/1358#issuecomment-606966971 but missed that it's an `input` element here.

I've documented the space key press workaround in [the wiki](https://github.com/w3c/aria-practices/wiki/Introduction-to-APG-Regression-Tests#workarounds-for-known-bugs).

jongund:
I made some updates for high contrast support for the dates and buttons in the dialog, can people review and comment on the changes.

jongund:
@zcorpan 
@smhigley 

I have updated the regression test to us `context.queryElements( t, selector)` and `t.plan(n)`, but there are 4 errors after I updated the regression test file.

Can you look at the branch and see what is wrong.

```
 4 tests failed

  combobox_datepicker » combobox/combobox-datepicker.html [data-test-id="textbox-aria-expanded-true"]: Combobox: aria-expanded set to "true" when dialog is open

  node_modules\selenium-webdriver\lib\error.js:550

  Rejected promise returned by test. Reason:

  InvalidArgumentError {
    remoteStacktrace: '',
    message: 'invalid type: map, expected a string at line 1 column 32',
  }



  combobox_datepicker » combobox/combobox-datepicker.html [data-test-id="calendar-button-aria-expanded-true"]: Button: aria-expanded set to "true" when the dialog box is open

  node_modules\selenium-webdriver\lib\error.js:550

  Rejected promise returned by test. Reason:

  InvalidArgumentError {
    remoteStacktrace: '',
    message: 'invalid type: map, expected a string at line 1 column 32',
  }



  combobox_datepicker » combobox/combobox-datepicker.html [data-test-id="gridcell-tabindex"]: Roving tab index on dates in gridcell

  node_modules\selenium-webdriver\lib\error.js:550

  Rejected promise returned by test. Reason:

  InvalidArgumentError {
    remoteStacktrace: '',
    message: 'invalid type: map, expected a string at line 1 column 32',
  }



  combobox_datepicker » combobox/combobox-datepicker.html [data-test-id="gridcell-aria-selected"]: aria-selected on selected date

  node_modules\selenium-webdriver\lib\error.js:550

  Rejected promise returned by test. Reason:

  InvalidArgumentError {
    remoteStacktrace: '',
    message: 'invalid type: map, expected a string at line 1 column 32',
  }

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! aria-practices@0.0.0 regression: `ava --timeout=1m "--match" "*combobox-datepicker*"`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the aria-practices@0.0.0 regression script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\User\AppData\Roaming\npm-cache\_logs\2020-05-05T21_56_52_783Z-debug.log
```

jongund:
@mcking65 
@carmacleod 
Updated the documentation on the high contrast for the dialog buttons and dates for the combobox datepicker.

zcorpan:
@jongund I have fixed the errors with `queryElements`. There were two main things:

* `queryElements` expects a string for the selector, i.e. don't use `By.css`
* In some cases the original was `findElement` (return the first element only), and when changing it to `queryElements` (return all elements as an array), you need to get the first element in the array. I've reported #1403 to make this more ergonomic.

I've also fixed spelling errors found by `npm run lint:spelling`, and added stubs for missing tests so that `npm run regression-report` doesn't list new missing tests.

jongund:
@zcorpan 

Thanks for making the fixes to the regression tests!

mcking65:
As discussed today, merging into a new base feature branch from which we can create a clean PR.

Issue #1007, opened 2019-03-23
Image Carousel Example: Use button elements for slide controls and make rotation control always visible


jongund:
Based on feedback in issue #971, revise the [Feb 7, 2019 version of the Auto-Rotating Image Carousel Example](https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190207/examples/carousel/carousel-1/carousel-1.html)
so the pause button is always visible and the slide controls are implemented with button elements.

### Preview Link

The revised carousel can be seen in the [issue1007-visible-pause-carousel branch](https://raw.githack.com/w3c/aria-practices/issue1007-visible-pause-carousel/examples/carousel/carousel-1/carousel-1.html).

mcking65:
Now resolved with commit 0ba3c1c. Thank you @jongund!!

Issue #971, opened 2019-01-15
Collect feedback on auto-rotating image carousel example introduced in APG 1.1 Release 3 (January 2019)


mcking65:
Collect feedback on the [Feb 7, 2019 version of the Auto-Rotating Image Carousel Example](https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190207/examples/carousel/carousel-1/carousel-1.html)
and make updates to the [editors draft of the example](http://w3c.github.io/aria-practices/examples/carousel/carousel-1/carousel-1.html) for the July 2019 publication of APG 1.1.

### Feedback Issues

Based on discussion in this issue, the following issues requesting changes to the example have been raised:

- [x] #1007: Use button elements for slide controls and make rotation control always visible

shirsha:
The content over the image, and previous/next button are difficult to see due to the busyness of the image behind it. Add a background color to the buttons and insert CSS color block between the text and the image so that they can be visible even when the background image is busy.
Note: aria-roledescription is not yet supported across the browser and screen reader combination

juliemoynat-tanaguru:
I think that the button to stop the automatic slide show should be visible by default. Actually, if you have an attention-deficit disorder, you're not necessarily using keyboard to navigate in websites. And you can't guess there is a button to stop the animation.

DavidMacDonald:
At first I was intrigued by the idea of pause on hover/focus, but I think we need a visible by default pause button in addition to that functionality. The main reason for pause (WCAG 2.2.2) is so the user with a cognitive disability is not distracted when they are elsewhere on the page. the current pause functionality requires the mouse or keyboard focus to remain on the component, which could be frustrating for someone who wants to do stuff elsewhere on the page without the rotating carousel distracting.

jnurthen:
> At first I was intrigued by the idea of pause on hover/focus, but I think we need a visible by default pause button in addition to that functionality. The main reason for pause (WCAG 2.2.2) is so the user with a cognitive disability is not distracted when they are elsewhere on the page. The current pause functionality requires mouse or keyboard focus to remain on the component, which could be frustrating for someone who wants to do stuff elsewhere on the page without the rotating carousel distracting.

+1 
I love the [Microsoft](https://www.microsoft.com/en-us/) carousel. Can we use that as inspiration?

ArnaudDelafosse:
> 
> 
> ...I think we need a visible by default pause button in addition to that functionality...

[...]

PR #1018, opened 2019-04-16
Image Carousel Example: Use buttons for controls and make rotation control always visible

jongund:
Resolves issue #1007 by making the stop/start rotation button always visible and uses button elements instead of links for the next and previous slide elements.

### Preview Link

The revised carousel can be seen in the [issue1007-visible-pause-carousel branch](https://raw.githack.com/w3c/aria-practices/issue1007-visible-pause-carousel/examples/carousel/carousel-1.html).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1018.html" title="Last updated on Jul 11, 2019, 6:03 AM UTC (e6b4bb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1018/fb18db7...e6b4bb5.html" title="Last updated on Jul 11, 2019, 6:03 AM UTC (e6b4bb5)">Diff</a>


carmacleod:
Please note this issue commend re: size: https://github.com/w3c/aria-practices/issues/971#issuecomment-492719504

test/tests/carousel-1.js
```diff
@@ -15,6 +15,7 @@ const exampleFile = 'carousel/carousel-1/carousel-1.html';
 
 const ex = {
   landmarkSelector: '#myCarousel',
+  buttonSelector: '#ex1 button',
   previousNextButtonSelector: '#ex1 .carousel-control',
```
spectranaut: This class has been removed, so this selector doesn't select anything in the test that uses it!

test/tests/carousel-1.js
```diff
@@ -35,14 +36,9 @@ ariaTest('Carousel 1: section has aria-roledescription set to carousel', example
   await assertAttributeValues(t, ex.landmarkSelector, 'aria-roledescription', 'carousel');
 });
 
-ariaTest('Carousel 1: previous and next buttons have role button', exampleFile, 'carousel-button-role-link', async (t) => {
+ariaTest('Carousel 1: pause, previous and next buttons have aria-label', exampleFile, 'carousel-button-aria-label', async (t) => {
```
spectranaut: The names of the tests are printed like this: `carousel/carousel-1/carousel-1.html [data-test-id="carousel-button-aria-controls"]: Carousel 1: previous and next buttons have aria-controls Rejected promise returned by test`

So `Carousel 1:` can be removed from the beginning of each test name in this file, as the example file being tested is prepended to the name.

test/tests/carousel-1.js
```diff
@@ -15,6 +15,7 @@ const exampleFile = 'carousel/carousel-1/carousel-1.html';
```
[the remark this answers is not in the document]
jongund: sure, you can fix it

test/tests/carousel-1.js
jongund: sure you can remove it

test/tests/carousel-1.js
jongund: we will eventually have another carousel example, but it is not going to make this release, but I guess it will have its own test file, so I think it could be removed to reduce the verbosity of the messages.

spectranaut:
hey @jongund! I add a commit to fix up the existing attribute tests. I also added a couple tests and have two observations:

I noticed a bug with `aria-live`.  `aria-live` is set to `off` on page load, but after interacting with the widget in any way (hovering mouse, moving focus), `aria-level` is set to `polite` and it remains that way after focus leaves the widget and the automatic transitions begin again. Unless you click `pause`, in which case it changes to `off` (which seems to be incorrect according to the table, which states `The live region is off when the carousel is automatically rotating.`). The test for `aria-live` is failing as it test what is described in the table.

You also might notice the test I wrote for `aria-disabled` is failing as well. Using selenium test driver, I'm having trouble replicating moving the focus from the "previous" button to the "pause" button (which should remove `aria-disabled` according to the table) -- I'm not sure if this is a bug in the example, but I can look into it tomorrow morning. Writing this test made me think that the use of `aria-disabled` here could be confusing, as the button will always work if you navigate to it or click it. Specifically, the button will turn the automatic rotation off once you leave the carousel region regardless of whether the automatic transition is currently happen. Maybe there is something more to "aria-disabled" than I realize, but is it really necessary for this widget?

jongund:
@spectranaut I will look into the aria-live issue, it should be pretty easy to fix

The aria-disabled is used instead of disable property so the button stays in tab order for screen reader users.   When focus is in the carousel the carousel automatically pauses, so the pause button is essentially disabled.

spectranaut:
Ok I added more tests for keyboard interaction, and I notice the Keyboard Interaction table needs to be updated now that the "pause" button no longer hides. Other than that it looks good :)

mcking65:
@jongund wrote:

> @spectranaut I will look into the aria-live issue, it should be pretty easy to fix

Excellent, thank you! Since we never want it live when auto-rotating, can the change of aria-live be triggered based on changes to the auto-rotation state regardless of which user event causes rotation to stop or start?

> The aria-disabled is used instead of disable property so the button stays in tab order for screen reader users. When focus is in the carousel the carousel automatically pauses, so the pause button is essentially disabled.

After discussing with Valerie and Jemma, we concluded that we can remove aria-disabled because, in practice, the button is never disabled. It is still operable with the mouse, screen reader reading cursor, or touch cursor even when rotation is paused automatically. And, it is useful to be able to click or tap it even when automatically paused; that causes the carousel to stay stopped.

jongund:
@spectranaut Thank you for adding the regression tests I will make the changes for disabled button and the live region

jongund:
@spectranaut @mcking65 @a11ydoer I made the changes to the code to fix the live region issue, so it is ready for review and hopefully merging

Issue #78, opened 2016-08-13
Draft guidance section for aria-live


mcking65:
When working on this issue, there is some old content that may be useful in the file aria-practices-DeletedSectionsArchive.html.
The relevant section can be seen here:
https://rawgit.com/w3c/aria-practices/master/aria-practices-DeletedSectionsArchive.html#docmgt

jnurthen:
Note https://www.w3.org/Bugs/Public/show_bug.cgi?id=28611 about aria-busy when drafting this section

carmacleod:
Would be great to also have some live region examples.
Here's one for status role: https://carmacleod.github.io/playground/status-role.html

JAWS-test:
I would like to make the following proposal for live regions and ARIA roles with implicit live regions:

### An output of the live region should be made:
* only with aria-live=polite or aria-live=assertive (explicit or implicit) and without aria-busy=true (implicit or explicit)

With aria-relevant=all/additions/text:
* Inserting a node into the DOM (only for aria-relevant=all/additions) 
* Changing or inserting text content in an existing node (only for aria-relevant=all/text)
* Change visibility: 
	* from display:none to display:block/inline/inline-block etc., 
	* from hidden to without hidden
	* from visibility:hidden/collapse to visibility:visible
	* from aria-hidden=true to aria-hidden=false
	* for display:none, visibility:hidden and aria-hidden=true also if the attribute or CSS property is removed
	
With aria-relevant=all/removals:
* Removal of a node from the DOM
* Remove text content
* Change the visibility with
	* display:none
	* hidden
	* visibility:hidden/collapse
	* aria-hidden=true

* Changing from aria-busy=true to aria-busy=false or removing aria-busy=true (but only if one of the above changes has additionally taken place in the time of aria-busy=true)	
	
For all of the above methods, it should work if
* The element with the live region is changed or
* the child or a descendant element or
* the parent or an ancestor element

For all CSS methods (display, visibility) it should work if
* a class is added or removed,
* the class definition is changed,
* the content of the style attribute is changed,
* the style attribute is added or removed.
	
### An output should not occur:
* if the live region is already present in the source code when the page is loaded
* if the ARIA role (or ARIA attributes - see open questions) for live regions are added to the DOM later (i.e. assigned to an existing node)
* aria-busy=true
* aria-live=off
* use of aria-relevant or aria-atomic without aria-live=political/assertive
* when changing visibility with CSS properties other than those described above, such as
	* opacity
	* color
[...]

PR #1027, opened 2019-05-03
Guideline for live regions

zcorpan:
Fixes https://github.com/w3c/aria-practices/issues/78

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1027.html#live_regions" title="Last updated on Mar 3, 2021, 6:13 PM UTC (27b2c17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1027/0d79d21...27b2c17.html" title="Last updated on Mar 3, 2021, 6:13 PM UTC (27b2c17)">Diff</a>


mcking65:
Need to add guidance that explains what types of DOM changes the spec says should trigger live regions. (related to issue #1138)

zcorpan:
I think this is covered under [the `aria-relevant` section](https://github.com/w3c/aria-practices/blob/zcorpan/live-regions/live-regions.md#aria-relevant), but it can be clearer. The ARIA spec says for `additions`:

> Element nodes are added **to the accessibility tree** within the live region.

whereas this PR currently says

> `additions`: Element node additions

I think the descriptions here should match those in the ARIA spec, and then give some examples (e.g. how changing CSS `display` can cause a removal from the accessibility tree).

https://github.com/w3c/aria-practices/pull/1045 introduces a section about accessibility trees, so this section can reference that.

zcorpan:
If there are interoperability bugs in the area for dynamic updates, I think that should be addressed in the ARIA spec (if it needs changes), tests and bug reports.

If there are implementation bugs that make something not suitable as best practice right now, then it seems reasonable to have APG call that out, however. Are there such cases that we should cover in APG?

(I'll wait a bit with converting this to HTML, since this may need more rounds of edit and review.)

zcorpan:
Per https://w3c.github.io/html-aam/#html-element-role-mappings `output` has `role=status` by default, which makes it a live region.

Credit to @scottaohara for this finding: https://www.scottohara.me/blog/2019/07/10/the-output-element.html

I think this section should at least include an example of `output` in the `role=status` section.

JAWS-test:
> I think the descriptions here should match those in the ARIA spec, and then give some examples (e.g. how changing CSS display can cause a removal from the accessibility tree).

I think it would be good if the chapter on Live Region in ARIA practices could be expanded to include more details on the correct implementation. There are several open questions (https://github.com/w3c/aria-practices/issues/78#issuecomment-524619306) and many bugs that can be seen in the implementation, both in the assistive technology and the browsers (for JAWS see https://github.com/FreedomScientific/VFO-standards-support/issues?q=live+region) as well as with the web developers. With WCAG 4.1.3, however, correct implementation is becoming increasingly important.

JAWS-test:
Support for output seems not to be good: https://github.com/FreedomScientific/VFO-standards-support/issues/268

zcorpan:
I agree that correct implementation is important, but I think changing APG is not an effective way to fix bugs in implementations.

APG is about how to use ARIA, not how to implement ARIA. Per http://w3c.github.io/aria-practices/#browser_and_AT_support it also does not describe or work around bugs in the implementation of ARIA.

If you think APG should take a different stance on discussing bugs in implementation, please file a new issue.

JAWS-test:
@zcorpan 

I don't think that APG should be concerned with browser and assistive technology bugs, but should contain good examples and detailed explanations of Live Regions, so that the many bugs of browser manufacturers and developers can be avoided. For example, there is a long chapter on Accessible Name in the APG preview (Chapter 5), which in itself is superfluous, because everything about it can be found in https://www.w3.org/TR/accname-1.1/ . But Chapter 5 of APG explains very well for developers the things that are hard to understand in https://www.w3.org/TR/accname-1.1/. 
I would also like to see something like this for Live Regions. Especially since I think that the current ARIA specification contains bugs and ambiguities (https://github.com/w3c/aria/issues/1048, https://github.com/w3c/aria-practices/issues/78#issuecomment-524619306)

JAWS-test:
@zcorpan:

For many of my test cases at https://github.com/FreedomScientific/VFO-standards-support/issues?q=is%3Aopen+is%3Aissue+author%3AJAWS-test+live+region I can't tell myself what a correct output of the screenreader should be like. I can only report an error that the output is not consistent. But which browser makes it right? As long as this cannot be clearly derived from the ARIA specification, there is a need for improvement in the ARIA specification on the one hand and for further explanations in the APG on the other.

zcorpan:
OK, I think this section could do more to explain the issues with live regions and go into more detail. It sounds like this topic could use some teleconference time or TPAC time to discuss what the issues are that APG should cover.

jnurthen:
If there is something specific to dicuss and the relevant parties are going to be there I am happy to put it on the TPAC agenda. To do this please add the F2FCandidate label to the relevant issue AND provide the requested details in https://github.com/w3c/aria/issues/1001

zcorpan:
I will be there.

zcorpan:
@jnurthen I don't recall if we discussed this at TPAC. :) Should we discuss this on next week's telecon?

zcorpan:
Add a link to second layout grid example

zcorpan:
@mcking65 wanted a more thorough list of cases that trigger live regions. Currently that section mostly links to the accessibility tree section, which has new content in https://github.com/w3c/aria-practices/pull/1045 -- but still doesn't enumerate all possible cases.

I haven't had bandwidth to address this. But I think it's also not clear from the relevant specifications how CSS affects the accessibility tree, so it's difficult to state anything with confidence in APG.

zcorpan:
I have now converted this to HTML.

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+      However, different kinds of live regions have different expected behavior;
+      users might expect an error message to disrupt what they are doing, since the error is important,
+      but want to wait with reading new chat messages until they are done typing their own message.
+    </p>
+
+    <section id="live_region_states_and_properties">
+      <h3>Live Region States and Properties</h3>
+
+      <p>
+        ARIA has the following attributes mark up live regions.
```
```suggestion
        ARIA has the following attributes to mark up live regions.
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+
+          <li><code>off</code>: assistive technologies will not provide this information unless this region has focus.</li>
+        </ul>
+
+        <p>
+          With the exception of a few roles, the default value of <code>aria-live</code> is <code>off</code>.
+        </p>
+
+        <p>
+          For example, when the user types into a search field, the page could update the page with search results as the user is typing.
```
```suggestion
          For example, when the user types into a search field, the page could update with search results as the user is typing.
```

redundant "the page"

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          To inform users of assistive technology that this has happened, a polite live region can be used.
```
```suggestion
          To inform assistive technology users that this has happened, a polite live region can be used.
```

Remove potential confusion over "To inform users _of something that happened_..."

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          When the user is done typing, and the search is complete, the user is informed of how many results were found,
+          without having to move focus away from the search field.
+        </p>
+
+        <pre><code>&lt;form role="search" aria-labelledby="search">
```
carmacleod: Aside: What tool is being used to convert from markdown to markup? It seems to have a bug where it outputs `&lt;` instead of `<` inside `<pre><code>` examples, which makes github changes harder to read. If possible, please open an issue against the conversion tool. Thanks! :)

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+ &lt;h2 id="search">Search&lt;/h2>
+ &lt;label>Search query: &lt;input type="search" name="q" oninput="updateSearch(event)">&lt;/label>
+ &lt;div id="search-result-status" role="region" aria-live="polite">&lt;/div>
```
carmacleod: [`region` is a landmark role](https://w3c.github.io/aria/#region).

Need to discuss with group whether a live region can use `role="region"` without being a landmark.

If so, then the ARIA spec needs to allow this use specifically for live regions.

Or, if the live region in this example is truly intended to be a landmark, then in needs an accessible name (or else conformance tools are going to flag it as missing one).

If, however, the group decides that `role="region"` should not be allowed on live regions without turning them into landmarks, and if it is decided that the live region in this example should not be a landmark (really, I think the role here should be "status", but we haven't described that yet), then we need to know what role, if any, should go on this example's live region div. (No role, i.e. "generic"? Or "group", perhaps? Or should we just use "status" role before it is discussed?)

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+}
+&lt;/script>
+</code></pre>
+      </section>
+
+      <section id="indicate_what_content_to_announce_with_aria-atomic">
+        <h4>Indicate What Content to Announce with <code>aria-atomic</code></h4>
+
+        <p>
+          The <code>aria-atomic</code> attribute takes the values "true" and "false". The attribute can also be omitted.
```
```suggestion
          The <code>aria-atomic</code> attribute takes the values "true" and "false". The attribute can also be omitted, as discussed below.
```

To keep the first-time reader from thinking that discussion of the default value must be missing.

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          When this attribute is omitted, the user agent will use the closest ancestor that has <code>aria-atomic</code> set to <code>true</code> or <code>false</code>,
+          or if there is no such ancestor, <code>false</code>.
+        </p>
+
+        <p>
+          For example, consider a clock that can be set to notify the user of the current time at a regular interval.
+          Without <code>aria-atomic</code>, the assistive technology might only notify the changed components, rather than the full time.
+        </p>
+
+        <pre><code>&lt;div id="clock" role="region" aria-live="polite" aria-atomic="true">
```
carmacleod: See previous discussion about `role="region"`.

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+&lt;/div>
+</code></pre>
+      </section>
+
+      <section id="indicate_which_content_changes_are_relevant_with_aria-relevant">
+        <h4>Indicate Which Content Changes are Relevant with <code>aria-relevant</code></h4>
+
+        <p>
+          The <code>aria-relevant</code> attribute can be used to inform assistive technologies about which kinds of changes are relevant to inform users about.
+          It takes a list of keywords, with the following meanings:
```
```suggestion
          It takes a space-separated list of keywords, with the following meanings:
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+        </p>
+
+        <ul>
+          <li><code>additions</code>: Element nodes are added to the accessibility tree within the live region.</li>
```
```suggestion
          <li><code>additions</code>: When element nodes are added to the accessibility tree within the live region.</li>
```

Consider adding "When" to these (short for "Relevant when...").

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          <li><code>text</code>: Text content or a text alternative is added to any descendant in the accessibility tree of the live region.</li>
```
```suggestion
          <li><code>text</code>: When text content or a text alternative is added to any descendant in the accessibility tree of the live region.</li>
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          <li><code>removals</code>: Text content, a text alternative, or an element node within the live region is removed from the accessibility tree.</li>
```
```suggestion
          <li><code>removals</code>: When text content, a text alternative, or an element node within the live region is removed from the accessibility tree.</li>
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          Instead of announcing the removal of something, it is often better to add new content that tells the user what happened.
+          For example, "Alice is now offline" is clearer than "Alice" or "Removed, Alice".
+          Therefore, avoid using <code>removals</code> or <code>all</code>.
+        </p>
+
+        <p>
+          For example, a list of online contacts:
+        </p>
+
+        <pre><code>&lt;div role="region" aria-live="polite" aria-labelledby="contacts">
```
carmacleod: See note above regarding `role="region"`. Note that in this particular example, since the `region` has a name ("Contacts") it will definitely be a landmark. For this particular example, this is probably what the author wants - it makes sense to have the contacts list be a landmark, therefore easily navigable. I'm just flagging this to include it in the discussion.

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+        </p>
+      </section>
+
+      <section id="triggering_live_regions">
+        <h4>Triggering Live Regions</h4>
+
+        <p>
+          Additions and removals in the accessibility tree can happen due to changes to the DOM tree or changes to the applied CSS.
+          For example, changing the CSS <code>display</code> property to <code>none</code> causes the element to be removed from the accessibility tree.
+          See the <a href="#accessibility-tree">Accessibility tree</a> section for more details.
```
carmacleod: This section needs a bit more discussion. Does setting `display: block;` on the live region trigger it? Or does it have to be rendered first, and then have a change? Does changing CSS display on live region descendants trigger it?

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+        <li><code>marquee</code></li>
+      </ul>
+
+      <section id="live_region_role_alert">
+        <h4>Live Region Role <code>alert</code></h4>
+
+        <p>
+          The <code>alert</code> role indicates important, usually time-sensitive, information.
+          Use this role when focus is not moved to the message, and the user is not expected to close the message.
+          For an alert dialog that can be closed, the the <code>alertdialog</code> role instead.
```
```suggestion
          For an alert dialog that can be closed, use the <code>alertdialog</code> role instead.
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+          The default value for <code>aria-live</code> is <code>polite</code>.
+        </p>
+
+        <p>
+          For example, a chat log would be an appropriate use case for the <code>log</code> role.
+        </p>
+
+        <pre><code>&lt;h1 id="irc-log">IRC Log&lt;/h1>
+&lt;div role="log" aria-labelledby="irc-log">
+ &lt;p>[10:26] &amp;lt;Charl> ok let&#39;s test it and see if it works&lt;/p>
```
```suggestion
 &lt;p>[10:26] <Charl> ok let&#39;s test it and see if it works&lt;/p>
```

This instance of `&lt;` was changed to `&amp;lt;` by the markdown-to-markup conversion, and that ends up being surfaced in the document. Important to have the converter fixed, or maybe just start with markup instead, to make changes easier to read. For example, this code contains `let&#39;s` instead of `let's` and `&lt;` in front of every html tag.

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+ &lt;h2 id="search">Search&lt;/h2>
+ &lt;label>Search query: &lt;input type="search" name="q" oninput="updateSearch(event)">&lt;/label>
+ &lt;div id="search-result-status" role="status">&lt;/div>
+&lt;/form>
+</code></pre>
+
+        <p>
+          The HTML <code>output</code> element has the <code>status</code> role by default.
+          The <code>output</code> element is defined to represent the result of a calculation performed by the application, or the result of a user action.
+          The result of a user search is thus appropriate use of the <code>output</code> element:
```
```suggestion
          The result of a user search is thus an appropriate use of the <code>output</code> element:
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+        <h4>Live Region Role <code>timer</code> and <code>marquee</code></h4>
+
+        <p>
+          Usage of the <code>timer</code> and <code>marquee</code> roles is discouraged.
+        </p>
+
+        <p>
+          The <code>timer</code> role is a numerical counter which indicates an amount of elapsed time from a start point, or the time remaining until an end point.
+          The <code>timer</code> role is a subclass of the <code>status</code> role.
+          The default value of <code>aria-live</code> value of <code>off</code>.
```
```suggestion
          The default value of <code>aria-live</code> is <code>off</code>.
```

Redundant "value of".

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+        </p>
+
+        <p>
+          The <code>marquee</code> role indicates non-essential information that changes frequently.
+          The default value of <code>aria-live</code> value of <code>off</code>.
```
carmacleod: Remove redundant "value of".

```suggestion
          The default value of <code>aria-live</code> is <code>off</code>.
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+
+    <p>
+      For assistive technologies that honor the <code>aria-busy</code> attribute, one possible implementation strategy is to not announce any content for elements when <code>aria-busy</code> is set to <code>true</code>.
+      Therefore, it is important that the attribute is changed back to <code>false</code> when the content ought to be available to the user.
+      For example, when the update is complete, but also when there is an error, or if the update is aborted.
+      Otherwise, content could be left inaccessible to assistive technology users.
+    </p>
+
+    <p>
+      Consider the search form from the earlier example in the <a href="#live-regions">Live Resions</a> section.
```
```suggestion
      Consider the search form from the earlier example in the <a href="#live-regions">Live Regions</a> section.
```

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+      When the user starts typing a new search into the search field, the script would update the search results live region,
+      and maybe update multiple times as new search results appear,
+      but it would be a better experience for users of assistive technology to only be notified when the new search is complete.
+    </p>
+
+    <pre><code>&lt;form role="search" aria-labelledby="search">
+ &lt;h2 id="search">Search&lt;/h2>
+ &lt;label>Search query: &lt;input type="search" name="q" oninput="updateSearch(event)">&lt;/label>
+ &lt;div id="search-result-status" role="region" aria-live="polite">&lt;/div>
```
carmacleod: See earlier discussion about `role="region"`.
For this example, consider simply using `role="status"` (in which case, `aria-live="polite"` can be removed).

aria-practices.html
scottaohara: if it's not named then it shouldn't have a `role=region`.  does this example even need a role?

JAWS-test:
I am very much in favor of to specify clearly when a live region must be output by AT. Then, in the next step, a ticket can be opened at the manufacturer of the AT, if this does not work as specified. 

I don't think it's useful to list what the current state of live region output is, because that doesn't solve the problem of not being clearly specified. Once the specification clearly states what must result in the output of a live region, the problems in the AT can be fixed.

accdc:
To clarify some of the areas where live regions are important in practice and expound on what I said in the meeting, I have some examples here that are all valid.

Combobox, where the first suggested item is automatically announced using a dynamic live region.
http://whatsock.com/tsg/Coding%20Arena/ARIA%20Comboboxes/ARIA%20Comboboxes%20(Native%20Inputs,%20Editable%20with%20Substring%20Match)/demo.htm

To identify the dynamic results of drag and drop widgets when actions are performed.
http://whatsock.com/tsg/Coding%20Arena/Drag%20and%20Drop/demo.htm

Dynamic error handling for inline errors.
http://whatsock.com/tsg/Coding%20Arena/Inline%20Form%20Field%20Validation%20and%20Dynamic%20Help%20Tooltips/Inline%20Form%20Field%20Validation%20(Simple)/demo.htm

Dynamic tooltips when focus remains on the same input to guide user responses.
http://whatsock.com/tsg/Coding%20Arena/Inline%20Form%20Field%20Validation%20and%20Dynamic%20Help%20Tooltips/Dynamic%20Help%20Tooltip/demo.htm

Dynamic chat implementations for new message feedback during user interaction.
http://whatsock.com/tsg/Coding%20Arena/Web%20Chat%20and%20Dynamic%20Message%20Announcement/Web%20Chat%20(Static)/demo.htm

A simple checkbox control to identify a shopping cart experience.
http://whatsock.com/tsg/Coding%20Arena/Web%20Chat%20and%20Dynamic%20Message%20Announcement/Dynamic%20Message%20Announcement/demo.htm

As I said during the call, live regions must never be relied upon to ensure accessibility, but using live regions can significantly enhance accessibility for accessible implementations.

JAWS-test:
> As I said during the call, live regions must never be relied upon to ensure accessibility, but using live regions can significantly enhance accessibility for accessible implementations.

That may be the current situation, but that must change. Live region must function reliably. One reason for this is e.g. the WCAG, [SC 4.1.3](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html)

aria-practices.html
carmacleod: > does this example even need a role?

Nope. :) Decision on the call was to delete `role=region` and leave it with no role.
(The "Contacts list" example later on can keep region and label because it's ok as a true landmark).

I didn't want to state "No role" as the "best" choice without floating a bunch of other possibilities out there first - even the wacky possibility of changing the spec to allow live region elements to have a "non-landmark role=region". 😄

accdc:
"That may be the current situation, but that must change. Live region must function reliably. One reason for this is e.g. the WCAG, SC 4.1.3"

I agree totally that live regions must function reliably and as expected.

However, it's important to note that having live regions alone cannot guarantee accessibility. E.G. Within JAWS, you can disable the announcement of llive regions in the Verbosity dialog (Insert+V), which will make your guaranteed accessible app totally inaccessible if it relies solely on live regions.

sinabahram:
I’m a bit confused by this justification of aria-live being able to be turned off. In application contexts with custom controls that don’t easily map to existing native concepts, aria-live is a primary modality for conveying information to a screen-reader user. Is there some sort of stateful buffer being proposed or some other mechanism that can act as a replacement? Otherwise, what is the alternative that cannot be turned off, which can also guarantee accessibility?

accdc:
Hi Sina, I'm not justifying anything, and unfortunately there is no alternative in many cases for a user to receive this information, in which case  it must be announced reliably.

My point is though, that the screen reader provider allows for this to be turned off, and we as developers have no control over this. If a user does so, then they will not be able to access the functionality unless it is accessible on the page in some other fashion, such as through basic navigation commands.

E.G. Providing invisible error alerts that are announced is not a sufficient mechanism for displaying an error if it cannot also be arrowed to within the page during standard navigation.

JAWS-test:
> However, it's important to note that having live regions alone cannot guarantee accessibility. E.G. Within JAWS, you can disable the announcement of llive regions in the Verbosity dialog (Insert+V), which will make your guaranteed accessible app totally inaccessible if it relies solely on live regions

I can disable the output of a lot of content, ARIA roles, etc., in JAWS. However, this is the responsibility of the user and is not a problem of live regions. By reliable output, I mean output in the default setting of the respective AT, or in a setting with high verbosity. As soon as this works reliably, I am satisfied.

sinabahram:
I completely agree with this:

“E.G. Providing invisible error alerts that are announced is not a sufficient mechanism for displaying an error if it cannot also be arrowed to within the page during standard navigation.”

 

But, I also agree that many things can be turned off by the user, and that is on them once done, so that always having an equivalency for aria-live must not be required; however, in the case of errors, that makes total sense.

StommePoes:
> However, it's important to note that having live regions alone cannot guarantee accessibility.

I'm always looking to see if something else can work too-- not because a screen reader user can turn off announcements (I agree that's on the user), but for screen mag and Braille-only it's still an issue.

In theory, screen mag could have bugs filed against them to add live regions for non-local error messages.

jnurthen:
> In theory, screen mag could have bugs filed against them to add live regions for non-local error messages.

Like this https://github.com/FreedomScientific/VFO-standards-support/issues/15 ?

sinabahram:
Wouldn’t concerns around screen magnifiers and related AT need to be addressed in presumably visual ways, whether they are stateful or not? The modality of consumption for liv regions is screenreader delivered audio or Braille, right?

StommePoes:
> Wouldn’t concerns around screen magnifiers and related AT need to be addressed in presumably visual ways, whether they are stateful or not?

Could be, wouldn't have to be. Right now I get control names read out but not states in ZT and it's helpful for those with issues reading the names visually. Other magnifiers don't do any audio and I'm not thinking those ones should add it for live regions... I'm just thinking of the ones who already do some audio.

In Real Life, developers are generally running on the assumption "live regions are for people who don't see [some obvious visual change OR alert text]" but they don't tend to realise the effects of limited viewports. If screen mag could read out live regions the way some read out control names, more users would get that benefit.

>The modality of consumption for liv regions is screenreader delivered audio or Braille, right?

Is someone delivering Braille? Last I heard (a long while back), there were good reasons not to, but that left the issue unsolved for deaf-blind users. The SR portion of course gets the live region, but if you can't hear your screen reader, you're not getting what is sometimes essential information.

It's something I'm always thinking about and keep wondering if anything in standards or guidelines could help. Maybe even just good docs reminding developers and designers about limited viewports would help?

sinabahram:
These comments around interplay between aria-live and screen magnification are super helpful, so thank you for that.

 

I agree that aria-live needs to be figured out for Braille. It’s not fair, especially to those who are deaf-blind, that this channel of information is not presented, and I can think of so many trivial ways screen readers could make this available, not the least of which is a simple keystroke that let’s you review the last 10 alerts, with a one character notifier of their existence, or a flash message if the user so chooses. It would be a small engineering ask and a monumental usability win, but sadly, I don’t run screen reader dev teams.

 

And, given how you use screen magnification, this idea that aria-live may be needing to be spoken even though other things are not being spoken also makes a lot of sense.

 

That answers my query, so thank you.

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+            <td><code>aria-atomic</code></td>
+            <td>Indicates what content to announce.</td>
+          </tr>
+          <tr>
+            <td><code>aria-relevant</code></td>
+            <td>Indicates which content changes are relevant.</td>
+          </tr>
+          <tr>
+            <td><code>aria-busy</code></td>
+            <td>Defers content updates. Not specific to live regions. See the <a href="#aria-busy">next section</a>.</td>
```
```suggestion
            <td>Defers announcement of content updates. Not specific to live regions. See <a href="#defer_exposing_content_updates_using_aria-busy">Defer Exposing Content Updates using aria-busy</a>.</td>
```

smhigley:
From the call last week, here are four common situations where I frequently come across use cases for live regions:

1. Form field errors, particularly inline errors
	Ironically, despite this being so common, it is also one of the flakiest examples. Since errors are often displayed on blur, the live region announcement conflicts with the user's focus change announcements, with unreliable results.
2. Single page app navigation or content change announcements
	I actually prefer moving focus for most artificial page changes, but in cases where that is not desired an assertive live region is a common fallback.
3. App-wide push notifications
	Any notification that is unrelated to the current activity of the user. Examples could include an incoming chat message or a notice of a background action (e.g. a deployment initiated earlier) succeeding or failing. For apps where this is common, I usually recommend having a notifications center that shows the number of unread notifications and where past notifications can be reviewed (a bit like github's own notifications button). That helps avoid any potential keyboard access and timeout issues, as well as aiding discovery by magnification users.
4. Component-specific content updates
	Examples include the number of results returned when filtering a grid or executing a site-wide search, or  content returned asynchronously.

My team and I also have a general live region guideline based on anecdotal feedback and preferences from screen reader users that live regions should not be used for changes that are immediate and predictable, or part of the expected functionality of a control. So, for example, we would not recommend using a live region for announcing results in a combobox dropdown, or the result of a drag & drop; instead we would recommend making the control's name, role, states, and description adequate for the user to understand what is going on.

StommePoes:
Sina:
> a simple keystroke that let’s you review the last 10 alerts,

This reminds me of a comment made I think by Derek about NVDA on Twitter. There's no queue (and this is deliberate) although I'm unfamiliar with any possible queueing in the refreshable braille part of things.

Which relates to Sarah's point:
> Form field errors, particularly inline errors
Ironically, despite this being so common, it is also one of the flakiest examples. Since errors are often displayed on blur, the live region announcement conflicts with the user's focus change announcements, with unreliable results.

React developer Flarnie ran into exactly this and "solved" it with a manual setTimeout() :(

(thread: https://twitter.com/ProvablyFlarnie/status/1166141795652141059  ...unfortunately some replies by one NVDA developer appear to be gone. Specifically, another NVDA developer (Derek) replied about queues in this sub-thread https://twitter.com/stommepoes/status/1166247305433026560)

I like Sarah's point #3 and have heard of people being specifically unhappy that the place where they expect to see these records do not find them there.

Regarding point #4 when offering remediation in audits, groups I've worked with have had trouble with things like large changes happening within a large table (rows being added or removed) in-place using JavaScript... we're certain nobody wants to suddenly start hearing all that new content, but instead a dedicated, short message only saying there were changes. Or even a system-bell type thing (for some apps which are used regularly by users, not one-off websites users have no chance or inclination to spend time learning).

I have seen developers put `aria-live=polite` on large containers of things like filter results, expecting that this Does What They Mean. :(

aria-practices.html
StommePoes: This helps!

smhigley:
@StommePoes good clarification on point 4 😄. I definitely would recommend a live region announcement along the lines of "N results found", not literally the content of every search result/filtered row 😅.

Regarding using a timeout to solve focus/live region conflict -- I've tried this, and it didn't work reliably for me unless the timeout was longer than the full focus announcement (which would be impossible to determine ahead of time). Sounds like it's time for even more live region testing 😁

carmacleod:
@smhigley 
> using a timeout to solve focus/live region conflict ... didn't work reliably

The following note is being added to the example in the [aria-errormessage spec in PR #1157](https://pr-preview.s3.amazonaws.com/w3c/aria/pull/1157.html#aria-errormessage). Does this seem helpful?

> NOTE
> aria-live="assertive" is used in this example in favor of aria-live="polite" to ensure the error message is read before the user is done interacting with the text input. Otherwise, the user may confuse the message with that of the next text input.

sinabahram:
Mal, thanks for the clarifications, but I can’t think of a single technical reason why queueing can’t happen for previous messages, Braille or otherwise. It just sounds like dogma sans any evidence, instead of a fact-based reason for not doing something. I’ve heard many such claims over the decades, and pretty much all of them have come down to some combination of not wanting to do or prioritize something, rather than it being hard or impossible. Hey, I could be completely wrong, though. At any rate, that’s probably quite off topic here, so happy to discuss elsewhere if interested.

 

As far as usage of aria-live, we generally advise folks to simply announce that there are N errors, set focus (only if appropriate) to the errors list which should be displayed on screen or the first component with said error, where the error is clearly associated with the field through some change that ends up modifying accessible name, and then to follow other best practices around invalid and required states. Obviously, moving someone’s focus is treated as sacred, and only done with the utmost care. Also, following the principle of semantic prioritization, an announcement of “has error” is appropriate, IMHO, to come first, before the original label announcement, but if-only-if this is tested well, to ensure it doesn’t happen on a blank form, for example. Original invalid due to empty is not error 😊.

 

I very much hope nobody is ok with large swaths of text being slammed into an aria-live region, polite or assertive, as that’s completely and utterly useless based on basic first principles (no examples or code needed, just cognitive science around working memory and lack of navigability). If we are trading horror stories really quick. Mikrotik routers used to have a web interface with aria-live=assertive around the bandwidth block, so anytime there was a single kb/s of change on the network, which of course is literally every single second, it would update, making the entire interface completely unusable, until of course I removed it by injecting JavaScript into the address bar (so fun!).

JAWS-test:
> The following note is being added to the aria-errormessage example in ARIA PR #1157. Does this seem helpful?

[At least with JAWS, the value in aria-live seems to have no effect.](https://github.com/FreedomScientific/VFO-standards-support/issues/284) Therefore, it would be useful to name the affected field in the blur event error message so that the AT user knows what the error message refers to

aria-practices.html
```diff
@@ -3734,6 +3734,390 @@ <h5>Examples</h5>
...
+        <p>
+          When this attribute is set to <code>true</code>, assistive technologies will render the element as a whole, and not just parts that have been changed.
+        </p>
+
+        <p>
+          When this attribute is set to <code>false</code>, assistive technologies will only render the changes (as per <code>aria-relevant</code>) to the user.
+        </p>
+
+        <p>
+          When this attribute is omitted, the user agent will use the closest ancestor that has <code>aria-atomic</code> set to <code>true</code> or <code>false</code>,
```
```suggestion
          When this attribute is omitted, the user agent will use the value of <code>aria-atomic</code> from the closest ancestor that has <code>aria-atomic</code> set to <code>true</code> or <code>false</code>,
```

mcking65:
Replaced by #2989

PR #1037, opened 2019-06-05
Fix keyCode.RETURN in examples/radio/radio-2

Seasonley:
it has `case this.keyCode.RETURN` in RadioGroup handler, but not defined


examples/radio/radio-2/js/radioGroupActiveDescendant.js
```diff
@@ -28,6 +28,7 @@ var RadioGroup = function (domNode) {
 
   this.keyCode = Object.freeze({
     'TAB': 9,
+    'RETURN': 13,
```
mcking65: Instead of adding here, I believe we need to remove the case for return from line 128 (now 129 in this file).

The table of supported keys does not include return or enter. The interaction pattern does not include enter or return. In this way, radios are similar to checkbox; enter does nothing. This should align with what native HTML radios do in most browsers.

jongund: @mcking65 I can fix it

mcking65:
@Seasonley, thank you for finding and raising this issue. 

As I mentioned before, we'd like to take a different approach because "Return" is not supposed to toggle a radio.

@jongund also found some other issues to resolve and has opened #1064. So, I am closing this PR.

If you are available to help with other bugs, please let me know. It is usually best to comment in an issue first if it is a known bug, or if it is a new bug, raise an issue first and tell us if you are able to submit a PR.

Issue #258, opened 2017-01-26
Draft guidance on aria-hidden


mcking65:
Provide both reasons for caution and exemplary use cases.

carmacleod:
@jongund's table in https://github.com/w3c/aria/issues/1256#issuecomment-623503204 is a great starting point for this.

PR #1045, opened 2019-06-20
Introduce sections for Accessibility tree and aria-hidden

zcorpan:
Fixes https://github.com/w3c/aria-practices/issues/258

### DIFF
--- a/aria-hidden.md
+++ b/aria-hidden.md
@@ -1,0 +1,44 @@
+== 7. Accessibility tree
+
+Browsers use a so-called accessibility tree to communicate the representation of web content with assistive technologies. See the [Core Accessibility API Mappings](https://w3c.github.io/core-aam/#dfn-accessibility-tree) specification for further information about accessibility trees.
+
+The accessibility tree is based on information in the DOM tree and information in CSS. For example, text in the DOM tree can also be in the accessibility tree. CSS generated content is not part of the DOM tree, but can be part of the accessibility tree.
+
+For example, given the HTML document
+
+```
+<!doctype html>
+<html>
+ <head>
+  <title>Cookie policy</title>
+ </head>
+ <body>
+  <p>Cookies are delicious delicacies. <button>OK</button></p>
+ </body>
+</html>
+```
+
+The DOM tree includes element nodes closely resembling the markup, with a root `html` element that has two element children `head` and `body`, and so on.
+
+The accessibility tree does not include the `head` or `title` elements, because the `head` element is hidden by default. The paragraph is included in the accessibility tree, with a text node child "Cookies are delicious delicacies. " and a `button` node, whose accessible name is "OK".
+
+CSS and ARIA can be used to control what is included in the accessibility tree, which is discussed in the following sections.
+
+== 8. Excluding subtrees from the accessibility tree
+
+An element that is hidden, either by default or by the author using CSS (in particular `display: none` and `visibility: hidden`), or using the HTML `hidden` attribute (which by default maps to CSS `display: none`), is not included in the accessibility tree. Its descendants are also not included. The element is hidden from all users, both visually and from assistive technologies.
+
+For example, a [disclosure](#disclosure) shows and hides content, and that content is intended to be hidden for all users, and could use CSS `display: none` or the HTML `hidden` attribute.
+
+Using `aria-hidden` with the value `true` excludes the element and its descendants from the accessibility tree, just like hidden elements, but such elements are *not* hidden visually. The descendants are excluded from the accessibility tree even if an element in the exluded subtree sets `aria-hidden` to `false`.
+
+`aria-hidden` can be used to hide decorative or redundant content from screen readers, for example an inline SVG icon.
+
+```
+<a href="/">
+ <svg aria-hidden="true">...</svg>
+ Home
+</a>
+```
+
+Warning. Only use `aria-hidden="true"` when hiding content from only assistive technologies improves the experience for users of assistive technologies.


aria-hidden.md
```diff
@@ -0,0 +1,44 @@
...
+
+CSS and ARIA can be used to control what is included in the accessibility tree, which is discussed in the following sections.
+
+## 8. Excluding subtrees from the accessibility tree
+
+An element that is hidden, either by default or by the author using CSS (in particular `display: none` and `visibility: hidden`), or using the HTML `hidden` attribute (which by default maps to CSS `display: none`), is not included in the accessibility tree. Its descendants are also not included. The element is hidden from all users, both visually and from assistive technologies.
+
+For example, a [disclosure](#disclosure) shows and hides content, and that content is intended to be hidden for all users, and could use CSS `display: none` or the HTML `hidden` attribute.
+
+Using `aria-hidden` with the value `true` excludes the element and its descendants from the accessibility tree, just like hidden elements, but such elements are *not* hidden visually. The descendants are excluded from the accessibility tree even if an element in the exluded subtree sets `aria-hidden` to `false`.
```
a11ydoer: typo in the last line, "exluded" to "excluded"

aria-hidden.md
```diff
@@ -0,0 +1,44 @@
...
+`aria-hidden` can be used to hide decorative or redundant content from screen readers, for example an inline SVG icon.
+
+```
+<a href="/">
+ <svg aria-hidden="true">...</svg>
+ Home
+</a>
+```
+
+Warning. Only use `aria-hidden="true"` when hiding content from only assistive technologies improves the experience for users of assistive technologies.
```
a11ydoer: How about "**the user experience** of assistive technologies" than "the experience for users of technologies"?

zcorpan:
Thank you for the review @a11ydoer! I've addressed the comments and moved the text into `aria-practices.html`, and rebased on latest master. If this looks OK, I believe this is ready to merge.

aria-practices.html
```diff
@@ -6279,6 +6279,38 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
 
   </section>
 
+  <section id="accessibility_tree">
+    <h2>Accessibility tree</h2>
+    <p>Browsers use a so-called accessibility tree to communicate the representation of web content with assistive technologies. See the <a href="https://w3c.github.io/core-aam/#dfn-accessibility-tree">Core Accessibility API Mappings</a> specification for further information about accessibility trees.</p>
```
carmacleod: I would prefer not to use the  phrase "so-called" because it typically means "inappropriately named".
Maybe something like:
> Browsers construct an accessibility tree to communicate the representation of web content with assistive technologies.

aria-practices.html
```diff
@@ -6279,6 +6279,38 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+    <p>The accessibility tree is based on information in the DOM tree and information in CSS. For example, text in the DOM tree can also be in the accessibility tree. CSS generated content is not part of the DOM tree, but can be part of the accessibility tree.</p>
```
carmacleod: > The accessibility tree is based on information in the DOM tree and information in CSS.

Should we also say, " and ARIA"? As in:
> The accessibility tree is based on information in the DOM tree and information in CSS and ARIA.

aria-practices.html
```diff
@@ -6279,6 +6279,38 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+    <p>For example, a <a href="#disclosure">disclosure</a> shows and hides content, and that content is intended to be hidden for all users, and could use CSS <code>display: none</code> or the HTML <code>hidden</code> attribute.</p>
+    <p>Using <code>aria-hidden</code> with the value <code>true</code> excludes the element and its descendants from the accessibility tree, just like hidden elements, but such elements are <em>not</em> hidden visually. The descendants are excluded from the accessibility tree even if an element in the excluded subtree sets <code>aria-hidden</code> to <code>false</code>.</p>
+    <p><code>aria-hidden</code> can be used to hide decorative or redundant content from screen readers, for example an inline SVG icon.</p>
+    <pre><code>&lt;a href=&quot;/&quot;&gt;
+ &lt;svg aria-hidden=&quot;true&quot;&gt;...&lt;/svg&gt;
```
carmacleod: @mcking65 Do we want to add `focusable="false"` to this SVG example code to remind authors that if they use `aria-hidden="true"` on an SVG, they will want to force the SVG to not take focus in IE11?

aria-practices.html
jnurthen: ARIA is in the DOM tree isn't it?

StommePoes: Will it be later when authors can directly send it via JavaScript (AOM)?
I'm not clear on whether AOM-changes will be reflected in the DOM or not.

carmacleod: @jnurthen 
> ARIA is in the DOM tree isn't it?

Good point.  :)

@StommePoes
> Will it be later when authors can directly send it via JavaScript (AOM)?
> I'm not clear on whether AOM-changes will be reflected in the DOM or not.

I'm not sure whether all of the upcoming AOM-changes will be reflected in the DOM (@alice?), however I recently tested some simple attribute reflection in pull https://github.com/w3c/aria-practices/pull/1161 and I can confirm that ARIA attributes set using dot notation are definitely reflected in the DOM. (yay!)

alice: Yay indeed!

For **[element reflection](https://github.com/WICG/aom/blob/gh-pages/explainer.md#reflecting-element-references)**, we will set the content attribute to the empty string if it's not possible to set it to anything meaningful (e.g. if you pass in an element with no ID, etc.) 

So you will have a "hint" in the DOM that you need to check via the property:

Say you had:
```html
<input id="input"></input>
<optgroup id="group">
  <option>Option 1</option>
  <option>Option 2</option>
</optgroup>
```

Then you did:

```js
input.ariaActiveDescendantElement = group.firstElementChild;
```

Then you would have

```html
<input id="input" aria-activedescendant></input>
<!-- ... -->
```

- but retrieving `input.ariaActiveDescendantElement` would return the first `<option>`.

For [**Custom Element default semantics**](https://github.com/WICG/aom/blob/gh-pages/explainer.md#default-semantics-via-customelementsdefine), it's a little trickier - just like for native elements, there won't be a straightforward way to retrieve the "internal" properties.

If a Custom Element's implementation does this:

```js
this._internals = customElements.createInternals(this);
this._internals.role = "tab";
this._internals.ariaSelected = false;
this._internals.ariaControlsElement = tabPanel;
```

then a script in the page context has no reliable way to know about that, since it is opaque to the page (again, like native elements).

For **[Virtual Accessibility Nodes](https://github.com/WICG/aom/blob/gh-pages/explainer.md#virtual-accessibility-nodes)**, absolutely nothing has been decided for sure, but it seems likely that they will not be DOM nodes and will only be available to script.

StommePoes: Oof. Off-topic I know, but recently auditing and checking that aria-current was being (re) set correctly as user moved thru a nav on an SPA. Could only see that in the DOM, did not show up in any browser's tree, and SRs were contradicting each other :( Hope extra tools come with AOM changes.

alice: i.e. extra testing/debugging tools to programmatically probe the accessibility tree?

jnurthen: (off-topic reply) FF seems to expose current in the a11y tree correctly as far as I can see. It would be nice if chrome exposed all ARIA attributes in the "ARIA Attributes" section of the Accessibility pane (I suspect it is simply an oversight that aria-current is not there and a bug needs to be logged)

aria-practices.html
zcorpan: Thanks, I didn't know "so-called" had that meaning.

aria-practices.html
zcorpan: That seems like a bug in IE11. If I understand correctly, APG takes the stance to generally not work around bugs in implementation. I'm not sure if that is only implementation *of ARIA* or in general, though?

zcorpan:
Thanks for the review! I've addressed comments, except I didn't add `focusable="false"` https://github.com/w3c/aria-practices/pull/1045#discussion_r322718328

JAWS-test:
> That seems like a bug in IE11.

There is a bug of IE 11 and older versions of Edge. I think APG are not responsible for browser bugs. But where a bug is well known and there is a simple solution to avoid it, it would make sense to offer this solution, especially if a bug has such serious consequences as an SVG that is hidden and gets the focus.

aria-practices.html
carmacleod: Completely agree that APG should not work around browser bugs, however since IE11 is never going to change, perhaps this bug needs to be worked around permanently? If you do decide to add `focusable="false"`, it deserves a decent comment in the code to explain why it's there.

zcorpan:
Thanks @JAWS-test and @carmacleod. I think it is a reasonable balance to include and document simple workarounds for unmaintained and widely used implementations, while in general assume conforming implementations.

zcorpan:
@a11ydoer or @carmacleod, can either of you review this? Thanks!

aria-practices.html
```diff
@@ -6279,6 +6279,39 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
 
   </section>
 
+  <section id="accessibility_tree">
+    <h2>Accessibility tree</h2>
+    <p>Browsers construct an accessibility tree to communicate the representation of web content with assistive technologies. See the <a href="https://w3c.github.io/core-aam/#dfn-accessibility-tree">Core Accessibility API Mappings</a> specification for further information about accessibility trees.</p>
+    <p>The accessibility tree is based on information in the DOM tree (including ARIA) and information in CSS. For example, text in the DOM tree can also be in the accessibility tree. CSS generated content is not part of the DOM tree, but can be part of the accessibility tree.</p>
+    <p>For example, given the HTML document</p>
```
```suggestion
    <p>For example, given the HTML document:</p>
```

colon after document, to introduce example code

aria-practices.html
```diff
@@ -6279,6 +6279,39 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+&lt;html&gt;
+ &lt;head&gt;
+  &lt;title&gt;Cookie policy&lt;/title&gt;
+ &lt;/head&gt;
+ &lt;body&gt;
+  &lt;p&gt;Cookies are delicious delicacies. &lt;button&gt;OK&lt;/button&gt;&lt;/p&gt;
+ &lt;/body&gt;
+&lt;/html&gt;</code></pre>
+    <p>The DOM tree includes element nodes closely resembling the markup, with a root <code>html</code> element that has two element children <code>head</code> and <code>body</code>, and so on.</p>
+    <p>The accessibility tree does not include the <code>head</code> or <code>title</code> elements, because the <code>head</code> element is hidden by default. The paragraph is included in the accessibility tree, with a text child node &quot;Cookies are delicious delicacies. &quot; and a child <code>button</code> element node, whose accessible name is &quot;OK&quot;.</p>
```
```suggestion
    <p>The accessibility tree does not include the <code>head</code> or <code>title</code> elements, because the <code>head</code> element is hidden by default. The paragraph is included in the accessibility tree, with a text child node &quot;Cookies are delicious delicacies.&quot; and a child <code>button</code> element node, whose accessible name is &quot;OK&quot;.</p>
```

delete extra space

aria-practices.html
```diff
@@ -6279,6 +6279,39 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+    <p>For example, a <a href="#disclosure">disclosure</a> shows and hides content, and that content is intended to be hidden for all users, and could use CSS <code>display: none</code> or the HTML <code>hidden</code> attribute.</p>
+    <p>Using <code>aria-hidden</code> with the value <code>true</code> excludes the element and its descendants from the accessibility tree, just like hidden elements, but such elements are <em>not</em> hidden visually. The descendants are excluded from the accessibility tree even if an element in the excluded subtree sets <code>aria-hidden</code> to <code>false</code>.</p>
+    <p><code>aria-hidden</code> can be used to hide decorative or redundant content from screen readers, for example an inline SVG icon.</p>
+    <pre><code>&lt;a href=&quot;/&quot;&gt;
+ &lt;svg aria-hidden=&quot;true&quot; focusable="false"&gt;...&lt;/svg&gt;
+ Home
+&lt;/a&gt;</code></pre>
+    <p class="note">The <code>focusable="false"</code> attribute is included to work around a bug in some versions of Internet Explorer, where <code>svg</code> elements are focusable by default.</p>
```
```suggestion
    <p class="note">The <code>focusable="false"</code> attribute is included to work around a bug in Internet Explorer, where <code>svg</code> elements are focusable by default.</p>
```

Remove "some versions of" so that authors are not guessing whether or not the bug is in IE11... it is.

aria-practices.html
```diff
@@ -6279,6 +6279,39 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+    <p class="warning">Only use <code>aria-hidden=&quot;true&quot;</code> when hiding content from only assistive technologies improves the user experience of assistive technologies.</p>
```
```suggestion
    <p class="warning">Only use <code>aria-hidden=&quot;true&quot;</code> to hide content from assistive technologies if it improves the experience for assistive technology users.</p>
```

Sentence felt awkward, partly because AT don't have experiences - AT users do.

carmacleod:
Consider making section 9. [Excluding subtrees from the accessibility tree](https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1045.html#excluding_subtrees_from_the_accessibility_tree)
and section 10. [Intentionally Hiding Semantics with the presentation Role](https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1045.html#presentation_role)
be subsections of section 8. [Accessibility tree](https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1045.html#accessibility_tree).

carmacleod:
Consider adding something to the effect of "don't put aria-hidden on interactive content".  to the warning box at the end of section 9. Excluding subtrees from the accessibility tree.
(This may be off-topic for this PR - if so, I can open a new issue).

zcorpan:
> Consider making section 9. Excluding subtrees from the accessibility tree
and section 10. Intentionally Hiding Semantics with the presentation Role
be subsections of section 8. Accessibility tree.

Done.

zcorpan:
> Consider adding something to the effect of "don't put aria-hidden on interactive content". to the warning box at the end of section 9. Excluding subtrees from the accessibility tree.

Done -- I added it to the same warning.

zcorpan:
@carmacleod can you review the latest changes? Thanks!

PR #1070, opened 2019-07-07
Add disclosure menu example for issue 1028

### LINKED ISSUE #1028: Add site navigation example that uses disclosure buttons that show lists of links

smhigley:
Resolves #1028.

Continuation of PR #1036, which contains a long thread of relevant discussion.

#### Preview Link for WIP

[View WIP on disclosure navigation menu example in the issue1028-disclosure-menu-example branch](https://raw.githack.com/w3c/aria-practices/issue1028-disclosure-menu-example/examples/disclosure/disclosure-navigation.html)

#### Relevant comments from mcking65 in issue 1028

> @smhigley, I am done with editorial revisions to the issue1028-disclosure-menu-example feature branch.

> You may want to review the test code. I added documentation for aria-current, which currently does not have a test. I also expanded the documentation for arrow keys, home, and end. Now, there are multiple rows with the same data-test-id, which is OK. However, it might be good to sanity check to make sure the testing covers all the documented behaviors.

> My revisions to the accessibility features section were the most significant. They are worth a close read. I attempted to adjust wording and rationale to be understandable by a broader audience. I'm curious if you agree whether I achieved that objective.

> I am wondering if the checkbox for optional navigation keys should be unchecked by default. Given there are both side effects and extra code driven by the optional behaviors, shouldn't they be off by default?

Edited to add preview link: https://raw.githack.com/w3c/aria-practices/issue1028-disclosure-menu-example/examples/disclosure/disclosure-navigation.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1070.html" title="Last updated on Jul 10, 2019, 7:05 PM UTC (79893a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1070/c4f4d9f...79893a9.html" title="Last updated on Jul 10, 2019, 7:05 PM UTC (79893a9)">Diff</a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "lint:html": "npm run vnu-jar",
     "regression": "ava test/tests",
     "regression-report": "node test/util/report",
-    "test": "npm run lint && npm run vnu-jar && npm run regression",
+    "test": "npm run lint && npm run regression",
     "vnu-jar": "java -jar node_modules/vnu-jar/build/dist/vnu.jar --errors-only --filterfile .vnurc --no-langdetect --skip-non-html *.html examples/"
   },
   "repository": {
--- a/test/tests/disclosure-navigation.js
+++ b/test/tests/disclosure-navigation.js
@@ -12,6 +12,7 @@ const exampleFile = 'disclosure/disclosure-navigation.html';
 const ex = {
   buttonSelector: '#exTest button',
   menuSelector: '#exTest > li > ul',
+  linkSelector: '#exTest > li > ul a',
   buttonSelectors: [
     '#exTest > li:nth-child(1) button',
     '#exTest > li:nth-child(2) button',
@@ -28,13 +29,13 @@ const ex = {
 
 ariaTest('"aria-controls" attribute on button', exampleFile, 'button-aria-controls', async (t) => {
   t.plan(1);
-  await assertAriaControls(t, ex.buttonSelectors[0]);
+  await assertAriaControls(t, ex.buttonSelector);
 });
 
 ariaTest('"aria-expanded" attribute on button', exampleFile, 'button-aria-expanded', async (t) => {
   t.plan(7);
 
-  await assertAttributeValues(t, ex.buttonSelectors[0], 'aria-expanded', 'false');
+  await assertAttributeValues(t, ex.buttonSelector, 'aria-expanded', 'false');
 
   let buttons = await t.context.session.findElements(By.css(ex.buttonSelector));
   let menus = await t.context.session.findElements(By.css(ex.menuSelector));
@@ -46,6 +47,40 @@ ariaTest('"aria-expanded" attribute on button', exampleFile, 'button-aria-expand
     );
     await assertAttributeValues(t, ex.buttonSelectors[i], 'aria-expanded', 'true');
   }
+});
+
+ariaTest('"aria-current" attribute on links', exampleFile, 'link-aria-current', async (t) => {
+  t.plan(36);
+
+  const buttons = await t.context.session.findElements(By.css(ex.buttonSelector));
+  const menus = await t.context.session.findElements(By.css(ex.menuSelector));
+
+  for (let b = 0; b < buttons.length; b++) {
+    const links = await menus[b].findElements(By.css('a'));
+
+    for (let l = 0; l < links.length; l++) {
+
+      await buttons[b].click();
+      await links[l].click();
+
+      t.is(
+        await links[l].getAttribute('aria-current'),
+        'page',
+        'after clicking link at index ' + l + ' on menu ' + b + 'aria-current should be set to page'
+      );
+
+      let ariaCurrentLinks = await t.context.session.findElements(
+        By.css(`${ex.linkSelector}[aria-current="page"]`)
+      );
+
+      t.is(
+        ariaCurrentLinks.length,
+        1,
+        'after clicking link at index ' + l + ' on menu ' + b + ', only one link should have aria-current set'
+      );
+    }
+  }
+
 });
 
 // Keys
--- a/test/tests/disclosure-navigation.js
+++ b/test/tests/disclosure-navigation.js
@@ -1,0 +1,229 @@
+'use strict';
+
+const { ariaTest } = require('..');
+const { By, Key } = require('selenium-webdriver');
+const assertAriaControls = require('../util/assertAriaControls');
+const assertAttributeValues = require('../util/assertAttributeValues');
+const assertTabOrder = require('../util/assertTabOrder');
+const assertHasFocus = require('../util/assertHasFocus');
+
+const exampleFile = 'disclosure/disclosure-navigation.html';
+
+const ex = {
+  buttonSelector: '#exTest button',
+  menuSelector: '#exTest > li > ul',
+  buttonSelectors: [
+    '#exTest > li:nth-child(1) button',
+    '#exTest > li:nth-child(2) button',
+    '#exTest > li:nth-child(3) button'
+  ],
+  menuSelectors: [
+    '#exTest > li:nth-child(1) ul',
+    '#exTest > li:nth-child(2) ul',
+    '#exTest > li:nth-child(3) ul'
+  ]
+};
+
+// Attributes
+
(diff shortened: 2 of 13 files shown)


smhigley:
@mcking65 regarding tests:
I didn't add `aria-current` tests originally since they are not part of the core disclosure menu javascript, and would be expected to be part of the static html of a page in a normal example (i.e. one without contrived page content). I'll add some for this anyway, so that functionality doesn't regress on this repo.

About the `package.json` changes: I noticed that the `vnu-jar` script was being run twice in succession inside the `test` script, so I fixed it.

I'll add a couple non-substantive changes to some formatting and typos before Tuesday. I also have some qualms about specifically calling out certain screen reader behaviors -- I think it might be better to keep it general (e.g."screen readers will usually intercept arrow key presses within this example markup") rather than get into specifics that don't apply across all screen readers. I'd also prefer to leave the sentence mentioning screen reader behavior out of the following list item, since I think it could read as if tabbing is only implemented for screen reader users:
> If implemented, the optional navigation keys supplement, but do not replace, tabbing among buttons and links.

Thanks again for the thorough editorial review and additions!

mcking65:
> @mcking65 regarding tests:
> I didn't add `aria-current` tests originally since they are not part of the core disclosure menu javascript, and would be expected to be part of the static html of a page in a normal example (i.e. one without contrived page content). I'll add some for this anyway, so that functionality doesn't regress on this repo.
> 
> About the `package.json` changes: I noticed that the `vnu-jar` script was being run twice in succession inside the `test` script, so I fixed it.
> 
> I'll add a couple non-substantive changes to some formatting and typos before Tuesday. I also have some qualms about specifically calling out certain screen reader behaviors -- I think it might be better to keep it general (e.g."screen readers will usually intercept arrow key presses within this example markup") rather than get into specifics that don't apply across all screen readers. I'd also prefer to leave the sentence mentioning screen reader behavior out of the following list item, since I think it could read as if tabbing is only implemented for screen reader users:
> 
> > If implemented, the optional navigation keys supplement, but do not replace, tabbing among buttons and links.
> 
> Thanks again for the thorough editorial review and additions!

@smhigley wrote:
> I didn't add `aria-current` tests originally since they are not part of the core disclosure menu javascript, and would be expected to be part of the static html of a page in a normal example (i.e. one without contrived page content). I'll add some for this anyway, so that functionality doesn't regress on this repo.

Thank you! Our testing is to ensure the reliability of our examples and their documentation, so we still have to test the stuff that is "contrived" for illustrative purposes.

That said, I don't think this pattern is limited to navigation among static pages. It could be used in a single-page app type of architecture.

@smhigley wrote:
>I also have some qualms about specifically calling out certain screen reader behaviors -- I think it might be better to keep it general (e.g."screen readers will usually intercept arrow key presses within this example markup") rather than get into specifics that don't apply across all screen readers.

We have consistent descriptions of widget-related mode switching throughout the examples. True, some screen readers like Mac-VO and Narrator are manual switching while JAWS and NVDA are automatic. But, this widget-related behavior is still relevant for all screen readers that run on desktop systems where keyboards are a primary input method. One thing that is still in our backlog, though, is an appendix section that describes the mode switching features in more detail. When that is done, we will link to it from notes like this.

I do appreciate your concern though. Rather than be more general though, I think it may be better to refine it further.

@smhigley wrote:
>I'd also prefer to leave the sentence mentioning screen reader behavior out of the following list item, since I think it could read as if tabbing is only implemented for screen reader users:
> 
> > If implemented, the optional navigation keys supplement, but do not replace, tabbing among buttons and links.

We could explain this without mentioning screen reader users as part of the justification. However, I think the point begs for explanation since it is contrary to every other pattern that provides arrow, home, and end key support. I'll propose another wording.

test/tests/disclosure-navigation.js
```diff
@@ -0,0 +1,229 @@
...
+    '#exTest > li:nth-child(2) ul',
+    '#exTest > li:nth-child(3) ul'
+  ]
+};
+
+// Attributes
+
+ariaTest('"aria-controls" attribute on button', exampleFile, 'button-aria-controls', async (t) => {
+  t.plan(1);
+  await assertAriaControls(t, ex.buttonSelectors[0]);
```
spectranaut: if you pass `ex.buttonSelector` instead of `ex.buttonSelectors[0]`, then `assertAriaControls` will test whether `aria-controls` is set on all menu button elements. Right now, it only tests whether `aria-controls` is on the first one.

test/tests/disclosure-navigation.js
```diff
@@ -0,0 +1,229 @@
...
+});
+
+ariaTest('"aria-expanded" attribute on button', exampleFile, 'button-aria-expanded', async (t) => {
+  t.plan(7);
+
+  await assertAttributeValues(t, ex.buttonSelectors[0], 'aria-expanded', 'false');
```
spectranaut: same here, you can pass `ex.buttonSelector`

test/tests/disclosure-navigation.js
```diff
@@ -0,0 +1,229 @@
...
+    '#exTest > li:nth-child(3) button'
+  ],
+  menuSelectors: [
+    '#exTest > li:nth-child(1) ul',
+    '#exTest > li:nth-child(2) ul',
+    '#exTest > li:nth-child(3) ul'
+  ]
+};
+
+// Attributes
```
spectranaut: Missing test for attribute "aria-current"

spectranaut:
@mcking65 I fixed and added tests! Should be good to go :)

package.json
```diff
@@ -15,7 +15,7 @@
     "lint:html": "npm run vnu-jar",
     "regression": "ava test/tests",
     "regression-report": "node test/util/report",
-    "test": "npm run lint && npm run vnu-jar && npm run regression",
```
spectranaut: hmmm `npm run lint` includes `npm run vnu-jar` so this is redundant, I believe!

spectranaut: oh wow somehow I read this as the addition. The double html linting is removed in this PR :)

test/tests/disclosure-navigation.js
```diff
@@ -0,0 +1,264 @@
...
+ariaTest('"aria-current" attribute on links', exampleFile, 'link-aria-current', async (t) => {
+  t.plan(36);
+
+  const buttons = await t.context.session.findElements(By.css(ex.buttonSelector));
+  const menus = await t.context.session.findElements(By.css(ex.menuSelector));
+
+  for (let b = 0; b < buttons.length; b++) {
+    const links = await menus[b].findElements(By.css('a'));
+
+    for (let l = 0; l < links.length; l++) {
```
smhigley: I don't want to block merging, just have an optional suggestion: the javascript that controls `aria-current` will function the same regardless of which link is clicked, so it's only necessary to test on one link click and cut down on test time a little.

Otherwise, looks good! Thanks for writing this! (Also, changing the loops is entirely up to you, I don't want to hold up merging the PR)

smhigley:
Thanks @spectranaut and @mcking65! It looks like the Travis build failed on a timeout in an unrelated test, so I restarted it. If it passes, let's merge!

PR #1073, opened 2019-07-11
Issue1072 datepicker dialog aria selected patch

jongund:
Patch to fix the aria-selected bug in the date picker dialog example.

[Preview link of patch](https://raw.githack.com/w3c/aria-practices/issue1072-datepicker-dialog-aria-selected-patch/examples/dialog-modal/datepicker-dialog.html)


jongund:
@spectranaut 

Valerie,
When I run the regression test for `datepicker-dialog` on this branch it reports 14 tests passes, but then it seems to get stuck and doesn't complete the testing.   
Could you take a look at this issue?

mcking65:
@jongund, We have conflicts, and I want to be very careful that we don't lose any of the changes that James and I made in the examples/dialog-modal/datepicker-dialog.html file. I touched a couple hundred lines or so of the HTML in my recent commits. Those changes are now all in master, and they are not in this branch.

For this patch, are any changes to the HTML necessary? I'm guessing very few or none. So, there are 2 fast and risk-free paths to resolving the conflicts:

1) Copy the entire HTML file from master and use it to replace the one in this branch. Then, if you need to make HTML changes for the patch, make them in the freshly updated HTML file.
2) Make a new branch from master and recreate your patch in that branch, and open a fresh PR from that branch.

I like option 2 better because then the commit history in this branch will be clean and easy to understand. Right now, it contains many commits from the previous feature branch that were all squashed together when I merged them into master. So, they look new (like they are not in master) even though all their changes are already in master. If we start with a fresh branch made from master, the commits that show up in the PR will be only commits related to the patch.

We will be able to get this in. I am not going to tag the publication branch commit for the CFC until tomorrow.

jongund:
@spectrnaut @mcking65

I Am on vacation right now so if Valerie can make fixes that would be great, otherwise I can work on them Monday

spectranaut:
@jongund have a good vacation! :)

@mcking65 I'll make a new PR with the aria-select changes.

spectranaut:
Opened new PR to remove conflicts: https://github.com/w3c/aria-practices/pull/1074

PR #1084, opened 2019-07-13
added missing roles to reference table script'

jongund:
@mcking65 

I added the following missing roles to the `reference-table.js` script:

* alert
* alertdialog
* dialog
* log
* marquee
* status
*timer


scripts/reference-tables.js
```diff
@@ -70,13 +75,15 @@ const ariaRoles = [
   'separator',
   'slider',
   'spinbutton',
+  'satus',
```
nschonni:
```suggestion
  'status',
```

jongund: @nschonni Thanks just fixed it

PR #1085, opened 2019-07-13
fix: Add missing aria roles

nschonni:
Compared with MDN list


jongund:
@nschonni 

Landmark roles are generated by a different method that the other aria roles, so we don't want them in the list.

nschonni:
How does look with it commented out with a reason?

mcking65:
Resolved by #1084.

nschonni:
No problem, I'll do a PR for just the purposely skipped `region` value in the array

PR #1095, opened 2019-07-15
Date picker dialog: fix bug when clicking on a disabled day

jongund:
@mcking65 

This patch fixes a bug when clicking on a disabled date in the date picker grid with a pointing device.
It now updates the grid with the new month, before it generated a javascript error due to reference to an undefined value.

### DIFF
--- a/examples/dialog-modal/js/datepicker-day.js
+++ b/examples/dialog-modal/js/datepicker-day.js
@@ -161,7 +161,7 @@ DatePickerDay.prototype.handleKeyDown = function (event) {
 DatePickerDay.prototype.handleMouseDown = function (event) {
 
   if (this.isDisabled()) {
-    this.datepicker.moveFocusToDay(this.date);
+    this.datepicker.moveFocusToDay(this.day);
   }
   else {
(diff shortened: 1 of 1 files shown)


spectranaut:
Hi @jongund -- I actually think that when you click on a disabled date, the expectation is to selected the date. It's a big weird when the calendar changes visually (after selecting the date) and you mouse is now over a new date where as the date you selected is highlighted and in a different place in the grid.

Maybe it would be better to just fix the console error and let people click on/select disabled dates!

jongund:
> Hi @jongund -- I actually think that when you click on a disabled date, the expectation is to selected the date. It's a big weird when the calendar changes visually (after selecting the date) and you mouse is now over a new date where as the date you selected is highlighted and in a different place in the grid.
> 
> Maybe it would be better to just fix the console error and let people click on/select disabled dates!

@spectranaut 
People in the working group when this was first implemented thought it should change to the "new" month, rather than select a date or do nothing (e.g. disabled).  So it would probably be better to keep the "new" month behavior, but I don't have a really strong opinion on the issue.

PR #1096, opened 2019-07-15
Finish tests for dialog datepicker and fix bugs

spectranaut:
This adds the rest of the tests for the dialog datepicker.

It also fixes two tiny bugs:
- The "data-date" is written as `2019-01-9` for all of the "ninth" days of the month. This was a simple off-by-one error.
- The keyboard shortcuts "PAGE UP", "PAGE DOWN", "SHIFT+PAGE UP" and "SHIFT+PAGE DOWN" didn't re-set the tab index after the calendar was update to the new month/year.

There is another bug that I'll make a followup issue for related to the SPACE button -- but I can only replicate it in webdriver not manually, so it's not urgent.

### DIFF
--- a/test/tests/datepicker-dialog.js
+++ b/test/tests/datepicker-dialog.js
@@ -29,6 +29,7 @@ const ex = {
   jan12019Button: '#example [role="dialog"] button[data-date="2019-01-01"]',
   jan22019Button: '#example [role="dialog"] button[data-date="2019-01-02"]',
   todayButton: `#example [role="dialog"] button[data-date="${todayDataDate}"]`,
+  currentlyFocusedButton: '#example [role="dialog"] button[tabindex=\'0\']',
   allFocusableElementsInDialog: [
     `#example [role="dialog"] button[data-date="${todayDataDate}"]`,
     '#example [role="dialog"] button[value="cancel"]',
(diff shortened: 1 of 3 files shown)


test/tests/datepicker-dialog.js
```diff
@@ -282,41 +287,372 @@ ariaTest('', exampleFile, 'dialog-shift-tab', async (t) => {
...
+    .findElement(By.css(ex.currentlyFocusedButton))
+    .getAttribute('data-date');
+
+  t.is(
+    dayInFocus,
+    day.toISOString().split('T')[0],
+    'After selected next month button, then next year button date, then previous month button, then previous year button, date should be ' + day.toISOString().split('T')[0] + ' but found: ' + dayInFocus
+  );
+
+  // TODO: Test SPACE
```
nschonni: Is it better to do `t.skip` instead of a comment? Or does that cause weirdness with the coverage scripts

nschonni: Oh, I see there actually is a todo https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#test-placeholders-todo

spectranaut: Hmm I dug into it more yesterday thinking it might have been a webdriver bug (because I couldn't replicate it) -- but then I realized it's a firefox only bug. I added a commit with the failing tests and made a issue: https://github.com/w3c/aria-practices/issues/1098

mcking65:
Darn, I wish I would have merged this before #1100 . 
@jongund, could you resolve the conflict?

Issue #1102, opened 2019-07-19
Open aria practice's examples in online javascript/html/css editor


spectranaut:
See draft PR #1110 for an example implementation using Codepen's [post API](https://blog.codepen.io/documentation/api/prefill/).

See draft PR #1111 for an example implementation using JSFiddle's [post API](https://docs.jsfiddle.net/api/display-a-fiddle-from-post)

We'd really like to hear from the community if they prefer JSFiddle or Codepen, it seems like both are perfectly workable solutions from the tech side.

mcking65:
@zcorpan, we are going with the codepen implementation in PR #1110.

However, instead of manually adding the button to open codepen to each example page, we would like to add it dynamically. The examples/js/examples.js script runs on every example page to add the HTML source of the example. We can add this capability to that script.

I'm assigning this issue to you to make that change to examples.js.

PR #1110, opened 2019-07-24
Infrastructure: Add support for opening examples in CodePen

spectranaut:
This PR is a proof of concept for issue #1102.

Codepen has a POST API to prefill the HTML/CSS/JS inputs.  This PR adds a dynamically created form that can send a POST request with the appropriate data to Codepen (look for "Open this example in Codepen" button on the `slider/multithumb-slider.html` page).

I tried to do this in a generalized way -- the code is mostly in `examples/js/examples.js` -- so that each example page would only need a small amount of html/js added, most of which could be added/copied directly from the template file. This solution should also work for examples pages that contain multiple examples.

Here is a PR for doing the same thing but using JSFiddle: #1111

#### Preview WIP

[View the slider example with the codepen button](https://raw.githack.com/w3c/aria-practices/issue1102-open-example-codepen/examples/slider/multithumb-slider.html)

### DIFF
--- a/examples/css/core.css
+++ b/examples/css/core.css
@@ -40,3 +40,72 @@ table.data.attributes tbody th,
 table.data.attributes tbody td {
   border: 1px solid silver;
 }
+
+/* CodePen button */
+.example-header {
+  display: flex;
+  align-items: center;
+  margin-top: 3rem;
+  page-break-after: avoid;
+  page-break-inside: avoid;
+  font: 100% sans-serif;
+  font-family: inherit;
+  line-height: 1.2;
+  hyphens: manual;
+}
+
+.example-header h2 {
+  margin: 0;
+}
+
+.example-header h2 + * {
+  margin-left: 1em;
+}
+
+form[action^="https://codepen.io/"] button {
+  display: inline-block;
+  position: relative;
+  padding: 0.4em 0.7em;
+  border: 1px solid hsl(213, 71%, 49%);
+  border-radius: 5px;
+  box-shadow: 0 1px 2px hsl(216, 27%, 55%);
+  color: #fff;
+  font-size: inherit;
+  text-shadow: 0 -1px 1px hsl(216, 27%, 25%);
+  background-color: hsl(216, 82%, 51%);
+  background-image: linear-gradient(to bottom, hsl(216, 82%, 53%), hsl(216, 82%, 47%));
+}
+
+form[action^="https://codepen.io/"] button:hover {
+  border-color: hsl(213, 71%, 29%);
+  background-color: hsl(216, 82%, 31%);
+  background-image: linear-gradient(to bottom, hsl(216, 82%, 33%), hsl(216, 82%, 27%));
+  cursor: default;
+}
+
+form[action^="https://codepen.io/"] button:focus {
+  outline: none;
+}
+
+form[action^="https://codepen.io/"] button:focus::before {
+  position: absolute;
+  z-index: -1;
+
+  /* button border width - outline width - offset */
+  top: calc(-1px - 3px - 3px);
+  right: calc(-1px - 3px - 3px);
+  bottom: calc(-1px - 3px - 3px);
+  left: calc(-1px - 3px - 3px);
+  border: 3px solid hsl(213, 71%, 49%);
+
+  /* button border radius + outline width + offset */
+  border-radius: calc(5px + 3px + 3px);
+  content: '';
+}
+
+form[action^="https://codepen.io/"] button:active {
+  border-color: hsl(213, 71%, 49%);
+  background-color: hsl(216, 82%, 31%);
+  background-image: linear-gradient(to bottom, hsl(216, 82%, 53%), hsl(216, 82%, 47%));
+  box-shadow: inset 0 3px 5px 1px hsl(216, 82%, 30%);
+}
--- a/examples/css/core.css
+++ b/examples/css/core.css
@@ -62,7 +62,7 @@ table.data.attributes tbody td {
   margin-left: 1em;
 }
 
-form[action^="https://codepen.io/"] button {
+.example-header button {
   display: inline-block;
   position: relative;
   padding: 0.4em 0.7em;
@@ -76,18 +76,18 @@ form[action^="https://codepen.io/"] button {
   background-image: linear-gradient(to bottom, hsl(216, 82%, 53%), hsl(216, 82%, 47%));
 }
 
-form[action^="https://codepen.io/"] button:hover {
+.example-header button:hover {
   border-color: hsl(213, 71%, 29%);
   background-color: hsl(216, 82%, 31%);
   background-image: linear-gradient(to bottom, hsl(216, 82%, 33%), hsl(216, 82%, 27%));
   cursor: default;
 }
 
-form[action^="https://codepen.io/"] button:focus {
+.example-header button:focus {
   outline: none;
 }
 
-form[action^="https://codepen.io/"] button:focus::before {
+.example-header button:focus::before {
   position: absolute;
   z-index: -1;
 
@@ -103,7 +103,7 @@ form[action^="https://codepen.io/"] button:focus::before {
   content: '';
 }
 
-form[action^="https://codepen.io/"] button:active {
+.example-header button:active {
   border-color: hsl(213, 71%, 49%);
   background-color: hsl(216, 82%, 31%);
   background-image: linear-gradient(to bottom, hsl(216, 82%, 53%), hsl(216, 82%, 47%));
--- a/examples/css/core.css
+++ b/examples/css/core.css
@@ -54,13 +54,11 @@ table.data.attributes tbody td {
   hyphens: manual;
 }
 
-.example-header h2,
-.example-header h3 {
+.example-header > :first-child {
   margin: 0;
 }
 
-.example-header h2 + *,
-.example-header h3 + * {
+.example-header > :first-child + * {
   margin-left: 1em;
 }
 
--- a/examples/css/core.css
+++ b/examples/css/core.css
@@ -54,11 +54,13 @@ table.data.attributes tbody td {
   hyphens: manual;
 }
 
-.example-header h2 {
+.example-header h2,
+.example-header h3 {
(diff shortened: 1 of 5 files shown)


mcking65:
This was discussed in the [September 10, 2019 meeting](https://www.w3.org/2019/09/10-aria-apg-minutes.html).

There was consensus that a separate section with its own heading is too heavy and that it would be better if the button were positioned to the right of the example heading.

I moved the button in the HTML, but I need someone to add the necessary styling to visually position it. I assume it would go in examples/css/core.css.

@jnurthen also suggested  a feature request to automatically run tidy on the HTML.

@jnurthen, are you making this request because the HTML in our files still isn't tidy in some cases? Or, is it that the formatting in our files is lost in transmission to CodePen?

jnurthen:
@mcking65 the HTML all seemed to be indented by a whole bunch of extra spaces. I want those removed.

mcking65:
@jnurthen commented:
> @mcking65 the HTML all seemed to be indented by a whole bunch of extra spaces. I want those removed.

Ah, thanks. Those are probably just the leading spaces that exist in our source. We should just trim the number of spaces from each line equal to the indent of the first line of the example.

ZoeBijl:
This additional white space is in our own source code. Basically, because the example code follows the indentation of the example page itself the example code has more white space. Similar to how this works for code in a `pre`-element.

Other than that this seems to work pretty well, thanks @spectranaut.

mcking65:
@ZoeBijl or @jnurthen, are one of you available to push a commit to this branch with a style for positioning the button? I think we want a class in examples/css/core.css that is applied to the "Open in CodePen" `input` element on the example page.

ZoeBijl:
@mcking65 if there’s an issue for that can you assign that to me?

ZoeBijl:
I have committed and pushed some core CSS changes that styles this button and positions it next to the “Example” heading text. Please review.

examples/css/core.css
```diff
@@ -76,18 +76,18 @@ form[action^="https://codepen.io/"] button {
...
+.example-header button:hover {
   border-color: hsl(213, 71%, 29%);
   background-color: hsl(216, 82%, 31%);
   background-image: linear-gradient(to bottom, hsl(216, 82%, 33%), hsl(216, 82%, 27%));
   cursor: default;
 }
 
-form[action^="https://codepen.io/"] button:focus {
+.example-header button:focus {
   outline: none;
```
jnurthen: For some reason I'm still seeing the browser outline on the button itself on Firefox.

ZoeBijl: I can solve this with a vendor prefixed property:

```
.example-header button::-moz-focus-inner {
  border: 0;
}
```

 But I feel like that would make this more convoluted than it needs to be. Thoughts?

StommePoes: Maybe we want developers to see that Firefox does this and maybe they'll be convinced to let it be.

The ones who just can't stand it in their own code go to StackOverflow to see hundreds of ways to kill focus outlines. Let them at least do that much leg-work.

carmacleod:
@ZoeBijl Please have a look at the grid page because we have longer headings

ZoeBijl:
The current CSS works with short and long headings. When you zoom in you’ll end up with something like this:
The title is displayed on the left over two lines with the button on the right, also displayed over two lines

---

@spectranaut have you tested this on a page with multiple examples?

spectranaut:
@ZoeBijl I remember designing the pattern with the possibility of multiple examples per page and I might have tested it locally when making this PR, but I didn't test it thoroughly. The function `sourceCode.add('sc1', 'ex1', 'editCode', 'cssJsFiles');` takes a bunch of ids to specify which example it's making the source code/moding the view in code pen button for. You can have two different ids for the two different buttons and two diferent sets of resource files.

spectranaut:
@jnurthen @mcking65 I added some regex to clean up the white space at the beginning of the html.

examples/slider/multithumb-slider.html
```diff
@@ -7,6 +7,7 @@
   <!--  Core js and css shared by all examples; do not modify when using this template. -->
   <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/base.css">
   <link rel="stylesheet" href="../css/core.css">
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
```
ZoeBijl: Is there a way we can do this without jQuery?

spectranaut: It would be a lot more obnoxious to get the javascript and css resources but certainly we could. Why should we not use jQuery?

ZoeBijl: Feels weird to include an entire library to open an example in an external resource 🤷‍♀. Is it a one time annoyance? Or is it annoying every time we add an example?

spectranaut: one time annoyance, to replace the `$.get` ajaxes in the `examples/js/examples.js` file.

spectranaut: Can't say that I won't to write that code, lol :) but I hear your point.

ZoeBijl: I can have a look

spectranaut:
@mcking65 -- I added a commit to add the button in the `example.js` file, and to test I also added a commit to the buttons to the [data grids example](https://raw.githack.com/w3c/aria-practices/issue1102-open-example-codepen/examples/grid/dataGrids.html) -- the buttons all work great!

However, you still have to edit the example page in three ways:
1. Add the a div with class "example-header" around the example's header to keep all the styling @ZoeBijl added.
2. Add an id to a element that contains all the js and css files you want to open in code pen (in the `Javascript and CSS Source Code` section).
3. Pass the id of the header and id of the container for the source files to the `sourceCode.add` call that already exists to put the `html` of each example on the example page.

AND, unfortunately, the data grids example does not work when opened in codepen. The code to initialize the data grid expects the grid to be in the APG example page. When we role this out, we are going to have to test each example -- it it works, then we can add the buttons (by doing the three steps above), if it doesn't work, then we will have to make a backlog issue to refactor the code so that the example can work in isolation.

mcking65:
@spectranaut, thank you. They are working beautifully!

Before merging, let's reduce this PR so that it is only adding the capability but not changing any example pages, i.e., changes only examples.js and core.css, reverting d96dd7d and changes to spinbutton.

>you still have to edit the example page in three ways:

We should do these mods in a series of separate PRs.

> 1. Add the a div with class "example-header" around the example's header to keep all the styling @ZoeBijl added.
> 2. Add an id to a element that contains all the js and css files you want to open in code pen (in the `Javascript and CSS Source Code` section).
> 3. Pass the id of the header and id of the container for the source files to the `sourceCode.add` call that already exists to put the `html` of each example on the example page.

Do we also have to load jquery on every page now as well?

It would be useful to have some way of tracking implementation progress and then later enforcing these requirements. What do you think of an "allowed to fail" lint rule that tests to see if these conditions might not be present?

> AND, unfortunately, the data grids example does not work when opened in codepen. The code to initialize the data grid expects the grid to be in the APG example page. When we role this out, we are going to have to test each example -- it it works, then we can add the buttons (by doing the three steps above), if it doesn't work, then we will have to make a backlog issue to refactor the code so that the example can work in isolation.

Bummer; I hope there aren't many such cases.

spectranaut:
Hi Matt, I just pushed a few commits:
1. Added more documentation to the `js/examples.js` API
2. Reverted the changes in slider and grid to add the button
3. Updated the example template to include the button by default (the example template turns out to be a good demo of the functionality, [check it out](https://raw.githack.com/w3c/aria-practices/issue1102-open-example-codepen/examples/coding-template/Example-Template.html)!)
4. Finally, your comment about jQuery reminded me about an earlier request from @ZoeBijl to remove jQuery, as it seems pretty silly to include it in every single example page for one cleaner function call in `js/examples.js`. So I removed the dependency on jQuery and wrote the ajax request in plain javascript.

Because of the last item, we should have one more code review on this PR, I think!

I also added a [how to page](https://github.com/w3c/aria-practices/wiki/How-to-add-the-%22Open-In-Codepen%22-button-to-an-example-page) in the wiki.

I didn't add anything to travis to check for whether the code pen button is on the page. Maybe we should keep a list of examples without CodePen links in the wiki page?

mcking65:
@spectranaut, this is looking great!! After merging, we can figure out how to track implementation in the example pages.

@smhigley can you please take a look at the new implementation? If it looks good to you, we'll merge.

examples/js/examples.js
```diff
@@ -300,4 +314,96 @@ function indentLines (input, indentation) {
...
+    var timerId = setInterval(() => {
+        console.log(totalFetchedFiles);
+        if (totalFetchedFiles === fileLinks.length) {
+            document.getElementById(jsonInputId).value = JSON.stringify(postJson);
+            clearInterval(timerId);
+        }
+    }, 500);
+
+    setTimeout(() => {
+        clearInterval(timerId);
```
smhigley, line 399: Would it make sense to have a `hideButton` call here too, with an error message about loading files timing out?

examples/js/examples.js
```diff
@@ -300,4 +314,96 @@ function indentLines (input, indentation) {
...
+      request.send();
+  }
+
+    var timerId = setInterval(() => {
+        console.log(totalFetchedFiles);
+        if (totalFetchedFiles === fileLinks.length) {
+            document.getElementById(jsonInputId).value = JSON.stringify(postJson);
+            clearInterval(timerId);
+        }
+    }, 500);
```
smhigley, line 396: Not a thing that needs to be fixed, but it looks like async/await is supported in all modern browsers. If we wanted to avoid using manual intervals, pushing all the `request.send` promises into an array then doing `await Promise.all` could work.

smhigley (approved):
It looks good to me!

The only thing I can think of is maybe it would be good to hide the "Open in Codepen" button until the requests have either completed or errored. I think we could just merge this now though and see how it goes, then maybe adjust in the future if desired :).

Same goes for the comments I left -- they were just thoughts I had when reading it through, not things that would need to be changed.

PR #1111, opened 2019-07-24
Proof of concept: Open example in JSFiddle

spectranaut:
This is a proof of concept for issue #1102.

JSFiddle has a POST API to prefill the HTML/CSS/JS inputs.  This PR adds a dynamically created form that can send a POST request with the appropriate data to JSFiddle (look for "Open this example in JSFiddle" button on the `slider/multithumb-slider.html` page).

I tried to do this in a generalized way -- the code is mostly in `examples/js/examples.js` -- so that each example page would only need a small amount of html/js added, most of which could be added/copied directly from the template file. This solution should also work for examples pages that contain multiple examples.

Here is a PR for doing the same thing but using Codepen: #1110

#### Preview WIP

[View the slider example with the JSFiddle button](https://raw.githack.com/w3c/aria-practices/issue1102-open-example-jsfiddle/examples/slider/multithumb-slider.html)

### DIFF
--- a/examples/js/examples.js
+++ b/examples/js/examples.js
@@ -33,6 +33,8 @@ var VOID_ELEMENTS = ['area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input',
 aria.widget.SourceCode = function () {
   this.location = new Array();
   this.code = new Array();
+  this.cssJsFiles = new Array();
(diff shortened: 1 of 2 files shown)


mcking65:
Closing; we are using the codepen approach in #1110

PR #1116, opened 2019-07-26
chore: Try Headless Firefox

nschonni:
Trying the config from https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-the-firefox-addon-in-headless-mode
Probably don't need "latest", but 60.0.2 is behind the latest ESR now, and is also about to bump https://www.mozilla.org/en-US/firefox/organizations/
Edit: Looks like this probably overlaps with https://github.com/w3c/aria-practices/blob/ecdb87a55dceed94ce0460bafc80d99c96757f3a/test/index.js#L11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: node_js
 node_js:
   - "node"
 
-addons:
-  firefox: "60.0.2"
-
 git:
   depth: 3
 
@@ -31,6 +28,8 @@ jobs:
 
     - stage: Test
       name: AVA Regression Tests
+      addons:
+        firefox: latest
       before_install:
       - |
         if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qE '(test/|examples/)'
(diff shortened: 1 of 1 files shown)


nschonni:
@spectranaut do you think there is value in adding jobs to test in Chrome and Safari?

spectranaut:
I do think there is value in testing Chrome and Safari -- we could catch browser bugs :) Browsers aren't perfect and we don't know what browser people will be looking at the example from.

nschonni:
Sounds good. I spun up the issue for now, but maybe i'll wait till your other improvements are in before sending a PR (if you don't beat me to it)

nschonni:
@spectranaut looks like it sped up the runs a little bit, but since it didn't run the regesion checks, maybe I should touch something so that is run

spectranaut:
Sorry this fell by the wayside for so long, @nschonni -- but you are right, the CI already runs headless filefox from the line you quoted. Just double checked I didn't invent the "CI" env variable: [travis docs](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables).

We should probably set firefox to latest. I'll just close this and open a new PR since I'm here :)

PR #1120, opened 2019-08-02
Add tabbed carousel example for issue 955

### LINKED ISSUE #955: Add carousel example with tabs as slide control
Add a carousel example where a tabbed interface is used to display slides and enable navigation to specific slides.

Related to issues #43 and #458.

jongund:
To resolve #955, adds an image carousel that includes a tablist control.

### Preview Link

[Preview example carousel in the compare branch of this pull request](https://raw.githack.com/w3c/aria-practices/issue955-add-tabbed-carousel-example/examples/carousel/carousel-2-tablist.html)

### Review checklist

* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @carmacleod 
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @spectranaut , @zcorpan

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 28, 2020, 6:15 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Faria-practices%2F2a9d5bdf3a03e234dfb33267a05ebf68887b9e70%2Faria-practices.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
[36mSpecification: https://rawcdn.githack.com/w3c/aria-practices/2a9d5bdf3a03e234dfb33267a05ebf68887b9e70/aria-practices.html?isPreview=true&publishDate=2020-04-28[39m
[36mReSpec version: 25.6.0[39m
[36mFile a bug: https://github.com/w3c/respec/[39m
[36mError: Error: Evaluation failed: Timeout: document.respecIsReady didn't resolve in 27213ms.[39m
[36m    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/ExecutionContext.js:122:13)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/ExecutionContext.js:48:12)[39m
[36m    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:128:12)[39m
[36m    at async fetchAndWrite (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:95:18)[39m
[36m    at async Object.generate [as respec] (/u/spec-generator/generators/respec.js:14:20)[39m
[36m    at async generate (/u/spec-generator/server.js:91:29)[39m
[36m  -- ASYNC --[39m
[36m    at ExecutionContext.<anonymous> (/u/spec-generator/node_modules/puppeteer/lib/helper.js:111:15)[39m
[36m    at DOMWorld.evaluate (/u/spec-generator/node_modules/puppeteer/lib/DOMWorld.js:112:20)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m  -- ASYNC --[39m
[36m    at Frame.<anonymous> (/u/spec-generator/node_modules/puppeteer/lib/helper.js:111:15)[39m
[36m    at Page.evaluate (/u/spec-generator/node_modules/puppeteer/lib/Page.js:860:43)[39m
[36m    at Page.<anonymous> (/u/spec-generator/node_modules/puppeteer/lib/helper.js:112:23)[39m
[36m    at generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:128:23)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m    at async fetchAndWrite (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:95:18)[39m
[36m    at async Object.generate [as respec] (/u/spec-generator/generators/respec.js:14:20)[39m
[36m    at async generate (/u/spec-generator/server.js:91:29)[39m
[36m[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/aria-practices%231120.)._
</details>


JAWS-test:
aria-roledescription is not supported by JAWS and leads to wrong output: https://github.com/FreedomScientific/VFO-standards-support/issues/288, https://github.com/FreedomScientific/VFO-standards-support/issues/154

ZoeBijl:
@JAWS-test the ARIA Authoring Practices is a document that shows how ARIA should _technically_ be used. It’s not a document that shows how to build accessible components that work with all browsers/assistive technologies today. Does that make sense?

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+                  <code>section</code>
+                </td>
+                <td>
+                  <ul>
+                    <li>Role <code>region</code> is implied for any <code>section</code> element that has an accessible name.</li>
+                    <li>Defines the carousel and its controls as a land mark region.</li>
+                  </ul>
+                </td>
+              </tr>
+              <tr data-test-id="carousel-region-aria-roledescription">
```
spectranaut: Test needed for this data-test-id

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+                  <code>section</code>
+                </td>
+                <td>
+                  <ul>
+                    <li>Informs assistive technologies to identify the element as a &quot;carousel&quot; rather than a &quot;region.&quot;</li>
+                    <li>Effects how the assistive technology renders the role but does not effect functionality, such as commands for navigating to landmark regions.</li>
+                  </ul>
+                </td>
+              </tr>
+              <tr data-test-id="carousel-region-aria-label">
```
spectranaut: Test needed for this data-test-id

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+                <td></td>
+                <th scope="row">
+                  <code>aria-label="Highlighted television shows"</code>
+                </th>
+                <td>
+                  <code>section</code>
+                </td>
+                <td>Provides a label that describes the content in the carousel region.</td>
+              </tr>
+              <tr data-test-id="carousel-aria-live">
```
spectranaut: test needed for this data-test-id

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+          <h3 id="kbd_label_1">Rotation Control Button</h3>
+          <table aria-labelledby="kbd_label_1 kbd_label" class="def">
+            <thead>
+              <tr>
+                <th>Key</th>
+                <th>Function</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr data-test-id="rotation-key-tab">
```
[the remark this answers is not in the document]
jongund: Added test

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+                <td>
+                  <ul>
+                    <li>Moves focus through interactive elements in the carousel.</li>
+                    <li>Rotation control and tablist buttons precede the slide content in the <kbd>Tab</kbd> sequence.</li>
+                  </ul>
+                </td>
+              </tr>
+              <tr data-test-id="rotation-enter-or-space-toggle">
+                <th><kbd>Enter</kbd> or <kbd>Space</kbd></th>
+                <td>Toggle the auto rotation of slides in the carousel.</td>
```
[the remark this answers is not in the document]
jongund: added test

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,364 @@
...
+const waitAndCheckAriaSelected = async function (t, index) {
+  return t.context.session.wait(async function () {
+    const tabs = await t.context.session.findElements(By.css(ex.tabSelector));
+    return (await tabs[index].getAttribute('aria-selected')) === 'true';
+  }, t.context.waitTime, 'Timeout waiting for aria-selected to be set to true.');
+};
+
+// Attributes
+
+ariaTest('Carousel 2: rotation button has aria-label that is updated based on pause state', exampleFile, 'rotation-aria-label', async (t) => {
```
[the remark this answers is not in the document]
jongund: @spectranaut removed "carousel-2" from regression messaging

examples/carousel/carousel-2-tablist.html
jongund: @spectranaut added test for data-test-id="carousel-region-aria-label"

examples/carousel/carousel-2-tablist.html
jongund: @spectranaut added test for data-test-id="carousel-aria-live"

examples/carousel/js/carouselTabpanel.js
```diff
@@ -0,0 +1,44 @@
...
+};
+
+CarouselTabpanel.prototype.hide = function () {
+  this.domNode.classList.remove('active');
+  this.domNode.removeAttribute('tabindex');
+};
+
+CarouselTabpanel.prototype.show = function () {
+  this.domNode.classList.add('active');
+  this.domNode.setAttribute('tabindex', '0');
```
smhigley: Do the tabpanels need to be focusable, since they all contain a focusable element? I'm not sure which is better in this case, it just struck me as perhaps confusing to have two tab stops that visually appear to be in the same place

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,364 @@
...
+  t.plan(4);
```
smhigley: `t.plan()` isn't necessary in these tests, since they are effectively synchronous with `async`/`await`, and the docs seem to [discourage it](https://github.com/avajs/ava/blob/master/docs/recipes/when-to-use-plan.md)

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,364 @@
...
+  await assertAriaLabelExists(t, ex.tablistSelector);
+});
+
+ariaTest('Carousel 2: role="tab" on button elements', exampleFile, 'tab-role', async (t) => {
+  t.plan(1);
+  await assertAriaRoles(t, 'ex1', 'tab', ex.tabCount, 'button');
+});
+
+ariaTest('"Carousel 2: aria-selected" set on role="tab"', exampleFile, 'tab-aria-selected', async (t) => {
+  t.plan(2 * ex.tabCount * ex.tabCount);
```
smhigley: Since the logic being tested here isn't unique to tab index, would it make sense to test before and after only one click, and then perhaps after changing with the keyboard as well?

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+    );
+  }
+});
+
+ariaTest('HOME key moves focus and selects tab', exampleFile, 'key-home', async (t) => {
+  t.plan(3 * ex.tabCount);
+
+  const tabs = await t.context.session.findElements(By.css(ex.tabSelector));
+  const tabpanels = await t.context.session.findElements(By.css(ex.tabpanelSelector));
+  for (let index = 0; index < tabs.length; index++) {
```
smhigley: The home key logic is independent of which tab has focus -- would it make sense to have two tests here instead, one with focus on the first tab, and one with focus on any other tab?

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+    );
+  }
+});
+
+ariaTest('END key moves focus and selects tab', exampleFile, 'key-end', async (t) => {
+  t.plan(3 * ex.tabCount);
+
+  const tabs = await t.context.session.findElements(By.css(ex.tabSelector));
+  const tabpanels = await t.context.session.findElements(By.css(ex.tabpanelSelector));
+  for (let index = 0; index < tabs.length; index++) {
```
smhigley: Same comment as above, but with focus on the last tab + focus on any other tab

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+ariaTest('TAB moves key through buttons', exampleFile, 'rotation-key-tab', async (t) => {
+  t.plan(1);
+
+  await assertTabOrder(t, ex.allFocusableItems);
+});
+
+ariaTest('ENTER pause and start carousel motion', exampleFile, 'rotation-enter-or-space-toggle', async (t) => {
+  t.plan(2);
+
+  let activeElement = await t.context.session.findElement(By.css(ex.activeCarouselItem)).getAttribute('aria-label');
```
smhigley: `ex.activeCarouselItem` doesn't appear to be defined anywhere in the test file

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+
+  await t.context.session.findElement(By.css(ex.pausePlayButtonSelector)).sendKeys(Key.ENTER);
+  // Move focus from widget
+  await t.context.session.findElement(By.css(ex.pausePlayButtonSelector)).sendKeys(Key.chord(Key.SHIFT, Key.TAB));
+
+  let compareWithNextElement = await t.context.session.wait(async function () {
+    let newActiveElement = await t.context.session.findElement(By.css(ex.activeCarouselItem)).getAttribute('aria-label');
+    return activeElement === newActiveElement;
```
smhigley: It might be worth it to manually set the `carouselTabList.timeInterval` manually to a very short interval, e.g. 100ms, then wait a specific amount of time greater than 100ms here and check that the active tab has changed. Theoreticallly, this implementation could wait until the carousel cycles all the way back to the first element, then return a false positive.

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+});
+
+ariaTest('ENTER pause and start carousel motion', exampleFile, 'rotation-enter-or-space-toggle', async (t) => {
+  t.plan(2);
+
+  let activeElement = await t.context.session.findElement(By.css(ex.activeCarouselItem)).getAttribute('aria-label');
+
+  await t.context.session.findElement(By.css(ex.pausePlayButtonSelector)).sendKeys(Key.ENTER);
+  // Move focus from widget
+  await t.context.session.findElement(By.css(ex.pausePlayButtonSelector)).sendKeys(Key.chord(Key.SHIFT, Key.TAB));
```
smhigley: Is there a reason to sent shift+tab to the play button? Focus on the button doesn't currently prevent the carousel from playing -- if that is the intended behavior, it could help to add a test without shift+tab to verify it

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+    let newActiveElement = await t.context.session.findElement(By.css(ex.activeCarouselItem)).getAttribute('aria-label');
+    return activeElement === newActiveElement;
+  }, t.context.WaitTime);
+
+  t.true(
+    compareWithNextElement,
+    'The active elements should stay the same when the pause button has been sent ENTER'
+  );
+
+  await t.context.session.findElement(By.css(ex.pausePlayButtonSelector)).sendKeys(Key.ENTER);
```
smhigley: It might also be worth it to verify that space and click work as well, perhaps with a reduced test that only checks the button classList, for instance

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
...
+
+// Keyboard interaction
+
+ariaTest('TAB moves key through buttons', exampleFile, 'rotation-key-tab', async (t) => {
+  t.plan(1);
+
+  await assertTabOrder(t, ex.allFocusableItems);
+});
+
+ariaTest('ENTER pause and start carousel motion', exampleFile, 'rotation-enter-or-space-toggle', async (t) => {
```
smhigley: There's some finicky logic around hover and focus and playing/pausing the carousel. Could there be tests added for that as well?

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,450 @@
+'use strict';
+
+const { ariaTest } = require('..');
+const { By, Key } = require('selenium-webdriver');
+const assertAttributeValues = require('../util/assertAttributeValues');
+const assertAriaControls = require('../util/assertAriaControls');
+const assertAriaLabelledby = require('../util/assertAriaLabelledby');
```
smhigley: `assertAriaLabelledby` is imported but not used

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+      }
+
+      document.write('<link rel="stylesheet" type="text/css" href="' + stylesheet + '">')
+
+      window.addEventListener('load', function () {
+        if (location.href.indexOf('moreaccessible') > 0) {
+          document.getElementById('id_link_more_accessible').setAttribute('aria-current', 'page');
+        }
+        else {
+          document.getElementById('id_link_less_accessible').setAttribute('aria-current', 'page');
```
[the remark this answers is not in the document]
smhigley: Ah, nvm, I see the different ids now :)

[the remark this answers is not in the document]
ZoeBijl: Sometimes more is less.

Can we change the IDs to be easier to differentiate?

test/tests/carousel_carousel-2-tablist.js
spectranaut: Good point, `t.plan()` is unnecessary in this case and many others -- but I'm not convinced it's bad to leave it in. It is reasonable to allow/encourage `t.plan()` everywhere because some tests need it, and it's an easy rule to express and follow (the rule: always use `t.plan()`?).

examples/carousel/js/carouselTab.js
```diff
@@ -0,0 +1,108 @@
...
+    event.preventDefault();
+  }
+};
+
+CarouselTab.prototype.handleClick = function () {
+  this.carousel.setSelectedTab(this, true);
+};
+
+CarouselTab.prototype.handleFocus = function () {
+  this.carousel.hasFocus = true;
```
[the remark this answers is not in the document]
spectranaut: This seems like the kind of thing we could consensus on and put into coding best practices. Issue #1180 is called "code readability", but could it be expanded to best practices/norms for the javascript in the examples? I'm fine with either of the latter two. @ZoeBijl and @jongund, do you have a preference between this options?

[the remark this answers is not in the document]
ZoeBijl: This is a perfect example of what should go into the Code Readability docs. In my opinion the second to last option is best. I often find myself wanting to split a component into more classes. But always keep it in one because I find that easier to read. Open to other opinions tho :).

examples/carousel/js/carouselTab.js
```diff
@@ -0,0 +1,108 @@
...
+*   @constructor CarouselTab
+*
+*
+*/
+var CarouselTab = function (domNode, carouselObj, tabpanelObj) {
+  this.domNode = domNode;
+  this.carousel = carouselObj;
+  this.tabpanel = tabpanelObj;
+
+  this.keyCode = Object.freeze({
```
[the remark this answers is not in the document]
ZoeBijl: Adding to my thumbs up: we should totes do this because IIRC `keyCode` has known internationalisation issues.

examples/carousel/css/carousel-2-tablist.css
```diff
@@ -0,0 +1,234 @@
+
+/* .carousel */
+
+.carouselTablist .carousel-inner {
```
ZoeBijl: Can we make the class names consistent?

`.carousel-tablist` instead of `.carouselTablist`

examples/carousel/js/carouselTab.js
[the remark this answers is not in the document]
smhigley: I'm for keeping them in one class, at least unless there's a sub-component that would truly be reusable across multiple componenst, e.g. a month spinbutton within a DatePicker dropdown. But I'd also be fine with that not being split out for the sake of cohesive examples, and because we're not building a production-style pattern library with shared code across components.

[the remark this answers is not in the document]
ZoeBijl: Perhaps we could combine it into something like: “All of a widgets’ logic should be contained in a single class. If there’s a good reason to split it into multiple classes communication between them should be done via events.” But like with better words.

Your comment does speak to why I prefer the single class strategy; we don’t need this to be production code—in fact, it’s better if it isn’t.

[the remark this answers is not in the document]
spectranaut: @jongund do you feel like you have enough guidance to do these changes? Thanks for your work so far and it's great we have your code to work out these norms with!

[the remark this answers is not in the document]
jongund: @spectranaut I have been traveling the last week, so I am trying to catch up with all the comments on this example, but they seem to make sense so far.

test/tests/carousel_carousel-2-tablist.js
```diff
@@ -0,0 +1,364 @@
...
+        'Tab at index ' + selectedEl + ' is selected, therefore, only the tabpanel at ' +
+          'index ' + selectedEl + ' should be displayed'
+      );
+    }
+  }
+});
+
+
+ariaTest('Carousel 2: "tabindex" on role="tab"', exampleFile, 'tab-tabindex', async (t) => {
+  t.plan(ex.tabCount * ex.tabCount);
```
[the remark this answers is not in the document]
spectranaut: Unlike above, the tab index is not explicitly checked in the keyboard navigation tests. They could be there. Would this satisfy you, @smhigley?

test/tests/carousel_carousel-2-tablist.js
spectranaut: We haven't ever written tests for functionality that is not specifically describe in the tables, so I don't think it's mandatory.

test/tests/carousel_carousel-2-tablist.js
smhigley: Anecdotally, this was confusing to me when I first starting looking at the existing tests in this repo, because it was so ubiquitous yet didn't seem to have an obvious reason. It also has the drawback that when modifying tests, it acts as an unnecessary gotcha -- unless you remember to update `t.plan()` when changing the number of tests run, you get a failure.

I don't think it would be too confusing to those writing tests to simply leave it out unless it's truly needed. It looks like the vast majority of our tests do not require it, since we're not doing anything fancy with interdependent async callbacks.

test/tests/carousel_carousel-2-tablist.js
smhigley: Thanks @spectranaut, that makes sense to me!

How do you feel about the other suggestion, to only tests before/after one or two clicks instead of all of them? Testing every tab seems unnecessary, since there's nothing in the code that would run differently for different tabs.

test/tests/carousel_carousel-2-tablist.js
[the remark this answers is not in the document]
smhigley: Yup, that sounds good!

test/tests/carousel_carousel-2-tablist.js
spectranaut: Right, it is annoying to update t.plan() when you change the number of assertions on purpose, but maybe not as annoying as false positives? The more common case we face in these tests are the for loops (and maybe some other branching) -- the test passes because the loop is over nothing due to do a change in the code that breaks the selector, for example.

test/tests/carousel_carousel-2-tablist.js
smhigley: This actually seems like a good place to ask about something I've wondered with the current testing setup: almost every interactive example we have has some functionality that doesn't fit neatly into any of the test ids, and yet still affects the user interaction and accessibility of the example.

So far, when I've been responsible for writing tests, I've written those tests under whichever test id seems closest (even if it's a bit of a stretch). However, we have no real way to catch their absence, and I'd love to have better guidance for how to handle test ids for these.

There's already an [issue related to this](https://github.com/w3c/aria-practices/issues/1205), but do you have any guidance in the meantime? I think this is a good case study, since it seems like there's a reasonable possibility that the hover and focus logic has some bugs that aren't being caught by the current tests.

test/tests/carousel_carousel-2-tablist.js
smhigley: Since the primary motivation to keep `t.plan()` is to catch empty selectors that then silently fail when iterating over a loop, how about this as a solution?: we create a helper to get elements that also has an assertion that the returned elements array is not empty.

It would have two benefits: we don't need to worry about people forgetting to add `t.plan()`, and we also aren't testing the number of elements that exist in the DOM, as long as there's at least one. So, as an example, if we end up updating the tabs example to have one more tab, we don't then need to go update a bunch of `t.plan()`s in the tests.

What do you think? I'd be happy to do the work for this if you like the approach.

smhigley: I’m also fine with this not being addressed in this PR

spectranaut: @smhigley this seems like a fine solution to me.

test/tests/carousel_carousel-2-tablist.js
spectranaut: I don't actually have guidance -- getting in the first set of tests (for all described behavior in the tables) was such a big pushed that I didn't consider behavior outside tables :) I appreciate you bringing up the issue now. Maybe we can make an issue to cover these scenarios for this example (to let this example page move forward) and explore options in the issue you opened?

ZoeBijl:
Why are there two different designs for this pattern? A less and more accessible one? Why not only have an accessible version?

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,791 @@
...
+                  </div>
+                </div>
+
+
+                <div id="myCarousel-items" class="carousel-items" aria-live="off">
+
+                  <div class="carousel-item active"
+                       id="carousel-item-1"
+                       role="tabpanel"
+                       tabindex="0"
```
ZoeBijl: Can we remove this `tabindex` or remove the `a` inside of the `tabpanel`? It’s super confusing to me that there are two different focusables per tab with no apparent visual/operational difference.

jongund: @ZoeBijl This is an open issue for discussion in the working group on whether the tabpanel should always received focus.   RIght now the regression test for tab panel require using the tabindex=0 on the tabpanel.   So the main question is predictability on the part of the screen reader users, should they always expect a tab stop on the tabpanel, or not.    So there is currently a disconnect between the authoring guidance and the regression test for this issue.   I think it should just remain until the group decides whether tab stop on tabpanel is required or optional and we get the regression tests and the practices lined up.

jongund:
> Why are there two different designs for this pattern? A less and more accessible one? Why not only have an accessible version?
> 
> * [ ]  Have only one pattern
> * [ ]  Better reflect the diversity of humanity in the photos
> * [ ]  Consolidate JS to one file/class

There is no one perfect pattern for carousel, so we are trying to provide options and information on different approaches and some approaches are more accessible, but less palatable to designers.  But the more accessible pattern will be easier for people with low vision to use, but many designers would have concerns about the additional real estate needed to place controls and captions outside of the rotating image.

examples/carousel/carousel-2-tablist.html
ZoeBijl: In that case, can we remove the `a`s inside the panels? They don’t seem to do anything other than add a tabstop that confuses people.

ZoeBijl:
> There is no one perfect pattern for carousel

This is true for all design patterns and we don’t add two version there. I’m sure we can design something that’s both accessible and appealing ⭐️.

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,793 @@
...
+                              aria-selected="false"
+                              aria-controls="carousel-item-2">
+                        <svg width="34" height="34" version="1.1" xmlns="http://www.w3.org/2000/svg">
+                          <circle class="border" cx="16" cy="15" r="10" />
+                          <circle class="tab-background" cx="16" cy="15" r="8" />
+                          <circle class="tab" cx="16" cy="15" r="6" />
+                        </svg>
+                      </button>
+                      <button id="carousel-tab-3"
+                              role="tab"
```
smhigley: Could there be `type="button"` on this and the other carousel buttons?

ZoeBijl:
I’m working on a new design for this component. Can someone please sign up to do the visual review?

examples/carousel/carousel-2-tablist.html
ZoeBijl: Good suggestion.

jongund: I will add type=button to the pull request

test/tests/carousel_carousel-2-tablist.js
smhigley: I'd be good with that, since it's not specific to this PR 👍 

I'm still interested in whether the specific hover/focus and play pause behavior has some bugs, but I'm fine with leaving the test for that to a later issue

.eslintrc.json
```diff
@@ -1,4 +1,7 @@
 {
+  "env": {
+    "es6": true
```
smhigley: This PR doesn't seem like the right place for this change. If this isn't going into 1.2 anyway, it could perhaps wait until after we decide how to handle ES6+ syntax, or it could go in sooner using function prototypes instead of classes.

spectranaut: I think this is not the right place to do this @svinkle can you remove this commit and open a different PR to discuss it?

jongund:
@mcking65 @ZoeBijl Updated tablist carousel example, documentation and test case to remove tab stop on tabpanel.

jongund:
@spectranaut Valarie can you look at the test file for the `tablist` carousel example.   It seems to go so far and the testing just stops with no error.  It seems to be waiting for some thing to happen that never does.  Thanks.

jongund:
@mcking65 Code has been updated to use prototoype instead of class, so it should pass all the linting tests

.eslintrc.json
jongund: @smhigley I have changed the tabbed carousel example to use prototoype, instead of class, so should not have linting problems with the current tests

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,779 @@
...
+    <main>
+      <h1>Example of Image Carousel with Automatically Rotating Images using a Tablist</h1>
+        <p>
+          Comments on this example are tracked in
+          <a href="https://github.com/w3c/aria-practices/pull/1120">Pull request 1120.</a>
+        </p>
+      <p>
+        The following example implementation of the
+        <a href="../../#carousel">carousel design pattern</a>
+        demonstrates features of the pattern that are essential to accessibility for carousels that automatically start rotating when the page loads and allow users to directly select which slide is presented.  Slide selction is typically provided using a series of circular buttons rendered over the rotating slides.
```
carmacleod:
```suggestion
        demonstrates features of the pattern that are essential to accessibility for carousels that automatically start rotating when the page loads and allow users to directly select which slide is presented.  Slide selection is typically provided using a series of circular buttons rendered over the rotating slides.
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,779 @@
...
+        Automatic rotation stops when users move focus to any control or link in the carousel or hover the mouse over carousel content.
+        The accessibility features section that follows the example describes these features in detail.
+      </p>
+
+      <section>
+        <h2>Example View Options</h2>
+        <ul>
+          <li><a id="id_link_less_accessible" href="carousel-2-tablist.html">Tablist carousel controls and captions displayed on image</a>: This option is less accessible than the other option because the controls are harder to perceive and captions are more difficult to read with rotating images behind them.</li>
```
carmacleod:
```suggestion
          <li><a id="id_link_less_accessible" href="carousel-2-tablist.html">Tablist carousel controls and captions displayed on image</a>: This option is less accessible than the one below because controls are harder to perceive and captions are more difficult to read with rotating images behind them.</li>
```

examples/carousel/css/carousel-tablist.css
```diff
@@ -0,0 +1,236 @@
...
+  stroke: white;
+  opacity: 1;
+}
+
+.carousel-tablist [role="tab"]:hover circle.border,
+.carousel-tablist [role="tab"]:hover circle.tab {
+  fill: #005a9c;
+  stroke: #005a9c;
+  opacity: 1;
+}
```
carmacleod: When tablist has focus and _mouse_ is used to select another tab, change circle.tab color to "hovered and selected" color (i.e. white center)

```suggestion
}

.carousel-tablist [role="tab"][aria-selected="true"]:hover circle.tab {
  fill: white;
}
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,779 @@
...
+          People with low vision or a cognitive disability that affects visual processing or reading benefit from being able to control slide rotation so they have sufficient time to explore slide content.
+          Similarly, since screen reader users cannot perceive automatic rotation, it can make reading the page confusing and disorienting.
+          For example, when slides are automatically rotating, a screen reader user may read an element on slide one, execute a screen reader command to read the next element, and, instead of hearing the next element on slide one, hear an element from slide 2 without any knowledge that the element just announced is from an entirely new context.
+        </p>
+        <p>This example includes the following features for giving users control over slide rotation:</p>
+        <ul>
+          <li>
+            Hovering the mouse over any carousel content pauses automatic rotation.
+            Automatic rotation resumes when the mouse moves away from the carousel unless another condition, such as keyboard focus,  that prevents rotation has been triggered.
```
carmacleod:
```suggestion
            Automatic rotation resumes when the mouse moves away from the carousel unless another condition that prevents rotation, such as keyboard focus, has been triggered.
```

carmacleod:
Found a strange bug during functional testing (in Chrome and Firefox on Windows, so far).

Steps:
1. Using the mouse, click anywhere in the tablist.
2. Type Shift+Tab. Focus should be on the rotation control (pause/play) button.
3. Type Space or Enter, nothing happens.

The rotation control button usually works fine with Space or Enter, but the above steps somehow cause it to stop working. Clicking on it with the mouse will make it work again.

jongund:
@carmacleod 

Is the mouse somewhere inside the carousel when you press shift-tab?
Anytime the mouse is anywhere inside the carousel the rotation is stopped.   So the pause/start button is disabled.

carmacleod:
Aha! Good call, @jongund - that was it. The carousel actually extends further up than I thought, so I was clicking on the tablist and moving the mouse to a place that I thought was "above the carousel", but in fact it is still inside it, so the play/pause button was disabled, and I didn't realize it.

Here's a screen snap - notice the invisible rectangle above the carousel (I assume this is reserved space for the option that has the controls above the carousel). Maybe the version with controls ontop of the image could collapse this space. :)  Maybe also the play/pause button could have a disabled look when it's disabled, so that folks like me can figure out what's going on. ;)

image

smhigley:
That said, since the passing-around-prototype-instances pattern is already present in other examples, this one doesn't need to hold up merging (as discussed on the APG call).

We can address the pattern across the board later when we work on a broader code structure/quality pass instead 👍

jongund:
@smhigley 

The new tabbed carousel example is using just one object in the [carousel-tablist.js file](https://github.com/w3c/aria-practices/blob/issue955-add-tabbed-carousel-example/examples/carousel/js/carousel-tablist.js) for implementing the example.  So it should be compliant with the new coding practices. 

The previous carousel example with just previous and next buttons need to be updated at some point to use the new coding practices, but that is not part of this pull request.

carmacleod:
Ok, I'm done the functional testing, including Chrome and Firefox on Windows, Safari and Chrome on MacOS, and Safari on my teeny-tiny iPhone. In general, this works very well, and it looks good. (A few specific feedback comments are given above). Nice work, @jongund - thanks!

examples/carousel/carousel-2-tablist.html
jongund: @carmacleod fixed

examples/carousel/carousel-2-tablist.html
jongund: @carmacleod fixed

examples/carousel/carousel-2-tablist.html
jongund: @carmacleod fixed

carmacleod:
Looks nice, @jongund - thanks!!

Can you also make the 2 enhancements that were suggested in https://github.com/w3c/aria-practices/pull/1120#issuecomment-574309631 ?
i.e.
- collapse the extra 50px padding above the controls-on-image carousel (I see that it has a very light grey background now, but that doesn't help much)
- make the play/pause button have a disabled look when the mouse is over the carousel (maybe add `disabled` and use css [disabled])

If you prefer, I can open a new issue to address these enhancements at a later time.

jongund:
@carmacleod 

I made some updates to fix the padding issue.  Let me know what you think.

carmacleod:
@jongund 
> I made some updates to fix the padding issue. Let me know what you think.

Nice! Much better - thanks!

examples/carousel/carousel-2-tablist.html
```diff
@@ -11,42 +11,47 @@
...
         <li><a href="https://github.com/w3c/aria-practices/issues/new">Report Issue</a></li>
         <li><a href="https://github.com/w3c/aria-practices/projects/10">Related Issues</a></li>
         <li><a href="../../#carousel">Design Pattern</a></li>
       </ul>
-    </nav>
-    <main>
-      <h1>Example of Image Carousel with Automatically Rotating Images using a Tablist</h1>
-        <p>
+    </na
+    centerTablist(controls, items);v
```
carmacleod: This looks like some javascript was accidentally pasted into the middle of the `</nav>` tag.

jongund: @carmacleod 

Fixed, Carolyn thank you for all you feedback

mcking65:
I just noticed that we have 2 html files -- one for each of the two different visual treatments (captions on overlayed on images and captions below images). The content in the files is duplicate. If we are going to ship this with multiple visual designs but otherwise identical content and function, let's remove the duplication to simplify maintenance.

test/tests/carousel_carousel-2-tablist.js
spectranaut: that is a reasonable, but this test is also otherwise complete so I think it's fine to leave as is!

test/tests/carousel_carousel-2-tablist.js
spectranaut: We added a way to do this re: Sarah's request: https://github.com/w3c/aria-practices/pull/1227

test/tests/carousel_carousel-2-tablist.js
spectranaut: context.session.wait actually is a selenium webdriver command which runs the command in a loop until the conditions are met or the time out occurs -- so it's pretty impossible for it to have looped around luckily!

spectranaut: but here is the link to the docs for the curious: https://selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_ThenableWebDriver.html#wait

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,799 @@
...
+          For example, when slides are automatically rotating, a screen reader user may read an element on slide one, execute a screen reader command to read the next element, and, instead of hearing the next element on slide one, hear an element from slide 2 without any knowledge that the element just announced is from an entirely new context.
+        </p>
+        <p>This example includes the following features for giving users control over slide rotation:</p>
+        <ul>
+          <li>
+            Hovering the mouse over any carousel content pauses automatic rotation.
+            Automatic rotation resumes when the mouse moves away from the carousel unless another condition that prevents rotation, such as keyboard focus, has been triggered.
+          </li>
+          <li>
+            Moving keyboard focus to any of the carousel content, including the next and previous slide elements, pauses automatic rotation.
```
carmacleod: This example does not have next and previous slide elements.
So either need to remove [**, including the next and previous slide elements,**]
or replace it with something else, such as [**, including the tab elements,**]

```suggestion
            Moving keyboard focus to any of the carousel content, including the tab elements, pauses automatic rotation.
```

a11ydoer:
@carmacleod Thanks for the review. I have a minor and picky request. 

I am just wondering whether we can make "Example of Image Carousel with Automatically Rotating Images using a Tablist" more concise. I know you intend to put emphasis on "Tablist" which is contrasting basic "Auto-Rotating Image Carousel Example" 

How about make this  "Auto-Rotating Image Carousel using a Tablist"?

a11ydoer:
@carmacleod Ah...that is the example name...

mcking65:
@a11ydoer, I agree, the title was pretty verbose. I shortened it to:

>Example of Rotating Image Carousel Using  a Tab List

I made the change to the title tag, h1, and text of referencing links.

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+}
+
+CarouselTablist.prototype.updateRotation = function () {
+
+  // Pause rotation when keyboard focus enters the carousel
+  if (this.hasFocus) {
+    this.isStopped = true;
+  }
+
+  if (!this.hasHover && !this.hasFocus && !this.isStopped) {
```
smhigley: `!this.hasFocus` here is unnecessary in combination with the preceding block

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+    this.rotate = true;
+    this.liveRegionNode.setAttribute('aria-live', 'off');
+  }
+  else {
+    this.rotate = false;
+    this.liveRegionNode.setAttribute('aria-live', 'polite');
+  }
+
+  if (this.isStopped) {
```
smhigley: This logic still confuses me -- why is the play/pause button updated when rotation is stopped by focus, but allowed to show the opposite icon and label when the carousel is hovered?

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+    this.pauseButtonNode.classList.remove('play');
+    this.pauseButtonNode.classList.add('pause');
+  }
+
+
+}
+
+CarouselTablist.prototype.toggleRotation = function () {
+  if (this.isStopped) {
+    if (!this.hasHover && !this.hasFocus) {
```
smhigley: Rather than nested if statements, a single `if (!this.hasHover && !this.hasFocus && !this.isStopped)` would be better.

Also, it seems like it would be cleaner to combine `toggleRotation` and `updateRotation`, or just drop `toggleRotation` entirely and call `updateRotation` in `handlePauseButtonClick`

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+CarouselTablist.prototype.handleImageLinkFocus = function () {
+  this.liveRegionNode.classList.add('focus');
+}
+
+CarouselTablist.prototype.handleImageLinkBlur = function () {
+  this.liveRegionNode.classList.remove('focus');
+}
+
+CarouselTablist.prototype.handleMouseOver = function (event) {
+  if (!this.pauseButtonNode.contains(event.target)) {
```
smhigley: It seems a bit odd to me that the carousel is paused on hover, but then has an exception for when the cursor is directly over the pause button.

I actually thought it was a bug, particularly when the carousel was playing and then unexpectedly stopped after I'd accidentally knocked the mouse a few pixels over without noticing.

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+
+CarouselTablist.prototype.setSelectedTofirstTabNode = function (moveFocus) {
+  if (typeof moveFocus != 'boolean') {
+    moveFocus = false;
+  }
+  this.setSelectedTab(this.firstTabNode, moveFocus);
+}
+
+CarouselTablist.prototype.setSelectedTolastTabNode = function (moveFocus) {
+  if (typeof moveFocus != 'boolean') {
```
smhigley:
```suggestion
  if (typeof moveFocus !== 'boolean') {
```

smhigley: It looks like there are actually 5 instances of a non-strict equals with `boolean`; it'd be nice to convert them all into a strict equals check for consistency

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+
+  // Start rotation
+  setTimeout(this.rotateSlides.bind(this), this.timeInterval);
+
+}
+
+CarouselTablist.prototype.centerTablistControls = function () {
+  width1 = this.tablistNode.getBoundingClientRect().width;
+  width2 = this.containerNode.getBoundingClientRect().width;
+  width3 = this.pauseButtonNode.getBoundingClientRect().width;
```
smhigley: all three of these should have a `var` declaration to avoid making them globals

examples/carousel/js/carousel-tablist.js
```diff
@@ -0,0 +1,373 @@
...
+    moveFocus = false;
+  }
+  this.setSelectedTab(this.lastTabNode, moveFocus);
+}
+
+CarouselTablist.prototype.rotateSlides = function () {
+  if (this.rotate) {
+    this.setSelectedToNextTab();
+  }
+  setTimeout(this.rotateSlides.bind(this), this.timeInterval);
```
smhigley: this is just a minor note -- this particular implementation (over a `setTimeout`/`clearTimeout` one) works, although it means this function will continue to be called even when paused.

examples/carousel/js/carousel-tablist.js
jongund: @smhigley 
Fixed

examples/carousel/js/carousel-tablist.js
jongund: @smhigley I made the change to combine the if statements, but I think it still needs the toggle function which is not part of updateRotation function.

examples/carousel/js/carousel-tablist.js
jongund: @smhigley 
fixed

examples/carousel/js/carousel-tablist.js
jongund: @smhigley 

Since `this.rotate` is controlled by a number of events and situtation, to me it was just seemed the best way to control the rotation.   For example rotation is paused and started based on hover and keyboard focus, do I need to keep track of all the different ways and conditions to restart rotation?  This seemed easier for me to understand.

examples/carousel/js/carousel-tablist.js
jongund: @smhigley 

fixed

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                        <a href="#!">
+                           Dynamic Europe: Amsterdam, Prague, Berlin
+                        </a>
+                      </h3>
+
+                      <div class="hidden-xs hidden-sm">
+                        <p><span class="contrast">7 pm Tuesday, March 3, on TV</span></p>
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
```
carmacleod:
```suggestion
                  </div>   <!-- .carousel-item -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                        <a href="#!">
+                          Travel to Southwest England and Paris
+                        </a>
+                      </h3>
+
+                      <div>
+                        <p><span class="contrast">Sept. 14 to Sept. 24 or 27</span></p>
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
```
carmacleod:
```suggestion
                  </div>   <!-- .carousel-item -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                        <a href="#!">
+                          Great Children's Programming on Public TV
+                        </a>
+                      </h3>
+
+                      <div>
+
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
```
carmacleod:
```suggestion
                  </div>   <!-- .carousel-item -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                        <a href="#!">
+                          Foyle&#8217;s War Revisited
+                        </a>
+                      </h3>
+
+                      <div>
+                        <p><span class="contrast">8 pm Sunday, March 8, on TV: Sneak peak at the final season.</span></p>
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
```
carmacleod:
```suggestion
                  </div>   <!-- .carousel-item -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                      <h3>
+                        <a href="#!">
+                          Great Britain Vote: 7 pm Sat.
+                        </a>
+                      </h3>
+
+                      <div>
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
```
carmacleod:
```suggestion
                  </div>   <!-- .carousel-item -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                        <a href="#!">
+                          Mid-American Gardener: Thursdays at 7 pm
+                        </a>
+                      </h3>
+
+                      <div class="hidden-xs hidden-sm">
+                        <p><span class="contrast">Watch the latest episodes.</span></p>
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
```
carmacleod:
```suggestion
                  </div>   <!-- .carousel-item -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+
+                </div>
+
+
+              </div> <!-- carousel-inner -->
```
carmacleod:
```suggestion
              </div> <!-- .carousel-inner -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+
+
+            </section><!-- /.carousel -->
```
carmacleod:
```suggestion
            </section> <!-- .carousel-tablist -->
```

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,798 @@
...
+                        </a>
+                      </h3>
+
+                      <div class="hidden-xs hidden-sm">
+                        <p><span class="contrast">Watch the latest episodes.</span></p>
+                      </div>
+                    </div>  <!-- .carousel-caption -->
+                  </div>   <!-- .item -->
+
+                </div>
```
carmacleod:
```suggestion
                </div> <!-- .carousel-items -->
```

jongund:
@carmacleod @smhigley @mcking65 
I updated the example to use a "stopped" pointer/cursor when the mouse hovers over the slide and stops rotation.
Maybe this partly address the issues on the call today.
I could not tell from the minutes of today's meeting if there were other changes requested.

carmacleod:
@jon @mck sorry this took so long to finish.
My main question is: How important is it to use a live region for this example?
It would be more robust if authors simply gave each tab a label that summarizes the slide.
For this example, the tab label could either point to the link text or duplicate the alt text.
Or, the author could decide to summarize the slide in some other way.
The live region usually - but not robustly - reads the link text, alt text and optional extra static text for the newly-selected slide.
Sometimes twice.
And occasionally 3 times.

Anyhow, here are my notes.
❌ follows anything I considered to be a bug, or at least "problematic".
❓ follows anything I considered to be "unexpected", but perhaps not a real problem.

#### Background:
- tab labels are "Slide 1", "Slide 2", etc
- tabpanel labels are "1 of 6", "2 of 6", etc (although this info is spoken for the tablist, so not sure it needs to be spoken again when tabpanel gets focus?)
- each tabpanel has roledescription="slide", label, image with alt, link text, and optional static text
- tabpanels (slides) are removed from/added to the live region using CSS display none/block

----

#### Chrome + JAWS
- changing tabs (left/right arrow): tab label, "tab selected, n of m", link text, static text
(doesn't read image alt ❓)
- focus slide: slide label, "slide", image alt

#### Chrome + NVDA
- changing tabs (left/right arrow): image alt, link text, static text, tab label, "tab selected, n of m"
- focus slide by typing tab from tablist: slide label, "slide", image alt

#### FF + JAWS
- changing tabs (left/right arrow): link text, static text, tab label, "tab selected, n of m", link text, static text
(doesn't read image alt ❓)
(reads live region contents twice ❌)
- focus slide: slide label, "slide", image alt

#### FF + NVDA
- changing tabs (left/right arrow): image alt, link text, static text, tab label, "tab selected, n of m"
- focus slide: slide label, "slide", image alt

#### CrEdge + Narrator
- changing tabs (left/right arrow): tab label, "tab item, n of m, selected", image alt (twice ❌), link text (twice ❌), static text
- Narrator eats arrow keys in tablists. Need to arrow to each dot image and then type space to switch tabs. ❌
- focus slide: slide label, "slide", image alt

#### CrEdge + JAWS
- changing tabs (left/right arrow): tab label, "tab selected, n of m"
(doesn't read live region ❌)
- focus slide: slide label, "slide", image alt

#### CrEdge + NVDA
- changing tabs (left/right arrow): image alt, link text, static text, tab label, "tab selected, n of m"
- focus slide: slide label, "slide", image alt

#### macOS Chrome + VO
- changing tabs (left/right arrow): tab label, "selected, tab, n of m", image alt (twice ❌), link text (three times ❌), static text
- focus slide: image alt, slide label, "slide"

#### macOS Safari + VO
- changing tabs (left/right arrow): tab label, "selected, tab, n of m"
(doesn't read live region ❌)
- focus slide: image alt, slide label, "slide"

#### iOS Safari + VO
- live region doesn't turn off when focusing tablist, i.e. slides keep rotating automatically ❌
- navigate to a slide: live region keeps going, VO attempts (and fails) to keep up and even valiantly tries to describe each image as it flashes by: "blue sky, cliff, water; three adults; a woman and a child..." ❌
- navigate to a link child of a slide: because slides are still rotating, the link disappears, and focus goes to the next slide

----

#### Additional notes (just for info):
- JAWS doesn't read pause/play button label change ❓ 
- VO reads pause/play button label change if using ctrl+option+space (but not if just typing space)
- NVDA and Narrator both read pause/play button label change ✔️ 
- color contrast and high contrast was tested by @ZoeBijl during visual design review 🎉

jongund:
@mcking65 @carmacleod 

Added link options to start the carousel in paused mode or to disable auto-rotation.

I also updated the code for the "stop" pointer when hovering the image to only appear when the carousel is auto-rotating.

carmacleod:
@smhigley's preview link: https://raw.githack.com/w3c/aria-practices/tabbed-carousel-opt2/examples/carousel/carousel-2-tablist.html

jongund:
@smhigley 
Can we wait to merge your changes to the carousel, I have been thinking about all the checkoxes and I think this is to complicated of an approach and will most likely be confusing to authors.  I think it should be simpler and we should be using links, like the example has now.   We don't need to show every permutation, just the major features and why they are important for accessibility.

smhigley:
@jongund are you thinking about code complexity, or the experience of checking/unchecking different options on the page?

jongund:
@smhigley 

Some of my concerns:
1. Changes in the first and the last checkboxes result in immediate changes to the visual UI, but the second checkbox requires a reload for it to take effect, this should at least be documented.
2. Right now it looks like the checkboxes are part of the example, some authors may think we are advocating for the inclusion of these user options in an actual design.  In limited attention people initially pay to examples, some authors will immediately discount the example since it doesn't look like what they are trying to do.  We are a very visual society when it comes to the web.
3. If the working group really likes the checkbox approach, the checkboxes should at least be moved to a section titled "Example Options" and placed after the actual example, so they are not the first thing people see when they come to the page.
4. Maybe my biggest concern it seems like most of the accessibility option information should be in the guidance docoument and the guidance document should point to examples demonstrating the features.  The current approach seems to be mixing up guidance information within the examples.  The example should document the accessibility features of this particular example and like other examples reference complementary examples.

jongund:
@smhigley 

Would it be possible to factor out the checkbox option feature from the other code changes, so we could at least merge the changes in behavior to the current PR?

smhigley:
@jongund I like some of those suggestions! I think we could definitely move the checkboxes to a different section under something like an "Example Options" heading, and the wording for each checkbox would certainly benefit from better wording that explains exactly what they do (I think @carmacleod mentioned she'd be up for taking a look at doing that in the call).

As far as the general approach of having same-page behavior/view options, I actually think it's a  better path going forward than having separate links to different variations. As a brief summary, here's where my thinking is regarding pros/cons of using checkboxes (or other same-page controls):

Pros:
- Having one URL for a single component is common across other UI libraries, and would likely be expected by visitors
- It's common to link directly to example pages, and someone browsing quickly might miss that there are other behavior and style options in other links
- People can still link directly to one particular view and behavior
- It fits in nicely with other common ways of documenting controls and control behavior, such as Storybook JS
- It would allow us to create a standard method of defining components along the lines of `new AriaWidget(selector, options)`, where the URL parameters define custom options. I'd actually be interested in updating the carousel to use this format now.

Cons:
- We need to document exactly what each behavior and style option does, and why it's presented, or it could be confusing
- Having an options section next to the example could be confusing / we should visually separate them better

I'm not entirely sure I understand what you mean by mixing guidance documentation within the examples, since I think the earlier link implementation did the same thing -- just using links and URL parameters instead of checkboxes with URL parameters. Perhaps I'm missing something there? We can still link to specific variations from within the guidance section, and the different variations have always shared the same guidance doc, as far as I can tell.

Does that help address some of your concerns? If not, another option we could explore is to just put multiple carousels on the same page, like we do with other examples (e.g. the data grid).

jongund:
@smhigley 

One idea I would like to see tried is to use links, rather than checkboxes to move between options. that way it is clear to less skilled people using the practices that the links define certain behaviors for the carousel. 

The two consistent links would be for the "moreaccessible" option of where captions are:
carousel-2-tablist.html?moreaccessible=true
carousel-2-tablist.html?moreaccessible=false

Two additional links would change based on the current state of more accessible:
carousel-2-tablist.html?moreaccessible=true&paused=true
carousel-2-tablist.html?moreaccessible=true&norotate=true
 
or

carousel-2-tablist.html?moreaccessible=false&paused=true
carousel-2-tablist.html?moreaccessible=false&norotate=true

There would be some visual indication as a part of or near each link indicating which of the two "more accessible" options is being shown on the page and whether either of the "paused" or "norotate" option is selected.

Does this make any sense to you?

smhigley:
I think perhaps the reason I don't really understand is because I do actually find the checkbox functionality (or switches, or any other form of same-page control) easier to use and more discoverable to everyone, not just skilled users.

I'd actually assume it'd be more confusing to someone less tech-savvy to click a link and come back to the same page, just with slightly different example behavior. For all users, my assumption was that not needing to reload the page to explore different options would be preferable.

It does seem like something to discuss on the call today. If generally people think that links are more user-friendly, then we can certainly go back to using those. Another option I think we should consider is having multiple carousels on the page with different behavior, like we have for other examples. That would allow us to document each behavior separately, and not require any action on the user's part to view all options.

jongund:
@smhigley 
Sara I am going to merge your changes later today or tomorrow morning and make some of the other changes from the call yesterday.   I will let you know so you can review the changes and suggest or make edits yourself.

smhigley:
@jongund sounds good! I added a PR here: https://github.com/w3c/aria-practices/pull/1377, merge whenever you're ready

jongund:
@smhigley 

Updates:
1. Merges Sara's changes for pause behavior
1. Updated the initialization so checking the checkboxes in the source code set initial features if there are no/some params in the URL.  Initialization was pretty brittle if you want to make any changes to the default settings.
1. Updated test cases and `arai-live` for new pause behavior
1. Updated focus styling for improved high contrast and compatibility with other recent examples

jongund:
@zcorpan 

Simon I added a test for `tab-aria-label`.

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,796 @@
...
+            Auto-rotation is initially paused (<a href="javascript:history.go(0)"><img class="reload" src="images/reload-icon.png" alt="">reload needed</a>).
+          </label>
+          <p id="rotation-paused-desc">This option controls whether the carousel is paused or playing on page load. The paused option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation, but allows users to start auto-rotation using the start/stop button.  If this option is changed, the page must be reloaded to see the effect on behavior.</p>
+
+          <label for="rotation-disabled">
+            <input type="checkbox" id="rotation-disabled" value="norotate" aria-describedby="rotation-disabled-desc">
+            Auto-rotation is disabled.
+          </label>
+          <p id="rotation-disabled-desc">This option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation by disabling the auto-rotation feature and removing the start/stop button from the user interface.  Users can use the tablist to manually navigate through the slides.</p>
```
carmacleod, line 68: Some new words for the checkboxes. Followed [N&N Guideline 7](https://www.nngroup.com/articles/checkboxes-vs-radio-buttons/): 
> Use positive and active wording for checkbox labels

```suggestion
          <label for="accessible-controls">
            <input type="checkbox" id="accessible-controls" value="moreaccessible" aria-describedby="accessible-controls-desc">
            Display controls and captions outside of image
          </label>
          <p id="accessible-controls-desc">This option is more accessible than rendering controls and captions within the image, because controls are easier to perceive and captions are easier to read without rotating images behind them.</p>

          <label for="rotation-paused">
            <input type="checkbox" id="rotation-paused" value="paused" aria-describedby="rotation-paused-desc">
            Pause auto-rotation on load (<a href="javascript:history.go(0)"><img class="reload" src="images/reload-icon.png" alt="">reload needed</a>).
          </label>
          <p id="rotation-paused-desc">This option controls whether the carousel is paused or playing on page load. The paused option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation, but allows users to start auto-rotation using the start/stop button.</p>

          <label for="rotation-disabled">
            <input type="checkbox" id="rotation-disabled" value="norotate" aria-describedby="rotation-disabled-desc">
            Disable auto-rotation
          </label>
          <p id="rotation-disabled-desc">This option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation by disabling the auto-rotation feature and removing the start/stop button from the user interface. Users can use the tablist to manually navigate through the slides.</p>
```

carmacleod: I also think the description text should be indented to align vertically with the label text.

Also, other settings pages that use checkboxes with descriptions usually make the description text lighter than the label text, so we could make the label text bold, so that it is more clearly the "call to action".

jongund:
@carmacleod 

1. Updated focus styling to be solid, like new chrome focus styling
1. Updated checkbox options with CM suggested changes
1. Initialization of paused states now includes operating system preferences for reduced motion or animations

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+            <input type="checkbox" id="accessible-controls" value="moreaccessible" aria-describedby="accessible-controls-desc">
+            Display controls and captions outside of image.
+          </label>
+          <p id="accessible-controls-desc">This option is more accessible than rendering controls and captions within the image, because controls are easier to perceive and captions are easier to read without rotating images behind them.</p>
+
+          <label for="rotation-paused">
+            <input type="checkbox" id="rotation-paused" value="paused" aria-describedby="rotation-paused-desc">
+            Pause auto-rotation on load (<a href="javascript:history.go(0)"><img class="reload" src="images/reload-icon.png" alt="">reload needed</a>).
+          </label>
+          <p id="rotation-paused-desc">This option controls whether the carousel is paused or playing on page load. The paused option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation, but allows users to start auto-rotation using the start/stop button.</p>
```
carmacleod, line 62:
```suggestion
          <p id="rotation-paused-desc">This option controls whether the carousel is paused or playing on page load. The paused option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation, but allows users to start auto-rotation using the start/stop button. Note: If the user has chosen reduced motion in system settings, auto-rotation will always be paused on load.</p>
```

smhigley:
@jongund I liked your suggestion and updates to make the behavior options less hard-coded, and I had a thought to make it even easier to update and change, and bring the coding style more in line with common UI component patterns.

I made an update to the way the defaults are checked and coded that I think simplifies the code even further (keeping the same behavior where it respects the checkbox settings in the HTML + the URL params).

The second change I made was just to move the `prefers-reduced-motion` logic directly into the carousel component, so anyone who copies the code will get that behavior.

jongund:
@smhigley 
@mcking65 

There needs to be some coordination on this pull request, I have a separate pull request I was in the middle of making changes based on todays call and all of a sudden my changes have been overridden and a bunch of files I have a separate pull request on have been changed in this pull request.

jongund:
@mcking65 
@carmacleod 
@smhigley 

There are changes to some files in this pull request that probably should not be included:
1. `aria-practices.html`
1. `carousel-1.html`
1. `test/tests/carousel_carousel-1.js`

I am not sure how these files got changed in this pull request, and it was most likely me since this pull request is been around for along time (7 months).   I think this pull request should be limited to just adding the tabbed carousel example, so I would like to remove the changes to these files from the pull request.    Since they are minor changes anyway and I have other pull requests that address the issues in these changes.

Thank you especially to Sara and Carolyn for all the work you have done in improving the tabbed carousel example these past few months, looking forward to getting this merges soon.

examples/carousel/carousel-2-tablist.html
```diff
@@ -0,0 +1,797 @@
...
+
+          <label for="rotation-paused">
+            <input type="checkbox" id="rotation-paused" value="paused" aria-describedby="rotation-paused-desc">
+            Pause auto-rotation on load (<a href="javascript:history.go(0)"><img class="reload" src="images/reload-icon.png" alt="">reload needed</a>).
+          </label>
+          <p id="rotation-paused-desc">This option controls whether the carousel is paused or playing on page load. The paused option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation, but allows users to start auto-rotation using the start/stop button. Note: If the user has chosen reduced motion in system settings, auto-rotation will always be paused on load.</p>
+
+          <label for="rotation-disabled">
+            <input type="checkbox" id="rotation-disabled" value="norotate" aria-describedby="rotation-disabled-desc">
+            Disable auto-rotation.
```
carmacleod, line 66: Nit: remove periods from checkbox labels.
```suggestion
            Display controls and captions outside of image
          </label>
          <p id="accessible-controls-desc">This option is more accessible than rendering controls and captions within the image, because controls are easier to perceive and captions are easier to read without rotating images behind them.</p>

          <label for="rotation-paused">
            <input type="checkbox" id="rotation-paused" value="paused" aria-describedby="rotation-paused-desc">
            Pause auto-rotation on load (<a href="javascript:history.go(0)"><img class="reload" src="images/reload-icon.png" alt="">reload needed</a>)
          </label>
          <p id="rotation-paused-desc">This option controls whether the carousel is paused or playing on page load. The paused option improves accessibility for users with visual impairments and people who are distracted or confused by auto-rotation, but allows users to start auto-rotation using the start/stop button. Note: If the user has chosen reduced motion in system settings, auto-rotation will always be paused on load.</p>

          <label for="rotation-disabled">
            <input type="checkbox" id="rotation-disabled" value="norotate" aria-describedby="rotation-disabled-desc">
            Disable auto-rotation
```

carmacleod:
@jongund I pushed 2 minor changes in:
- Add note about reduced motion setting to "pause on load" checkbox
- Remove periods from checkbox labels

Hope that's ok.

So the only thing I'm still wondering about is the `href="#!"`.

jongund:
@caribouW3 

I fixed the `href`s to use `#`,  not sure where the `#!` came from.

Thanks for adding the note on the pause at start up.

Jon

mcking65:
@jongund, thank you, thank you for all your work on this!!

Issue #1151, opened 2019-08-29
Date picker spin button example broken in Safari


backwardok:
The [example for the `spinbutton` role](https://www.w3.org/TR/wai-aria-practices/examples/spinbutton/datepicker-spinbuttons.html) works well in Chrome, but has a javascript error in Safari. 

On load, there's a SyntaxError: `The string did not match the expected pattern.` on this line in [spinbutton-date.js](https://www.w3.org/TR/wai-aria-practices/examples/spinbutton/js/spinbutton-date.js):
```domNode.querySelector('[role="spinbutton');```

My guess is that Chrome is being generous with the selector here and Safari is breaking because it's technically an invalid selector. It should be `domNode.querySelector('[role="spinbutton]');` (it was missing the closing right bracket)

JAWS-test:
Further problems at https://www.w3.org/TR/wai-aria-practices/examples/spinbutton/datepicker-spinbuttons.html:
* When increasing the number for the day over 31, "NaN" is displayed.
* When decreasing number for the day below 1, the top number freezes at 31 and the two bottom numbers become negative.

JAWS-test:
JAWS outputs: "To set the value use the Arrow keys **or type the value**", because normally free text input in spin button fields is possible. But not here. Maybe readonly has to be used? Or the text input is possible?

backwardok:
Verified that the issue was not happening in Chrome, Firefox, and Edge, but is also happening in IE11. Working on a fix for this specific issue

@JAWS-test can you file separate issues for the other things you're finding?

backwardok:
@JAWS-test made a fix for the issue with the NaN: https://github.com/w3c/aria-practices/pull/1154

JAWS-test:
@backwardok 

> can you file separate issues for the other things you're finding?

https://github.com/w3c/aria-practices/issues/1155

Thank you so much for fixing the other problems so quickly!

mcking65:
Thanks to @backwardok, this is fixed with commit e451c93!

PR #1153, opened 2019-08-29
Fix spinbutton example query selector

backwardok:
Fixes issue: https://github.com/w3c/aria-practices/issues/1151

The query selector for the `spinbutton` element was incomplete, which caused it to not query properly in Safari or IE. 

This fix adds in the missing characters to the query so that it queries properly.

Before, clicking on the arrows didn't do anything and the element couldn't be tabbed to:
Video of clicking on arrows not changing the values in the datepicker

After, clicking on the arrows updates the value up/down and when tabbing to the picker values, using the up/down arrow keys adjusts the values:
Video of clicking on arrows updating values up or down, and then keyboard focus on values with the values changing in the datepicker

### DIFF
--- a/examples/spinbutton/js/spinbutton-date.js
+++ b/examples/spinbutton/js/spinbutton-date.js
@@ -14,7 +14,7 @@ var SpinButtonDate = function (domNode, values, callback)  {
 
   var initialValue = domNode.getAttribute('aria-valuetext');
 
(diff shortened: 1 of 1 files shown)


backwardok:
Also sorry about the infinite gifs - the tool I used doesn't have an option to not loop

jnurthen:
Thanks Diane. @mcking65 I think this can be merged.

PR #1154, opened 2019-08-29
Spin Button Date Picker Example: Fix incrementing and decrementing past limits

backwardok:
The date picker spin button example would show NaN on all the dates if you tried to loop around to the first date. Also, despite there being a min and max year set, the year could be cycled through potentially infinitely beyond the min or max year.

This was due to the way the min and max values were being queried. The values are set on the DOM nodes for the node with the `spinbutton` role. However, the node that was being queried for those values was the parent node.

This change fixes the node used so that the initial values are set appropriately, and the spinners work as expected.

Before, once you tried to wrap around beyond the last day of the month, all the values would then be set to NaN, and there'd be no way to get it to go back to having real numbers without refreshing the page. Additionally, the years could be incremented (or decremented) as high as desired, even though the min year was set to 2019 and the max year is set to 2040. 
Video of tabbing through date spinners and where NaN appears after going past 31 for August and the year is set up to 2117

After, you can wrap around the days of the month. Also, 2019 shows up as the min year (with no previous year showing) and 2040 is the max year (with no next year showing). 
Video of tabbing through date spinners and the date loops properly, with the year starting at 2019 and ending at 2040

### DIFF
--- a/examples/spinbutton/js/spinbutton-date.js
+++ b/examples/spinbutton/js/spinbutton-date.js
@@ -35,10 +35,10 @@ var SpinButtonDate = function (domNode, values, callback)  {
     }
   }
   else {
-    this.valueMin  = parseInt(domNode.getAttribute('aria-valuemin'));
-    this.valueMax  = parseInt(domNode.getAttribute('aria-valuemax'));
-    this.valueNow  = parseInt(domNode.getAttribute('aria-valuenow'));
-    this.valueText = domNode.getAttribute('aria-valuenow');
+    this.valueMin  = parseInt(this.spinbuttonNode.getAttribute('aria-valuemin'));
+    this.valueMax  = parseInt(this.spinbuttonNode.getAttribute('aria-valuemax'));
+    this.valueNow  = parseInt(this.spinbuttonNode.getAttribute('aria-valuenow'));
+    this.valueText = this.spinbuttonNode.getAttribute('aria-valuenow');
   }
 
   this.keyCode = Object.freeze({


mcking65:
#### Preview For Testing

[View the spin button example in the compare branch for this PR](https://raw.githack.com/backwardok/aria-practices/backwardok-fix-spinbutton-nan/examples/spinbutton/datepicker-spinbuttons.html)

mcking65:
@backwardok, thank you very much for this fix!

Looks good to me -- looking at the code.

However, this is my first time testing this example in Safari on Mac ... I can't get it to work at all. I wonder if VoiceOver is somehow not reporting the value or accessible name correctly. The value is always Jun 30, 2019, no matter what I do.

I even tried passing through the up/down arrow keys with ctrl-opt-tab. When that didn't work, I made sure focus was on a spinner, turned off VoiceOver, then pressed up several times, then turned VO back on. VO still reported the same date for the group of spinners.

I tried activating the previous/next buttons with ctrl-opt-space. That did not change the date either.

It all works in Chrome with VoiceOver. Although, it is pretty painful that VO does not read the values as they change. You have to manually read them with ctrl-opt-w or ctrl-opt-f4. Amazing how broken VO is with these spinners.

I'll put this on the review agenda for September 3 meeting. I thought I'd be able to test and merge, but apparently this is not testable by a VoiceOver user -- at least not in Safari.

backwardok:
Hey @mcking65 - it's not testable in Safari because of the bug that I have a fix for in https://github.com/w3c/aria-practices/pull/1153 - you won't be able to test it in Safari until that one is merged (though you could theoretically merge the two together and see the full fix).

mcking65:
In Safari, the label on the group is announced as "Choose a date, current value undefined" followed by the date, e.g., August 31, 2019. In Chrome, and with Windows screen readers, the label is correct, e.g., Choose a date, current value is Saturday August 31, 2019. Is the function that provides the name of the day of the week not supported in Safari?

Also, when using VoiceOver on the Mac in either Safari or Chrome, the value of the year stepper is not announced. When VO reads the day and month, it will say the value and then the name and then the role of stepper, e.g., "August Month Stepper". For the year, it just says "Year Stepper" and does not announce the value. Is there something different about the way the year value is communicated? I guess with the other 2 steppers we are using aria-valuetext and with the year we are not. So, perhaps VO has a bug where it does not announce the value of a spin button unless aria-valuetext is provided. If that is the case, this is a good test case for VO. Or, is there some other possible cause of this behavior?

backwardok:
@mcking65 I'm not sure, to be honest - I actually aimed to fix the Safari issue because I couldn't test the example with VO before to see what it sounded like. The Chrome behavior definitely seems better, but I don't know if that's an issue with VO + Safari or if there's another bug in the example.

carmacleod:
Tested in Windows 10 on Chrome, FF and Edge.
- Keyboard works great in all 3 browsers.
- JAWS and NVDA both work great in Chrome.
- FF is getting the day of the week wrong (off by 1) in the hidden myDatepickerDate label.
   (for example, today is Tues Sept 10 2019, but the label says Monday September 10 2019)
   (probably need some special case code for FF?)
- JAWS/Edge is eating the arrow keys in the spin buttons (probably a JAWS/Edge bug)
- JAWS/Edge also doesn't announce the name of the parent group (again, probably a JAWS/Edge bug)
- NVDA/Edge does spin the buttons on arrow keys, but it only speaks values from 0 - 100 (probably an NVDA/Edge bug)

Issue #914, opened 2018-10-01
Menu example: Update to illustrate use of labeled groups for grouping menuitem, menuitemradio,  and menuitemcheckbox elements


mcking65:
Resolution of w3c/aria#737 changes ARIA 1.2 to allow grouping menuitemcheckbox elements with the group role. Add an APG 1.2 menu example that demonstrates how to use and label group elements inside of menus.

Decision to make before doing: Since the change in 1.2 is to support for menuitemcheckbox but 1.1 already supported labeled groups on menuitemradio and menuitem, we could add a labeled group to a set of menuitemradios or menuitems in one of the existing examples and include the same example in both APG 1.1 and APG 1.2. Is this sufficient?

Note: if the example is going to be different for APG 1.2, then it should be a new example (differently named directory). And, we should remove the most similar 1.1 example from APG 1.2. This will avoid merge conflicts.

This is related to issue #764.

PR #1144, opened 2019-08-28
Add a group around the menuitemcheckboxes

jnurthen:

PR #1160, opened 2019-09-01
Editor Menubar Example: add group around font style menu item checkboxes

mcking65:
To resolve issue #914, adds role group around the menuitemcheckbox elements for font style.
Also updates applicable regression test for the group role.
Also adds missing role="none" on `<li>` elements.

Uses commits from #1144 by @jnurthen, but merges to master instead of apg-1.2.

#### Preview Link

[View the editor menubar in the compare branch for this PR](https://raw.githack.com/w3c/aria-practices/issue914-menubar2-add-fontstyle-group/examples/menubar/menubar-2/menubar-2.html)

### DIFF
--- a/examples/menubar/menubar-2/menubar-2.html
+++ b/examples/menubar/menubar-2/menubar-2.html
@@ -62,8 +62,8 @@
           <ul role="menu" aria-label="Style/Color">
             <li role="none">
               <ul role="group" rel="font-style" aria-label="Font Style">
-                <li role="menuitemcheckbox" rel="font-bold" aria-checked="false">Bold</li>
-                <li role="menuitemcheckbox" rel="font-italic" aria-checked="false">Italic</li>
+                <li role="menuitemcheckbox" data-option="font-bold" aria-checked="false">Bold</li>
+                <li role="menuitemcheckbox" data-option="font-italic" aria-checked="false">Italic</li>
               </ul>
             </li>
             <li role="separator"></li>
--- a/examples/menubar/menubar-2/menubar-2.html
+++ b/examples/menubar/menubar-2/menubar-2.html
@@ -60,10 +60,14 @@
         <li role="none">
           <a role="menuitem" aria-haspopup="true" aria-expanded="false" href="#">Style/Color</a>
           <ul role="menu" aria-label="Style/Color">
-            <li role="menuitemcheckbox" data-option="font-bold" aria-checked="false">Bold</li>
-            <li role="menuitemcheckbox" data-option="font-italic" aria-checked="false">Italic</li>
+            <li role="none">
+              <ul role="group" rel="font-style" aria-label="Font Style">
+                <li role="menuitemcheckbox" rel="font-bold" aria-checked="false">Bold</li>
+                <li role="menuitemcheckbox" rel="font-italic" aria-checked="false">Italic</li>
+              </ul>
+            </li>
             <li role="separator"></li>
-            <li>
+            <li role="none">
               <ul role="group" data-option="font-color" aria-label="Text Color">
                 <li role="menuitemradio" aria-checked="true">Black</li>
                 <li role="menuitemradio" aria-checked="false">Blue</li>
@@ -71,7 +75,7 @@
                 <li role="menuitemradio" aria-checked="false">Green</li>
               </ul>
             <li role="separator"></li>
-            <li>
+            <li role="none">
               <ul role="group" data-option="text-decoration" aria-label="Text Decoration">
                 <li role="menuitemradio" aria-checked="true">None</li>
                 <li role="menuitemradio" aria-checked="false">Overline</li>
--- a/examples/menubar/menubar-2/menubar-2.html
+++ b/examples/menubar/menubar-2/menubar-2.html
@@ -61,7 +61,7 @@
           <a role="menuitem" aria-haspopup="true" aria-expanded="false" href="#">Style/Color</a>
           <ul role="menu" aria-label="Style/Color">
             <li role="none">
-              <ul role="group" rel="font-style" aria-label="Font Style">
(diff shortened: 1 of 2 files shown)


mcking65:
@jnurthen, the only thing I changed from #1144 was to add `role="none"` to 3 `<li>` elements.

mcking65:
Tests are failing because, amazingly, the enter and space keys no longer choose any of the style options, e.g., bold or italic. Needs investigation. As I was manually verifying this, I also noticed that some keys are being trapped that should not be trapped when focus is inside the menubar, e.g., alt+d and ctrl+l. We should be sure that we are not unnecessarily capturing modified keys. We only need alpha/numeric keys for type ahead in this example, and they should be unmodified except for shift.

jongund:
@mcking65 @jnurthen 

I fixed the bug with bold and italic, somehow the "data-option" attribute got changed to "rel" attribute.

There was also failing a regression test related to the aria-disabled state.  The test needs to be updated to just look at the font size smaller and font size large menu options, right now it seems to want to check almost every menu item.   I marked that test as a known regression test failure.

mcking65:
@jongund, we are getting the following in the PR test:

```
313 not ok 39 - menubar_menubar-2 › menubar/menubar-2/menubar-2.html [data-test-id="submenu-menuitem-aria-disabled"]: Test aria-disabled="false" for all submenu role="menuitem"s
314   ---
315     name: Error
316    message: >-
317      Test was expected to fail, but succeeded, you should stop marking the test as
318      failing
```

jongund:
@mcking65 
@spectranaut  

 I have commented out he test for now, we could use Valerie's help in getting this test fixed related to `aria-disabled` property to properly test the feature

jongund:
@mcking65 

Looks like this PR can be merged now

jongund:
@mcking65 
@spectranaut 

I fixed the broken test for `aria-disabled` for the font size buttons, so everything should be good for merging this pull request.  Valerie may want to look at my changes, there maybe a better way to do it.

test/tests/menubar_menubar-2.js
```diff
@@ -305,6 +319,26 @@ ariaTest('Test aria-disabled="false" for all submenu role="menuitem"s',
...
+      await menuitem[1].sendKeys(Key.ENTER);
+
+      await menus[3].sendKeys(Key.ARROW_DOWN);
+      await menuitem[0].sendKeys(Key.ARROW_DOWN);
+      await menuitem[1].sendKeys(Key.ENTER);
+
+      await menus[3].sendKeys(Key.ARROW_DOWN);
+      await menuitem[0].sendKeys(Key.ARROW_DOWN);
+      await menuitem[1].sendKeys(Key.ENTER);
+
```
smhigley: After looking into this a little, what's happening is `session.wait` will poll the DOM until the first argument passed to it is truthy, and a side effect of that is if you pass a function first, the function will execute until either it returns true, or the timeout is reached. It seems a bit fragile (as evidenced by breaking in this PR), and I'm not sure the best solution is to add more logic to the function itself.

A better solution might be to directly select X-Small/X-Large once instead of needing to select Smaller/Larger multiple times. I've copied a passing test below that does not rely on `t.context.session.wait`:

```
ariaTest('Test aria-disabled="false" for all submenu role="menuitem"s',
  exampleFile, 'submenu-menuitem-aria-disabled', async (t) => {
    // "aria-disable" should be set to false by default
    await assertAttributeValues(t, ex.submenuMenuitemSelector, 'aria-disabled', 'false');

    const menus = await t.context.session.findElements(By.css(ex.menubarMenuitemSelector));
    const sizeMenu = await t.context.session.findElement(By.css('[aria-label="Size"]'));
    const menuitems = await sizeMenu.findElements(By.css('[role="menuitem"]'));
    const menuItemRadios = await sizeMenu.findElements(By.css('[role="menuitemradio"]'));

    // select X-Small size
    await menus[3].sendKeys(Key.ARROW_DOWN);
    await menuItemRadios[0].sendKeys(Key.ENTER);

    const disabledFirstItem = await menuitems[0].getAttribute('aria-disabled');

    // Test that the item was successfully disabled
    t.true(
      disabledFirstItem === 'true',
      'The first menuitem in the last dropdown should become disabled after X-Small is selected'
    );

    // Select the X-Large size
    await menus[3].sendKeys(Key.ARROW_DOWN);
    await menuItemRadios[menuItemRadios.length - 1].sendKeys(Key.ENTER);

    const disabledSecondItem = await menuitems[1].getAttribute('aria-disabled');

    // Test that the item was successfully disabled
    t.true(
      disabledSecondItem === 'true',
      'The second menuitem in the last dropdown should become disabled after X-Large is selected'
    );
  });
```

jongund: @smhigley this code is a more straightforward test, so I updated the pull request to use your suggested code

examples/menubar/menubar-2/menubar-2.html
```diff
@@ -60,18 +60,22 @@ <h2 id="ex1_label">Example</h2>
         <li role="none">
           <a role="menuitem" aria-haspopup="true" aria-expanded="false" href="#">Style/Color</a>
           <ul role="menu" aria-label="Style/Color">
-            <li role="menuitemcheckbox" data-option="font-bold" aria-checked="false">Bold</li>
-            <li role="menuitemcheckbox" data-option="font-italic" aria-checked="false">Italic</li>
+            <li role="none">
+              <ul role="group" rel="font-style" aria-label="Font Style">
```
nschonni: I don't think the `rel` value is correct. I think that was striped off to the data attributes before
```suggestion
              <ul role="group" data-option="font-style" aria-label="Font Style">
```

mcking65:
@nschonni, good catch, thank you!

Issue #727, opened 2018-06-20
Review 1.2 IDL(Interface Definition Language) button example


mcking65:
The APG 1.2 [IDL Button Example](https://rawgit.com/w3c/aria-practices/apg-1.2/examples/button/button_idl.html) developed for issue #701 is ready for task force review.

annabbott:
No issues noted in [APG 1.2 IDL Button Example](https://rawgit.com/w3c/aria-practices/apg-1.2/examples/button/button_idl.html).

jnurthen:
In order to test this you need to run Chrome Canary and use the --enable-blink-features=AccessibilityObjectModel flag.

scottaohara:
Testing this today with Firefox, Safari and Chrome on macOS.

Both buttons work as expected with Safari and Chrome with VoiceOver on or off.  

The mute button does not change state when mouse clicked with Firefox.
The SVG changes when using space or enter to toggle the button, though aria-pressed does not change.  It doesn't matter if i've turned on the AOM flag in firefox or not.

mcking65:
Resolved with commit 3d83acd.

PR #1161, opened 2019-09-03
Button Examples (IDL Version): Clarify purpose and code for issue 727

carmacleod:
Closes issue #727 
- clarify purpose of example
- remove "similar examples" list
- set reflected attributes in init()
- add "equivalent to" comments

### DIFF
--- a/examples/button/button.html
+++ b/examples/button/button.html
@@ -52,7 +52,7 @@
         <p>This <q>Mute</q> toggle button uses an <code>a</code> element.</p>
         <a tabindex="0" role="button" id="toggle" aria-pressed="false">
           Mute
-          <svg aria-hidden="true">
+          <svg aria-hidden="true" focusable="false">
             <use xlink:href="images/mute.svg#icon-sound"></use>
           </svg>
         </a>
--- a/examples/button/button.html
+++ b/examples/button/button.html
@@ -52,7 +52,7 @@
         <p>This <q>Mute</q> toggle button uses an <code>a</code> element.</p>
         <a tabindex="0" role="button" id="toggle" aria-pressed="false">
           Mute
-          <svg>
+          <svg aria-hidden="true">
             <use xlink:href="images/mute.svg#icon-sound"></use>
           </svg>
         </a>
--- a/examples/button/button_idl.html
+++ b/examples/button/button_idl.html
@@ -57,7 +57,7 @@
         <p>This <q>Mute</q> toggle button uses an <code>a</code> element.</p>
         <a tabindex="0" id="toggle">
           Mute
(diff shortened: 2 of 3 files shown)


carmacleod:
NOTE 1: This looks like a lot of changes, but it's not really. The easiest way to see these code changes is to compare `button.html` with `button_idl.html`, and `button.js` with `button_idl.js`. (Not sure how to do that in github... I was using my IDE to compare).

carmacleod:
NOTE 2: To test, need to do the following:

Chrome: Enable chrome://flags/#enable-experimental-web-platform-features
(note: don't need to use Chrome Canary - just need to enable flag)

Edge: Use [Edge Beta, Dev or Canary](https://www.microsoftedgeinsider.com/en-us/download) and enable edge://flags/#enable-experimental-web-platform-features

Firefox: not supported yet, not even in beta with accessibility.AOM.enabled=true on the about:config page

Safari: [stand by... will test tomorrow]

Preview link for testing: https://raw.githack.com/w3c/aria-practices/car/issue727/examples/button/button_idl.html

JAWS-test:
Problem in IE 11:
* SVG graphic is not displayed
* SVG graphic gets focus (focusable=false missing)

carmacleod:
`[role="button"]`

carmacleod:
Thanks for the review, @JAWS-test !

> Problem in IE 11:
> SVG graphic is not displayed
- All of the CSS in `button.css` keys off of `[role="button"]`, and so if that is not being set properly, then the example looks very strange. I suspect that is what is happening in IE11.

> SVG graphic gets focus (focusable=false missing)
- Thank-you! Added.

> Link "IDL Interface defined in ARIA 1.2" refers to the same page, but should probably refer to ARIA 1.2
- I think (??) that the link will be corrected by respec after the `apg-1.2` branch is merged into `master`. @jnurthen Can you verify that this link is correctly specified? The code is `<a class="specref" href="#idl-interface">`.

carmacleod:
Additional notes:
- @jongund I decided to continue to use `getElementById()` instead of switching to `createElement()`, but I did move the setting of `.role` and `.ariaPressed` into javascript instead of keeping it in the html. I think it's best if the code changes between the button and button IDL versions of the code are minimal so that the IDL changes very easy to see. Hope that's ok. Please compare the `init ()` function in `button.js` and `button_idl.js` to see what I mean.
- @mcking65 I deleted the "Similar examples" list from `button_idl.html` because those examples aren't similar (i.e. they do not use reflection).

carmacleod:
@mcking65, I like all of your updates - thank-you! Feel free to merge. :)

Issue #1132, opened 2019-08-19
Some examples for visually impaired people not accessible


JAWS-test:
Visually impaired users need sufficient contrast and partly "High Contrast Mode" of the operating system. I have checked all examples and found some errors with contrast and "High Contrast Mode". If I have noticed any other errors, I will also list them in the following:

https://w3c.github.io/aria-practices/examples/accordion/accordion.html
Contrast arrow icon: 2.7:1
Contrast border 1.5:1

https://w3c.github.io/aria-practices/examples/dialog-modal/alertdialog.html
Does not work with IE 11 (button "Save", button "Yes")
"Saved" message is shown only for 3 seconds

https://w3c.github.io/aria-practices/examples/button/button.html
IE 11: Icon at "Mute" button not visible, state of the button not perceptible

https://w3c.github.io/aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
If there is an element at the bottom of the browser, the list entries are not displayed.

https://w3c.github.io/aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-both.html and the two other aria 1.0 combobox pattern
IE 11: List entries not visible

https://w3c.github.io/aria-practices/examples/dialog-modal/datepicker-dialog.html
Does not work with IE 11 
Contrast date picker icon: 2.3:1

https://w3c.github.io/aria-practices/examples/disclosure/disclosure-img-long-description.html, 
https://w3c.github.io/aria-practices/examples/disclosure/disclosure-faq.html,
https://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html,
https://w3c.github.io/aria-practices/examples/menubar/menubar-2/menubar-2.html,
https://w3c.github.io/aria-practices/examples/treeview/treeview-2/treeview-2b.html
transparent image for status: not or poorly visible in high contrast mode

https://w3c.github.io/aria-practices/examples/grid/LayoutGrids.html
Example 2: Does not work with IE 11 
Example 2: Contrast delete icon: 1.6:1

https://w3c.github.io/aria-practices/examples/grid/dataGrids.html
Example 2: sort icon not correct visible in high contrast mode
Example 2: pen icon not visible in high contrast mode (background image)

http://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html
Example 1: button "Not Important" does not work with IE 11
Example 1: arrow icon not visible in high contrast mode (background image)
[...]

PR #1172, opened 2019-09-18
Datepicker Example: Increase colour contrast ratio to pass WCAG (#1170)

### LINKED ISSUE #1170: datepicker has insufficient contrast
Consider bumping the icon color up to past 3:1 and the header and footer blue background to 4.5:1

image

I suggest #008EA2 for the icon
and #005793 for the header and footer background.

They are fairly random value suggestions, but they at least pass.

ZoeBijl:
Corrected the colour contrast of various UI elements so that they pass WCAG minimums.

Resolves #1170

### DIFF
--- a/examples/dialog-modal/css/datepicker.css
+++ b/examples/dialog-modal/css/datepicker.css
@@ -17,7 +17,7 @@
 
 .datepicker button.icon:focus {
   outline: none;
-  border-color: hsl(216, 80%, 55%);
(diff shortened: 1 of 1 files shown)


ZoeBijl:
@smhigley I moved the high contrast issues to #1132. More specifically, [this comment in that thread](https://github.com/w3c/aria-practices/issues/1132#issuecomment-534418273). Could you file a separate issue for the date/month switching thing?

Issue #1105, opened 2019-07-19
Develop example of the meter


mcking65:
Develop an example of an implementation of the meter pattern using the new ARIA 1.2 meter role. It would be helpful if it illustrated a scenario where HTML meter is insufficient.

Related to #979.

a11ydoer:
Example can be found here. 
https://rawcdn.githack.com/w3c/aria-practices/1f7d940ca9872ff2f240c65ce4520338fbdbd3c6/examples/meter/meter.html

carmacleod:
Question: meter isn't keyboard interactive, so maybe it shouldn't have `tabindex="0"`?

carmacleod:
Should probably have a visible label instead of using `aria-label`?

scottaohara:
suggestions:
the meter has an aria-label of "disk usage" but as there's no way to infer that from the visual alone it'd be better to have visible text and use aria-labelledby.

changing the color of the meter as it approached max value / capacity would be another way to distinguish this from a progress bar.

PR #1189, opened 2019-09-30
Add meter design pattern

### LINKED ISSUE #979: Draft meter pattern
The new meter role has landed in ARIA master branch. Guidance please and thank you. :)

smhigley:
Initial draft of a meter pattern (the example implementation is tracked separately in #1105).

Resolves #979

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1189.html" title="Last updated on Nov 5, 2019, 8:10 PM UTC (9525fea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1189/27e1721...9525fea.html" title="Last updated on Nov 5, 2019, 8:10 PM UTC (9525fea)">Diff</a>

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -2150,6 +2150,37 @@
       </section>
     </section>
 
+    <section class="widget" id="meter">
+      <h3>Meter</h3>
+      <p>
+        An <a class="role-reference" href="#meter">meter</a> is an element that allows authors to define a scalar value within known minimum and maximum limits.
+      </p>
+      <p class="note">
+        The <code>meter</code> widget should not be used to indicate progress, such as loading or percent completion of a task. For that, use the <a href="#progressbar" class="role-reference">progressbar</a> role instead.
+      </p>
+
+      <!-- <section class="notoc">
+        <h4>Example</h4>
+        <p>TODO: add example link</p>
+      </section> -->
+
+      <section class="notoc">
+        <h4>Keyboard Interaction</h4>
+        <p>A meter does not require any keyboard interaction.</p>
+      </section>
+
+      <section class="notoc">
+        <h4>WAI-ARIA Roles, States, and Properties</h4>
+        <ul>
+          <li>The widget has a role of <a class="role-reference" href="#meter">meter</a>.</li>
+          <li>The meter element has <a class="property-reference" href="#aria-valuenow">aria-valuenow</a> set to a decimal value between <code>aria-valuemin</code> and <code>aria-valuemax</code> representing the current value of the meter.</li>
+          <li>The meter element has <a class="property-reference" href="#aria-valuemin">aria-valuemin</a> set to a decimal value less than or equal to <code>aria-valuemax</code>.</li>
+          <li>The meter element has <a class="property-reference" href="#aria-valuemax">aria-valuemax</a> set to a decimal value greater than or equal to <code>aria-valuemin</code>.</li>
+          <li>If the meter has a visible label, it is referenced by <a href="#aria-labelledby" class="property-reference">aria-labelledby</a> on the meter element. Otherwise, the meter element has a label provided by <a href="#aria-label" class="property-reference">aria-label</a>.</li>
+        </ul>
+      </section>
+    </section>
+
     <section class="widget" id="radiobutton">
       <h3>Radio Group</h3>
       <p>
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -2156,7 +2156,7 @@
         A <a class="role-reference" href="#meter">meter</a> allows authors to define a value within a known range. The minimum and maximum limits of the range must be defined numerically, and the value should be a number within the range.
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


aria-practices.html
```diff
@@ -2150,6 +2150,37 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
       </section>
     </section>
 
+    <section class="widget" id="meter">
+      <h3>Meter</h3>
+      <p>
+        An <a class="role-reference" href="#meter">meter</a> is an element that allows authors to define a scalar value within known minimum and maximum limits.
```
carmacleod: Change "An meter" to "A meter"

aria-practices.html
```diff
@@ -2150,6 +2150,37 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
+      <section class="notoc">
+        <h4>Keyboard Interaction</h4>
+        <p>A meter does not require any keyboard interaction.</p>
+      </section>
+
+      <section class="notoc">
+        <h4>WAI-ARIA Roles, States, and Properties</h4>
+        <ul>
+          <li>The widget has a role of <a class="role-reference" href="#meter">meter</a>.</li>
+          <li>The meter element has <a class="property-reference" href="#aria-valuenow">aria-valuenow</a> set to a decimal value between <code>aria-valuemin</code> and <code>aria-valuemax</code> representing the current value of the meter.</li>
```
carmacleod: Not sure we should say "the meter element", just in case authors think we mean "the html meter element"?
In other patterns, we have said either:
- "the element with role meter", or
- "the meter"

(5 instances)

smhigley: I'm happy to change this, though if it's a general guidance it might be good to revisit some of the other pattern wording, e.g. `radiobutton`, `slider`, `spinbutton`, and `menu` (there are probably more, I didn't look through comprehensively. Those are the ones I looked at when writing meter)

aria-practices.html
```diff
@@ -2150,6 +2150,37 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
+
+      <section class="notoc">
+        <h4>Keyboard Interaction</h4>
+        <p>A meter does not require any keyboard interaction.</p>
+      </section>
+
+      <section class="notoc">
+        <h4>WAI-ARIA Roles, States, and Properties</h4>
+        <ul>
+          <li>The widget has a role of <a class="role-reference" href="#meter">meter</a>.</li>
```
carmacleod: Not a widget? (because not interactive)

aria-practices.html
```diff
@@ -294,7 +294,7 @@ <h4>Example</h4>
 
       <section class="notoc">
         <h4>Keyboard Interaction</h4>
-        <p>An alert (WAI-ARIA live region) does not require any keyboard interaction.</p>
+        <p>Not applicable.</p>
```
ZoeBijl: 👍

aria-practices.html
```diff
@@ -2150,6 +2150,43 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
       </section>
     </section>
 
+    <section class="widget" id="meter">
+      <h3>Meter</h3>
+      <p>
+        A <a class="role-reference" href="#meter">meter</a> allows authors to define a value within a known range. The minimum and maximum limits of the range must be defined numerically, and the value should be a number within the range.
```
ZoeBijl: I really like this definition, nice work!

aria-practices.html
```diff
@@ -2150,6 +2150,43 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
+      </p>
+      <p>
+        Meter examples could include current battery percentage, or a car's fuel level. A <code>meter</code> could not be used to represent a value like the current world population, since there is no meaningful maximum limit.
```
ZoeBijl: > Meter examples could include current battery percentage, or a car's fuel level

Propose to change this to:

> Meter example include a device’s current batter percentage and a car’s fuel level.

---

> A <code>meter</code> could not be used to

Propose we change to either:

> A <code>meter</code> cannot be used to…

…or…

> A <code>meter</code> should not be used to…

I think the first one is a little more strict and to the point. But the second one reads a little bit friendlier (and is more consistent with other text). Thoughts?

aria-practices.html
smhigley: 😊

aria-practices.html
smhigley: I like should, friendly sounds like a good approach :)

smhigley:
@mcking65 thanks for all your suggestions! I updated the PR to include them 👍

mcking65:
carmacleod wrote:

> Looks good. Should we consider removing "or equal to" from valuemin and valuemax?
> i.e. if a meter goes from 0 to 0 and has a value of 0, would AT say 0% or 100%? :)

Per discussion in today's meeting, changed this in commit 9525fea.

Issue #1180, opened 2019-09-21
Draft guidance on code readability


ZoeBijl:
Our code should be as easy to read as possible. People should be able to jump into it and understand it straight away. To achieve this we’ll need simpler code. And to achieve that we’ll need guidance on code readability that we can point to.

---

## Proposal topics

This is a list of topics/rules around which I want to draft this guidance. Will elaborate/explain these later. This is just a brain dump at the moment.

- No unnecessary comments for documentation tools or code editors
- No unnecessary comments at all
- No explaining how something works, this is not a tutorial
- Fine to have comments that explain what a line of code does (ex: trigger collapse action when return is pressed)
- easy function notation, none of this dot notation stuff
- HTML should use as few classes as possible
- HTML should use as few ARIA as possible (meaning, use a `button`-element if usage of the `button`-role is otherwise irrelevant to the example)
- CSS should be easy to follow meaning there should be some logic behind the order of declaration groups
- Don’t need to enforce an order of CSS properties as that’s very arbitrary
- File names should be easy to read too and ideally there should only be one for each type per example
- Where possible differences in implementation should be done via attributes/settings rather than creating an entirely new JS file
- Be clear, not clever (if five lines are clearer than one line of clever code, go for the clear five line version)
- No “more” and “less” accessible version of a code example. We’re the APG, we should only have accessible stuff.
- Keep your code example in one JS file. No one wants to figure out which of the five provided files contains the bit they’re looking for.
- Dash notation for CSS classes: `component-part` not `componentPart` or `component_part`

## Proposal guidance

Written out text for the above topics.

### HTML

If you watch Effortless Style by Heydon and read my Cool Semantic Styles article and apply what is said in those you should be pretty close to what we want.

#### Resources
- [Effortless Style](https://vimeo.com/101718785) by Heydon Pickering ([slides](https://slides.com/heydon/effortless-style))
- [Cool Semantic Styles](https://codepen.io/Moiety/full/zvebZO) by Zoë Bijl

### CSS

tbw

### JS

tbw

mcking65:
@ZoeBijl, please add any new guidance to this wiki page:

https://github.com/w3c/aria-practices/wiki/Code-Guide

ZoeBijl:
@mcking65 I will add it to those wiki pages. But I’ll first draft things in this thread so they can be more easily discussed.

jongund:
@ZoeBijl 

[...]

PR #1190, opened 2019-10-01
Add meter example for issue 1105

### LINKED ISSUE #1105: Develop example of the meter

a11ydoer:
Resolves #1105

initial draft of [meter example](https://raw.githack.com/w3c/aria-practices/issue979-add-meter-example/examples/meter/meter.html).

### Review checklist

* [x] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @mcking65 @carmacleod 
* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @smhigley @sh0ji @carmacleod 
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @ZoeBijl 
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @smhigley @ZoeBijl 

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1190.html" title="Last updated on Oct 8, 2019, 7:11 PM UTC (0292f55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1190/fe9a10f...0292f55.html" title="Last updated on Oct 8, 2019, 7:11 PM UTC (0292f55)">Diff</a>


examples/meter/meter.html
```diff
@@ -0,0 +1,169 @@
...
+          <td>
+            <ul>
+              <li>Identifies the element as a <code>meter</code> widget.</li>
+            </ul>
+          </td>
+        </tr>
+        <tr data-test-id="meter-tabindex">
+            <td></td>
+            <th scope="row">
+              <code>tabindex=&quot;0&quot;</code>
```
carmacleod: Question: meter isn't keyboard interactive, so maybe it shouldn't have tabindex="0"?

examples/meter/meter.html
```diff
@@ -0,0 +1,169 @@
...
+  The following meter examples demonstrate the <a href="">meter design pattern</a>.
+
+  <section>
+    <h2 id="ex_label">Example</h2>
+
+    <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
+
+    <div id="example">
+      <p>This meter changes its value randomly every 5 seconds.</p>
+      <div role="meter" tabindex="0" aria-valuenow="90" aria-valuemin="0" aria-valuemax="128"
```
carmacleod: Question: meter isn't keyboard interactive, so maybe it shouldn't have tabindex="0"?

examples/meter/meter.html
```diff
@@ -0,0 +1,169 @@
...
+        aria-label="Disk Usage">
```
carmacleod: Should probably have a visible label instead of using aria-label?

examples/meter/meter.html
```diff
@@ -0,0 +1,169 @@
...
+                <td><code>div</code></td>
+                <td>
+                  <ul>
+                    <li>A numeric value that is the current value of the meter</li>
+                  </ul>
+                </td>
+              </tr>
+              <tr data-test-id="meter-aria-label">
+                  <td></td>
+                  <th scope="row"><code>aria-label</code></th>
```
carmacleod: Should probably have a visible label with `aria-labelledby` instead of using `aria-label`?

examples/meter/meter.html
```diff
@@ -0,0 +1,169 @@
...
+  <ul>
+    <li><a href="../../../#browser_and_AT_support">Browser and Assistive Technology Support</a></li>
+    <li><a href="https://github.com/w3c/aria-practices/issues/new">Report Issue</a></li>
+    <li><a href="https://github.com/w3c/aria-practices/projects/10">Related Issues</a></li>
+    <li><a href="../../../#meter">Design Pattern</a></li>
+  </ul>
+</nav>
+<main>
+  <h1> Meter Example</h1>
+  The following meter examples demonstrate the <a href="">meter design pattern</a>.
```
carmacleod: Will there be more than one meter example?
If not, then need to change "The following meter examples demonstrate..."
to "The following meter example demonstrates..."

examples/meter/meter.html
```diff
@@ -0,0 +1,169 @@
...
+        aria-label="CPU Usage">
+        <div id="fill"></div>
+      </div>
+    </div>
+    <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
+  </section>
+  <section>
+      <h2 id="kbd_label">Keyboard Support</h2>
+
+      <p>It would not have any keyboard support since this is system generated information.</p>
```
carmacleod: Consider changing the Keyboard Support to "Not applicable.", similar to Breadcrumb pattern.

examples/meter/meter.html
a11ydoer: fixed @carmacleod

examples/meter/meter.html
a11ydoer: done. @carmacleod

examples/meter/meter.js
```diff
@@ -0,0 +1,32 @@
+/** @type {!Element} Meter element. */
```
ZoeBijl: Can we remove the @type cruft?

examples/meter/meter.html
a11ydoer: done @carmacleod

examples/meter/meter.html
a11ydoer: done @carmacleod

aria-practices.html
```diff
@@ -2151,6 +2151,43 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
       </section>
     </section>
 
+    <section class="widget" id="meter">
```
smhigley: This section will conflict with #1189, but I can add anything from here to that PR if desired, or we can figure out another way to merge them

examples/meter/meter.html
```diff
@@ -0,0 +1,155 @@
...
+  <h1> Meter Example</h1>
+  The following meter example demonstrates the <a href="">meter design pattern</a>.
+
+  <section>
+    <h2 id="ex_label">Example</h2>
+
+    <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
+
+    <div id="example">
+      <p>This meter changes its value randomly every 5 seconds.</p>
```
smhigley: Would it make sense to have text here that describes the purpose of the meter, e.g. "A meter showing percent CPU usage"?

examples/meter/meter.html
```diff
@@ -0,0 +1,155 @@
...
+      <div role="meter" tabindex="0" aria-valuenow="90" aria-valuemin="0" aria-valuemax="128"
```
smhigley: Should the max for CPU usage be 100%?

examples/meter/meter.js
```diff
@@ -0,0 +1,32 @@
...
+let meterEl;
+/** @type {!Element} Element visually filling the meter. */
+let visualFillEl;
+/** @type {number} Minimum value. */
+let minUsage;
+/** @type {number} Maximum value. */
+let maxUsage;
+/** @type {number} Current value. */
+let usage;
```
smhigley: Are we using `let` and `const` in examples? Also, perhaps these should be locally defined within the functions below rather than defined globally.

examples/meter/meter.js
```diff
@@ -0,0 +1,32 @@
...
+  const range = maxUsage - minUsage;
+  usage = Math.floor((Math.random() * range) + minUsage);
+  meterEl.setAttribute('aria-valuenow', usage);
+  render();
+}
+function render() {
+  const asPercentage = (100 * usage) / (maxUsage - minUsage);
+  visualFillEl.style.width = asPercentage + '%';
+}
+function init() {
```
smhigley: Since these are in the global namespace, would it make sense to define them as `initMeter`, `renderMeter`, and `randomizeMeterUsage`, just to be slightly safer? Another option would be to define everything within a `Meter` constructor function.

draft-labelling-describing.md
```diff
@@ -0,0 +1,308 @@
+﻿Labeling and Describing
```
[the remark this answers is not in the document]
carmacleod: good catch, @ZoeBijl 
maybe this should just be deleted? it seems to be some old markdown for a draft of the labeling and describing section.

examples/meter/meter.js
```diff
@@ -0,0 +1,32 @@
...
+  meterEl = document.querySelector('[role=meter]');
```
smhigley: Would it be worth it to use `document.querySelectorAll('[role=meter]')`, then loop through and initiate them, to allow for the possibility of multiple meters per page?

examples/meter/meter.html
ZoeBijl: That makes more sense to me.

examples/meter/meter.js
ZoeBijl: I think our eslint prevents us from doing so (which I have thoughts about). But not sure if this would impact our build.

smhigley: I'd definitely like to start using es6 syntax and update eslint (lmk if you want help with that), though maybe in a separate PR :D

examples/meter/meter.html
a11ydoer: done @carmacleod

examples/meter/meter.js
a11ydoer: @smhigley,
@ZoeBijl @mcking65 and I are still in the proces of setting up "code review" workflow after the feedback from TPAC meeting in Japan. Thanks for your patience.

examples/meter/meter.html
a11ydoer: @smhigley, change was made and it will be comitted with other changes. thanks for the catch. (I am glad this is not temperature meter because we don't know what will be the max value due to continous global warming. ;-)

draft-labelling-describing.md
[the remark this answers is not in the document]
a11ydoer: @ZoeBijl @carmacleod, 
@mcking65 and I decide to correct this PR, rather than making the separate PR. Thanks for the feedback.

examples/meter/meter.js
sh0ji: > Another option would be to define everything within a `Meter` constructor function.

I'd favor this since it will make the migration to ES6 classes easier down the road.

examples/meter/meter.js
sh0ji: That's a great practice for writing libraries where usage isn't controlled by the author, but we also control the content on this page so I don't think it's exactly necessary.

If we ever decide to add another meter to the page, the `init` method can be updated as part of that PR.

a11ydoer: Thanks for your feedback @smhigley and @sh0ji

examples/meter/meter.js
sh0ji: I'm 100% in agreement about moving on to ES6, but I think we should resolve #1180 before making any changes to code style. All the `let`/`const` keywords should just be replaced with `var` to land this.

ZoeBijl: What’s the reason we don’t allow `let` & `const`?

nschonni: Code is ES5 for IE11 compat AFAIK. Although IE11 mostly supports those with a caveat on `let` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

ZoeBijl: Thanks @nschonni. We can discuss that as part of our support discussion. I think we were all in favour of dropping IE11 support. But no decision has been made yet.

examples/meter/meter.js
smhigley: I'd agree with this, if the only issue were `document.querySelector` vs `querySelectorAll` + a loop to call an init function:
> That's a great practice for writing libraries where usage isn't controlled by the author, but we also control the content on this page so I don't think it's exactly necessary.

My main concern is actually that the state for the meter component is being stored in one-off global variables, so if more than one meter were needed, all the logic would need to be rewritten. If we move away from this to a constructor function or class-like approach, then I wouldn't have an issue with not looping and `init`-ing multiple possible meters.

sh0ji: Yeah, you're right. I meant to convey that this isn't breaking anything now, so I don't think it should block merging. @a11ydoer is of course more than welcome to change it now.

Keep in mind that we'll need to do a large-scale refactor later if we switch to ES6, so that would also be an appropriate time to add the loop, as well as just skipping the constructor function and going straight to a class.

draft-labelling-describing.md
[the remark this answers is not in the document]
mcking65: I removed draft-labelling-describing.md.

examples/meter/meter.html
```diff
@@ -0,0 +1,156 @@
...
+  The following meter example demonstrates the <a href="../../#meter">meter design pattern</a>.
+
+  <section>
+    <h2 id="ex_label">Example</h2>
+
+    <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
+
+    <div id="example">
+      <p>This meter changes its value randomly every 5 seconds.</p>
+      <h3 id="cpu_usage_label">Central Processing Unit(CPU) Usage</h3>
```
carmacleod: Need a space between "Central Processing Unit" and "(CPU)".

examples/meter/meter.html
```diff
@@ -0,0 +1,156 @@
...
+                  </ul>
+                </td>
+              </tr>
+              <tr data-test-id="meter-aria-labelledby">
+                  <td></td>
+                  <th scope="row"><code>aria-labelledby</code></th>
+                  <td><code>div</code></td>
+                  <td>
+                    <ul>
+                      <li>Identifies the accessible name for <code>meter</code> widget.</li>
```
carmacleod: Add "the", i.e. accessible name for the meter widget.

aria-practices.html
a11ydoer: @smhigley it seems like #1189 for design pattern, not the example. Did I misunderstand?

smhigley: It looks like the changes in this file have already been removed from the PR 👍

a11ydoer:
@scottaohara had a feedback of "changing the color of the meter as it approached max value / capacity would be another way to distinguish this from a progress bar." at issue 1105

examples/meter/meter.html
ZoeBijl: Fixed in 93cbf98

examples/meter/meter.html
ZoeBijl: Fixed in 9bcfe10

ZoeBijl:
I’ve made some visual changes (in b2df353) to make it a bit more modern looking:
The meter with rounded corners and slightly thinner borders
The value bar wasn’t showing up in Windows High Contrast Mode. This is because it uses a background colour. I changed it to use a border instead but am interested in different solutions if available.

smhigley:
I updated the code to be more easily reusable, and added a pause button and color fill that changes with the width.

@ZoeBijl I modified your styles to have a border just around the outside to allow for a color bg. Another option might be to use an SVG fill instead, and modify the `fill` color, which should still show up in HCM. Let me know what you think of the current style, or if it's worth changing (and if you like the SVG approach)!

meter-capture

examples/meter/meter.html
```diff
@@ -35,12 +35,12 @@ <h2 id="ex_label">Example</h2>
...
+      <p>This meter changes its value randomly every 5 seconds, with an option to pause changes.</p>
       <h3 id="cpu_usage_label">Central Processing Unit (CPU) Usage</h3>
-      <div role="meter" aria-valuenow="90" aria-valuemin="0" aria-valuemax="100"
-        aria-labelledby ="cpu_usage_label">
-        <div id="fill"></div>
-      </div>
+      <p>
+        <button type="button" class="play-meters">Pause Updates</button>
+      </p>
+      <div role="meter" aria-valuenow="90" aria-valuemin="0" aria-valuemax="100" aria-labelledby ="cpu_usage_label"></div>
```
ZoeBijl: Is there a space between the `aria-labelledby` attribute and its `=` sign?

examples/meter/css/meter.css
```diff
@@ -7,9 +7,10 @@
...
 }
 
-#fill {
+.fill {
   box-sizing: border-box;
   height: 100%;
   border-radius: 3px;
-  border: 20px solid black;
+  background-color: black;
+  outline: 3px solid black;
```
ZoeBijl: Out of interest: is there a reason to use `outline` rather than `border` here?

ZoeBijl:
Thank you for your changes @smhigley, this looks great! I think the SVG would be better if that means we can retain the fill. To me it feels wrong to only show the two shapes with borders. Feels like that takes away some clarity of what this is.

What do you think?

examples/meter/meter.html
smhigley: I think this should have been fixed in the last commit

examples/meter/css/meter.css
smhigley: I think I switched it back to border, and reduced it to 2px

carmacleod:
Looks good, @smhigley !

I agree with @ZoeBijl that the meter would be "more clearly a meter" if we could remove the inner border. (using SVG sounds fine - but curious why we wouldn't just change the border color in js at the same time as the fill color? Sorry if that's a CSS noob question! 😄 )

Also just wondering whether having the Pause button after the meter would feel more natural?
i.e. a screen reader user navigating sequentially would encounter the Pause button and might wonder what it was for? And it might look more visually coherent to have the meter right next to its label, instead of having a Pause button in between? Would love to have other opinions on this thought.

smhigley:
@carmacleod putting the pause button after the meter makes sense to me! The reason I have a border on there is because not all colors between green and red pass 3:1 contrast (particularly in the yellow range).

@mcking65 / @spectranaut independent of what we do with the fill and pause button position, the tests should be ready for review! I remember Matt saying something about wanting to get it done on Friday on the call :)

carmacleod:
> The reason I have a border on there is because not all colors between green and red pass 3:1 contrast (particularly in the yellow range).

Ah, of course. I misunderstood what @ZoeBijl said earlier, but I get it now. :)

smhigley:
@carmacleod I don't think you misread Zoë's comment, if you mean the one from before I added my changes. Using just `background` doesn't work for Windows HCM, and if that were the only issue then your suggestion would work. But then I added a changing fill bg color so now there's the additional contrast issue :).

carmacleod:
@smhigley Actually, I meant this comment:
> I think the SVG would be better if that means we can retain the fill. To me it feels wrong to only show the two shapes with borders. Feels like that takes away some clarity of what this is.

I misread the 2nd sentence as:
> To me it feels wrong to ~~only~~ show the two shapes with borders.

smhigley:
The meter fill has been updated to use an SVG, and shows up in Windows HCM like this:

screen shot of meter in high contrast mode, with a solid green fill and green border

PR #1200, opened 2019-10-07
Accordion Example: Correct color contrast (Issue #1132)

ZoeBijl:
Resolves a part of https://github.com/w3c/aria-practices/issues/1132

### DIFF
--- a/examples/accordion/css/accordion.css
+++ b/examples/accordion/css/accordion.css
@@ -12,7 +12,7 @@
 }
 
 .Accordion.focus {
-  border-color: hsl(216, 94%, 73%);
+  border-color: hsl(216, 94%, 43%);
 }
 
 .Accordion.focus h3 {
@@ -42,6 +42,10 @@
   background: hsl(216, 94%, 94%);
 }
 
+.Accordion-trigger:focus {
+  outline: 4px solid transparent;
+}
+
 .Accordion *:first-child .Accordion-trigger {
   border-radius: 5px 5px 0 0;
 }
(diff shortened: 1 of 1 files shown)


ZoeBijl:
Thanks for the comment and feedback! I’ll change the two (high) contrast things you mentioned. 

I kinda want to redo the example at some point. So I’ll leave the Firefox specific thing for another time/person.

smhigley:
@sh0ji I think you might be reading the text of the requirement correctly, though in this case it seems like a grey area because the divide between the button and the accordion content could be meaningful.

Even if the button borders aren't a technical 1.4.11 failure, it still seems like better practice to meet at least 3:1. My beyond-AA rule of thumb is to check how usable something is if the borders were removed entirely. In this case it doesn't seem all that wonderful a user experience, so as long as color contrast is being updated, might as well update it for everything.

+1 to the input borders getting better contrast too :)

sh0ji:
Great point, @smhigley. I don't have any qualms with going above and beyond AA of course (seems like there should be a AAA for this) and am not recommending we change this back. I just want to make sure I understand the requirement well so that we know when to address this type of thing in the future.

I had a similar thought about the button borders in the context of the accordion, but the fact that [the details/summary elements](https://codepen.io/sh0ji/pen/ZEEbByB) don't have a border in any browser made me question that. I do think it's good UX to include a visual indicator of the boundaries of clickable regions when it extends beyond the text box. There's a WCAG SC for [target size](https://www.w3.org/TR/WCAG21/#target-size), but not one for target perceivability, so that doesn't seem like an accessibility requirement.

ZoeBijl:
@smhigley and @sh0ji can I get your approval (or request for changes) on this one?

PR #1214, opened 2019-10-15
Modify log and timer name advice

### LINKED ISSUE #1201: Change naming and describing required for timer and log
Once https://github.com/w3c/aria/pull/1065 is merged then https://www.w3.org/TR/wai-aria-practices-1.1/#naming_role_guidance should be modified.

- I suggest `log` is modeled after `alert` as they appear similar
- for `timer` I really have no idea how a screen reader would ever read the accessible name... Anyone have any ideas as to what should be written here?

jnurthen:
Modify log and timer naming advice based on them no longer being required
Closes #1201

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -4462,8 +4462,12 @@ So, As advised by <a href="#naming_rule_avoid_fallback">Rule 4: Avoid Browser Fa
             </tr>
             <tr>
               <td><a href="#log" class="role-reference"><code>log</code></a></td>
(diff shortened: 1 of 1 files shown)


jnurthen:
> I must have missed any discussion on why we are not mentioning aria-labelledby if there's a visible label?

Not sure there was any discussion. IMO that is a reasonable conversation to have in a new issue which should also include discussion of status and alert which have identical advice.

Issue #1225, opened 2019-10-24
Reversion title and footer on example pages


mcking65:
The string "WAI-ARIA Authoring Practices 1.1" needs to be changed to "WAI-ARIA Authoring Practices 1.2" in 2 places on every example page in the master branch:
1. In the title tag.
2. In the link in the nav element at the bottom of the page.

spectranaut:
Should the links to 1.1 in the landmark example pages be updated as well? The links are all explicitly to aria1.1 instead of the editors draft, so they won't be incorrect.

[banner landmark page](https://w3c.github.io/aria-practices/examples/landmarks/banner.html)

PR #1226, opened 2019-10-25
Update title and footers in example pages to 1.2 (issue #1225)

spectranaut:
To explain the several commits:

1. In order to be able to change all of the references in one `sed`, I fixed some example file's formatting for the title and footers in separate commits for disclosure example pages and combobox example pages.

2. There were a few relative links with link text that said "WAI-ARIA Authoring Practices 1.1", so I removed the "1.1" in order for the links to not be wrong when looking at the editors draft. This is in the "toolbar" and "breadcrumbs" commits.

3. I also updated the reference table template, but examples/index.html is still out of date.

### DIFF
--- a/examples/coding-template/Depricated-MultipleImplementationExample-Template.html
+++ b/examples/coding-template/Depricated-MultipleImplementationExample-Template.html
@@ -20,7 +20,7 @@
   The name should be unique if the pattern will have multiple implementations,
   e.g., "Simple Two-State Checkbox" or "Three-State Checkbox".
   -->
-<title>EXAMPLE_NAME | WAI-ARIA Authoring Practices 1.1</title>
+<title>EXAMPLE_NAME | WAI-ARIA Authoring Practices 1.2</title>
 
 <!--  Core js and css shared by all examples; do not modify when using this template. -->
 <link href="../css/core.css" rel="stylesheet">
@@ -297,7 +297,7 @@
   </main>
   <nav>
     <!--  Update the pattern_ID parameter of this link to refer to the APG design pattern section related to this example.  -->
-    <a href="../../#pattern_ID">EXAMPLE_NAME Design Pattern in WAI-ARIA Authoring Practices 1.1</a>
+    <a href="../../#pattern_ID">EXAMPLE_NAME Design Pattern in WAI-ARIA Authoring Practices 1.2</a>
   </nav>
 </body>
 </html>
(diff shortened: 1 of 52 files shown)


examples/toolbar/help.html
```diff
@@ -19,7 +19,7 @@
 
 <h1>Toolbar Help</h1>
 
-<p>Learn more about toolbar behaviors by reading the <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#toolbar">ARIA Authoring Practices Guide</a>.</p>
+<p>Learn more about toolbar behaviors by reading the <a href="https://w3c.github.io/aria-practices/#toolbar">ARIA Authoring Practices Guide</a>.</p>
```
ZoeBijl: Should we change this to a relative URL? Not sure why it wasn't to begin with tbh.

examples/combobox/aria1.0pattern/combobox-autocomplete-none.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <meta charset="utf-8" />
-<title>Legacy ARIA 1.0 Combobox without Autocomplete Example | WAI-ARIA Authoring Practices 1.0</title>
```
ZoeBijl: General note: why the heck did this one say 1.0 🙈 .

examples/coding-template/Depricated-MultipleImplementationExample-Template.html
```diff
@@ -20,7 +20,7 @@
   The name should be unique if the pattern will have multiple implementations,
   e.g., "Simple Two-State Checkbox" or "Three-State Checkbox".
   -->
-<title>EXAMPLE_NAME | WAI-ARIA Authoring Practices 1.1</title>
```
ZoeBijl: @mcking65 is there a reason why this is deprecated? And should we remove it?

examples/toolbar/help.html
spectranaut: good point! pushed a fix for that.

examples/coding-template/Depricated-MultipleImplementationExample-Template.html
mcking65: The group decided that we should prefer only one example per page. We could consider removing it now. A separate issue is breaking apart the pages that currently have multiple.

mcking65:
@ZoeBijl, thank you for the thorough review!

examples/coding-template/Depricated-MultipleImplementationExample-Template.html
ZoeBijl: Excellent, have made a new issue for that: https://github.com/w3c/aria-practices/issues/1235

Issue #1205, opened 2019-10-08
Review testing requirements in PR review process


smhigley:
The current test review section of the [PR Review Process](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test-review) doc has these requirements for AVA regression tests:

- Have appropriate data-test-id attributes on every row of the keyboard and attribute documentation tables in the HTML file.
- Are defined in a js file that is named correctly.
- Test the behavior as documented.
- Exercise boundary conditions.
- Run successfully.

This doesn't always cover all logic contained in example code. Should we consider including something like [CodeCov](https://codecov.io/) to require that a certain percent of lines are tested (or just display percent coverage in PRs)? Another possibility is adding unit tests to add coverage for some of the discrete chunks of logic and edge cases.

spectranaut:
As a simple solution, we could add an additional take data-test-id (like: "additional-behavior-tests") that will prevent `ariaTest` from confirming it's existence in the example page, then test authors can writes additional tests to their hearts content. This would just be a line of code.

spectranaut:
Example of test where we should use this data-test-id: https://github.com/w3c/aria-practices/pull/1191#discussion_r339118043

PR #1227, opened 2019-10-25
Allow for tests that do not refer to example tables rows

spectranaut:
@smhigley request something like this in https://github.com/w3c/aria-practices/issues/1205.

If you want to write regression tests for an APG example that cover behavior not specifically discussed in the attributes and keyboard interaction tables, you can use the data-test-id `test-additional-behavior`. 

If everyone is ok with it, we can merge this and I'll add it to the documentation!


spectranaut:
@mcking65 can you merge this? :)

ZoeBijl:
Matt is on vacation. But I’ll merge this as soon as the build is done.

Issue #1066, opened 2019-07-03
Combobox: Escape key should not clear current input


eldargab:
Currently ARIA 1.1 Combobox example says about ESC key that it:

* Clears the textbox
* If the listbox is displayed, closes it.

To my feeling textbox clearing contradicts almost everything users are used to and expect.

1. Escape key usually means "Abort what you are currently doing". From that point of view clearing is definitely an unwanted side effect.

2. Native browser select does not clear selection on escape key.

3. Every UI library I am aware of, which implements combobox pattern, does not clear textbox as well. [Ant design](https://ant.design/components/select/), [react-select](https://react-select.com/home), [select2](https://select2.org) to name a few and those are very popular.

### to-do list added by Matt

This will be fixed in APG 1.2 with the following:

- [x] Fixed in pattern with commit 315a4dd
- [x] Fixed in combobox-autocomplete-both.html with commit 85085b2
- [x] Fixed in combobox-autocomplete-list.html with commit 85085b2
- [x] Fixed in combobox-autocomplete-none.html with commit 85085b2
- [ ] Fix in grid-combo.html

smhigley:
I agree, I think this might also get in the way of 1.4.13/Content on hover or focus, since it's currently not possible to enter a value in the input without keeping the menu open.

mcking65:
Commit 315a4dd fixed this in the pattern but not in the examples.

mcking65:
Sarah, thank you for the fabulous review.

@smhigley wrote:

> * NVDA + Firefox: reads the page title four times almost every time I change options using the arrow key with the listbox open (this may be related to the fact that all option elements are mutated every time one changes selection; I'd be happy to investigate the root cause further). This is a code bug, not an issue with the pattern, as it did not repro in a different implementation.
> * JAWS + Firefox: would sporadically re-announce the listbox when arrowing through options. Seems likely related to the above

Opened #1264 and scheduled to resolve in next milestone.

> * IE: none of the listbox comboboxes work at all, with or without screen readers running

I didn't open an issue for this since we previously discussed dropping IE support in the new year.

> * IE: the grid combobox does not work with JAWS; options are not read when arrowed to.

Known bug in JAWS support for grids.

> * macOS + VoiceOver: the suggested "open" keyboard command does not work (Caps + Space), though Option + down arrow does
[...]

Issue #1026, opened 2019-05-01
Develop example of select-only combobox


carmacleod:
It would be useful to have an example of a readonly combobox in the APG.
The [AccDC simulated readonly combobox example](http://whatsock.com/tsg/Coding%20Arena/ARIA%20Comboboxes/ARIA%20Comboboxes%20(Simulated,%20Readonly)/demo.htm) works well, if @accdc is ok with us using that as a basis for an APG example. (Note that "simulated" just means it doesn't use a text input).

accdc:
Hi Carolyn,
Yes of course, you are welcome to use any of my code for any purpose you like. :) 

If you need any help breaking this out just let me know.

Kind regards,
Bryan

carmacleod:
Thanks, Bryan!
Much appreciated!
Carolyn

mcking65:
Issue #1330 has some good discussion of implementation details and links to useful examples.

Based on both that discussion as well as the language in the pattern, I am changing the title from "read-only" to "select-only". A read only combobox would be a combobox where the value cannot be changed.

carmacleod:
@mcking65 
> I am changing the title from "read-only" to "select-only". A read only combobox would be a combobox where the value cannot be changed.

I don't mean to hijack this issue, but I think an abundance of clarity is needed around the term "read-only" because combobox supports aria-readonly. Pain points are as follows:
- Windows platform has used the term "read-only" (since, like, forever) to describe what we are now calling a "select-only" combobox. Authors with a Windows background will need this new terminology explained in a big Note box.
- typically, applications use "disabled" (or aria-disabled) on a control to indicate that the value cannot be changed.
- so... what, exactly, does aria-readonly="true" mean on a combobox? Can the user invoke the popup? If so, can they select an item from the popup to change the combobox value? When you say "the value cannot be changed", does it simply mean _the value cannot be typed into the text field_? And if that is the case (i.e. if only the text part is read-only), then is there any difference between a select-only and aria-readonly combobox?
- how does a screen reader user know that they are in a select-only combobox? Would it be helpful to know that values cannot be typed into the combobox? Wouldn't that be indicated with aria-readonly?

Neither the [combobox spec](https://w3c.github.io/aria/#combobox), nor the [APG combobox pattern](https://w3c.github.io/aria/#combobox) adequately clarify this concept.

mcking65:
@carmacleod commented:

[...]

PR #1255, opened 2019-11-08
Combobox Pattern and Examples: Convert to support draft ARIA 1.2 specification

### LINKED ISSUE #1244: update combobox examples to use ARIA 1.2 pattern
Develop 4 examples that match the proposed ARIA 1.2 combobox specification by making the following changes to existing examples.
And, eliminate all 1.0 and 1.1 examples.

move the following 3 examples to the /combobox directory and convert to the 1.2 pattern and :
/combobox/aria1.0pattern/combobox-autocomplete-both.html
/combobox/aria1.0pattern/combobox-autocomplete-list.html
/combobox/aria1.0pattern/combobox-autocomplete-none.html
Then remove the aria1.0pattern subdirectory.

Move the following to the combobox directory and convert it to the 1.2 pattern:
/combobox/aria1.1pattern/grid-combo.html
Then remove the /aria1.1pattern/ subdirectory.

### LINKED ISSUE #1250: Revise combobox design pattern to reflect ARIA 1.2 requirements
The [combobox specification in ARIA 1.2](http://w3c.github.io/aria/#combobox)
includes several changes. Revise the [APG combobox design pattern](http://w3c.github.io/aria-practices/#combobox)
to reflect these changes.

- [x] Revise intro to reflect the change that the input does not have to be a text box; it could be an input element that does not support editable text.
- [x] Revise description of popup trigger conditions to reflect the change from textbox to input for some implementations
- [x] Revise description of use cases to reflect change from textbox to input
- [x] Revise description of autocomplete behaviors to reflect change from textbox to input
- [x] Revise keyboard section to have an input section, with separate lists of interactions depending on whether the input is editable
- [x] Revise states and properties section so there is no mention of versioning. There is just one list of requirements. At the end, add a note about the 1.0 pattern using aria-owns instead of aria-controls stating it is supported for legacy reasons but advises authors to remediate the implementations that use aria-owns to use aria-controls.
- [x] Throughout entire pattern, ensure all mentions of textbox are either replaced with input, appropriately conditional, or removed.
This page in the ARIA wiki may be a useful resource:
https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues

mcking65:
Resolves #1250, resolves #1244.

### Preview link

[Preview work in progress in the compare branch for this PR](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/aria-practices.html#combobox)

### Review checklist

* [x] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @carmacleod, @charmarkk, @jongund
* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @carmacleod, @smhigley 
* [x] [Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review) by @zcorpan - https://github.com/w3c/aria-practices/issues/1263
* [x] [Windows Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @carmacleod, @smhigley 
* [x] [Mac Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @smhigley 
* [ ] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @smhigley 

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1255.html" title="Last updated on Nov 14, 2019, 7:39 PM UTC (f535354)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1255/2b1c5b1...f535354.html" title="Last updated on Nov 14, 2019, 7:39 PM UTC (f535354)">Diff</a>


jongund:
@mcking65 Added the gird combobox example to the branch

jongund:
@mcking65 I updated the regression tests for the modified combobox examples for both, inline and none

mcking65:
@jongund, thank you, thank you for the updates!!! I finished the editorial updates to the grid combobox example page. So, now, except for the tests, all 4 examples are ready for review.

Do you have time to finish work on the tests?

Tomorrow, Simon will work on the pattern revisions, and then the entire PR will be ready for review.

examples/combobox/js/listbox.js
```diff
@@ -7,7 +7,7 @@ var Listbox = function (domNode, comboboxObj) {
     msgPrefix = 'Listbox constructor argument domNode ';
 
   // Check whether domNode is a DOM element
-  if (!(domNode instanceof Element)) {
+  if (!domNode instanceof Element) {
```
zcorpan:
```suggestion
  if (!(domNode instanceof Element)) {
```
This needs the inner parantheses, since `!` binds more strongly than `instanceof`. That is, the above is equivalent to

```
if ((!domNode) instanceof Element) {
```

examples/combobox/js/listboxOption.js
```diff
@@ -2,7 +2,7 @@
 *   This content is licensed according to the W3C Software License at
 *   https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
 */
-var ListboxOption = function (domNode, listboxObj) {
+var Option = function (domNode, listboxObj) {
```
zcorpan: `Option` is a standard constructor.

https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element

This would override the standard constructor, which works fine so long as there isn't other code that uses it.

Until we rewrite examples to use modules, at least, we should avoid global variables with the same names as standard constructors.

zcorpan: I noticed now that this also fails in the linter:
```
/home/travis/build/w3c/aria-practices/examples/combobox/js/listboxOption.js
  5:5  error  'Option' is already defined as a built-in global variable  no-redeclare
```

examples/combobox/js/listbox.js
zcorpan: I noticed now that this also fails in the linter:
```
/home/travis/build/w3c/aria-practices/examples/combobox/js/listbox.js
  10:7  error  Unexpected negating the left operand of 'instanceof' operator  no-unsafe-negation
```

aria-practices.html
```diff
@@ -712,31 +712,30 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
     <section class="widget" id="combobox">
       <h3>Combo Box</h3>
```
zcorpan: The spelling of combo box (2 words) vs combobox (1 word) is inconsistent. I prefer combobox.

aria-practices.html
```diff
@@ -712,31 +712,30 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
       <p>
-        A <a href="#combobox" class="role-reference">combobox</a> is a widget made up of the combination of two distinct elements:
-        1) a single-line textbox, and 2) an associated pop-up element for helping users set the value of the textbox.
+        A <a href="#combobox" class="role-reference">combobox</a> is a widget that controls a pop-up element for helping users set the value of the combobox.
```
zcorpan: The spelling of pop-up vs popup is inconsistent. I prefer popup (without a dash).

aria-practices.html
```diff
@@ -984,75 +975,55 @@ <h5>Tree Popup Keyboard Interaction</h5>
...
-                Note that elements with role <code>combobox</code> have an implicit <code>aria-haspopup</code> value of <code>listbox</code>.
-              </li>
-            </ul>
+          <li>The popup is an element that has role <a href="#listbox" class="role-reference">listbox</a>, <a href="#tree" class="role-reference">tree</a>, <a href="#grid" class="role-reference">grid</a>, or <a href="#dialog" class="role-reference">dialog</a>.</li>
+          <li>
+            If the popup has a role other than <code>listbox</code>, the element with role <code>combobox</code> has <a href="#aria-haspopup" class="property-reference">aria-haspopup</a> set to a value that corresponds to the popup type.
+            That is, <code>aria-haspopup</code> is set to <code>grid</code>, <code>tree</code>, or <code>dialog</code>.
+            Note that elements with role <code>combobox</code> have an implicit <code>aria-haspopup</code> value of <code>listbox</code>.
           </li>
+          <li>When the popup is visible, the combobox element has <a href="#aria-controls" class="property-reference">aria-controls</a> set to a value that refers to the popup element.</li>
```
zcorpan: Should we remove "When the popup is visible"? If I understand correctly, the popup element is always present in the DOM, and `aria-controls` should be present always. The popup is just styled with `display: none` when hidden.

aria-practices.html
```diff
@@ -984,75 +975,55 @@ <h5>Tree Popup Keyboard Interaction</h5>
...
+          <li>When a combobox receives focus, DOM focus is placed on the combobox element.</li>
+          <li>When a descendant of a listbox, grid, or tree popup is focused, DOM focus remains on the combobox and the combobox has <a href="#aria-activedescendant" class="property-reference">aria-activedescendant</a> set to a value that refers to the focused element within the popup.</li>
+          <li>For a combobox that controls a listbox, grid, or tree popup, when a suggested value is visually indicated as the currently selected value, the <code>option</code>, <code>gridcell</code>, <code>row</code>, or <code>treeitem</code> containing that value has <a href="#aria-selected" class="state-reference">aria-selected</a> set to <code>true</code>.</li>
           <li>
             If the combobox has a visible label, the element with role combobox has <a href="#aria-labelledby" class="property-reference">aria-labelledby</a> set to a value that refers to the labelling element.
```
zcorpan: The combobox can be an HTML element (`input`) that can be labeled with the HTML `label` element. Should we say that `label` element is preferred?

aria-practices.html
```diff
@@ -771,39 +772,31 @@ <h3>Combo Box</h3>
...
-      <p class="note">
-        The options for a combobox to popup a grid, tree, or dialog were introduced in ARIA 1.1.
-        Changes made in the ARIA 1.1 specification also add support for a code pattern that enables assistive technologies to present the textbox and popup as separately perceivable elements.
-        both ARIA 1.0 and 1.1 patterns are described in the following sections.
-        While using the ARIA 1.1 pattern is recommended as soon as assistive technology support is sufficient,
-        there are no plans to deprecate the ARIA 1.0 pattern.
-      </p>
 
       <section class="notoc">
         <h4>Examples</h4>
```
zcorpan: There is currently no example of a combobox that is not editable. We should add such an example.

zcorpan:
@mcking65 and @jongund I would appreciate your review on my changes. Thanks!

Also see https://github.com/w3c/aria-practices/issues/1250#issuecomment-552560140

aria-practices.html
zcorpan: Changed in https://github.com/w3c/aria-practices/pull/1255/commits/f82e215c70ec084dde67eb3c60eef4e34bd09415

aria-practices.html
zcorpan: Changed in https://github.com/w3c/aria-practices/pull/1255/commits/621b26c6c89c79c5219ae061b2736d0f87df8855

aria-practices.html
zcorpan: Changed in https://github.com/w3c/aria-practices/pull/1255/commits/2b926c8cd16970ac8f438f6b4c3380ec7ba08917

zcorpan:
We should update the tests as well.

examples/combobox/js/listboxOption.js
zcorpan: Fixed in https://github.com/w3c/aria-practices/pull/1255/commits/29e94eee65a532be3dcbbf0e4c9bad5083584454

examples/combobox/js/listbox.js
zcorpan: Fixed in https://github.com/w3c/aria-practices/pull/1255/commits/29e94eee65a532be3dcbbf0e4c9bad5083584454

aria-practices.html
zcorpan: Changed in https://github.com/w3c/aria-practices/pull/1255/commits/e6e8499495b3084fa3d131c88ebca414e0164eda

jongund:
@mcking65 Why did you change the test ids grid-combo.html?  Everything was working before the changes.  I work on fixing the test cases to use the new test ids.

mcking65:
@jongund asked:

> @mcking65 Why did you change the test ids grid-combo.html? Everything was working before the changes. I work on fixing the test cases to use the new test ids.

Jon, I'm sorry ... I didn't mean to screw that up. I thought I made only changes where the test ID didn't match what needed to be tested. Technically, since we are using a text input, the ones that include textbox in the ID are fine, and I could have left that part of the ID alone.

aria-practices.html
mcking65: The actual requirement is aria-controls only needs to be present when the popup is visible. There is discussion about what browsers should expose when aria-controls refers to something not present or not visible. That said, it is not an error to refer to an element that is not visible.

jongund:
@mcking65 I updated the grid combobox regression test files

jongund:
@mcking65 

In the ARIA practices section for combobox "WAI-ARIA Roles, States, and Properties"

Item 1
> If the combobox element is editable, it has a value for aria-multiline of false. Note that the default value of aria-multiline is false.

1. The design pattern mentions the use of `aria-multiline=false` if the combobox is editable, but none of our editable comnbobox examples defines the `aria-multiline` property.  What is even more confusing is that the practices say the default value is `false`, so does it need to be defined or not?  If not why do we say anything.   
2. We also do not say what to do if the values in the combobox can only be from the options in the popup.

Item 2
> When a combobox receives focus, DOM focus is placed on the combobox element

Does this statement make any sense now that the textbox has the combobox role?

carmacleod:
Agree with @jongund regarding confusion around mentioning `aria-multiline`. It makes the reader wonder if it's possible to have a combobox with `aria-multiline="true"`? Does combo allow selecting and typing paragraphs? Maybe "Select your favorite cat quote"?

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
     </section>
 
     <section class="widget" id="combobox">
-      <h3>Combo Box</h3>
+      <h3>Combobox</h3>
       <p>
-        A <a href="#combobox" class="role-reference">combobox</a> is a widget made up of the combination of two distinct elements:
-        1) a single-line textbox, and 2) an associated pop-up element for helping users set the value of the textbox.
+        A <a href="#combobox" class="role-reference">combobox</a> is an input widget with an associated popup that enables users to select a value for the combobox from a collection of possible values.
+        In some implementations, the popup presents allowed values, while in other implementations, the popup presents suggested values , and users may either select one of the suggestions or type a value.
```
carmacleod: delete space before comma

examples/grid/LayoutGrids.html
```diff
@@ -240,7 +240,7 @@ <h4 id="ex3_search_results_label">Search Results for "W3C WAI-ARIA"</h4>
               </h5></div> <span tabindex="-1" role="gridcell">World Wide Web Consortium</span> <span
               tabindex="-1" role="gridcell"
             >Jul 21, 2016 - This version of ARIA 1.1 changes the aria-kbdshortcuts property to
-              aria-keyshortcuts , expands roles that can be used in combo boxes, adds a ...</span>
+              aria-keyshortcuts , expands roles that can be used in comboboxes, adds a ...</span>
```
carmacleod: delete space before comma

zcorpan:
@jongund I think both of those can be removed.

The ARIA spec has the requirement about `aria-multiline`, but including it in aria-practices is more confusing than not mentioning it at all.

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
+      <p>
+        The combobox pattern supports several optional behaviors.
+        The one that most shapes interaction is text input.
+        Some comboboxes allow users to type and edit text in the combobox and others do not.
+        If a combobox does not support text input, it is referred to as select-only, meaning the only way users can set a value is by selecting a value in the popup.
+        For example, in some browsers, an HTML <code>select</code> element with <code>size="1"</code> is presented to assistive technologies as a combobox.
+        Alternatively, if a combobox supports text input, it is referred to as editable.
+        An editable combobox may either allow users to input any arbitrary value, or it may restrict its value to a discrete set of allowed values, in which case typing input serves to filter suggestions presented in the popup.
+      </p>
+      <p>The popup is hidden by default, i.e., the default state is collapsed. The conditions that trigger expansion -- display of the popup --are specific to each implementation. Some possible conditions that trigger expansion include:</p>
```
charmarkk:
```suggestion
      <p>The popup is hidden by default, i.e., the default state is collapsed. The conditions that trigger expansion -- display of the popup -- are specific to each implementation. Some possible conditions that trigger expansion include:</p>
```
Added space after em dash following "display of the popup" to separate it from "are specifically."

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
       </ol>
       <p>
-        The nature of the suggested values and the way the suggestions are presented is called the autocomplete behavior.
+        The nature of possible values presented by a popup and the way they are presented is called the autocomplete behavior.
         Comboboxes can have one of four forms of autocomplete:
       </p>
       <ol>
         <li>
-          <strong>No autocomplete:</strong> When the popup is triggered, the suggested values it contains are the same regardless of the characters typed in the textbox.
+          <strong>No autocomplete:</strong> The combobox is editable, and  when the popup is triggered, the suggested values it contains are the same regardless of the characters typed in the combobox.
```
charmarkk:
```suggestion
          <strong>No autocomplete:</strong> The combobox is editable, and when the popup is triggered, the suggested values it contains are the same regardless of the characters typed in the combobox.
```
Remove space after first "and" following "editable."

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
           For example, the popup suggests a set of recently entered values, and the suggestions  do not change as the user types.
```
charmarkk:
```suggestion
          For example, the popup suggests a set of recently entered values, and the suggestions do not change as the user types.
```
Remove space after "suggestions."

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
         </li>
         <li>
-          <strong>List autocomplete with manual selection:</strong> When the popup is triggered, it presents suggested values that complete or logically correspond to the characters typed in the textbox.
-          The character string the user has typed will become the value of the textbox unless the user selects a value in the popup.
+          <strong>List autocomplete with manual selection:</strong> When the popup is triggered, it presents suggested values. If the combobox is editable,  the suggested values complete or logically correspond to the characters typed in the combobox.
```
charmarkk:
```suggestion
          <strong>List autocomplete with manual selection:</strong> When the popup is triggered, it presents suggested values. If the combobox is editable, the suggested values complete or logically correspond to the characters typed in the combobox.
```
Remove space after comma following "editable."

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
-        both ARIA 1.0 and 1.1 patterns are described in the following sections.
-        While using the ARIA 1.1 pattern is recommended as soon as assistive technology support is sufficient,
-        there are no plans to deprecate the ARIA 1.0 pattern.
+        Two other widgets that are also visually compact and enable users to make a single choice from a set of discrete choices are <a href="#Listbox">listbox</a> and <a href="#menubutton">menu button</a>.
+        One feature that distinguishes combobox from both listbox and menu button is that the user's choice can be presented as a value in an editable field, which gives users the ability to select some or all of the value for copying to the clipboard.
+        Comboboxes and menu buttons can be implemented so users can explore the set of allowed choices without losing a previously made choice.
+        That is, users can navigate the set of available choices in a combobox popup or menu and then press <kbd>escape</kbd>, which closes the popup or menu without changing previous input.
```
charmarkk:
```suggestion
        That is, users can navigate the set of available choices in a combobox popup or menu and then press <kbd>Escape</kbd>, which closes the popup or menu without changing previous input.
```
Capitalize "Escape" for consistency.

aria-practices.html
```diff
@@ -812,81 +814,90 @@ <h5>Textbox Keyboard Interaction</h5>
...
-        <ol>
-          <li>include keys for input, cursor movement, selection, and text manipulation.</li>
-          <li>Standard key assignments for editing functions depend on the device operating system.</li>
-          <li>The most robust approach for providing text editing functions is to rely on browsers, which  supply them for HTML inputs with type text and for elements with the <code>contenteditable</code> HTML attribute.</li>
-          <li><strong>IMPORTANT:</strong> Be sure that JavaScript does not interfere with browser-provided text editing functions by capturing key events for the keys used to perform them.</li>
-        </ol>
+        <div class="note">
+          <p>Standard single line text editing keys appropriate for the device platform:</p>
+          <ol>
+            <li>include keys for input, cursor movement, selection, and text manipulation.</li>
```
charmarkk:
```suggestion
            <li>Include keys for input, cursor movement, selection, and text manipulation.</li>
```
Capitalize "include" for consistency with other `<li>`s.

aria-practices.html
```diff
@@ -812,81 +814,90 @@ <h5>Textbox Keyboard Interaction</h5>
...
+            <li>Standard key assignments for editing functions depend on the device operating system.</li>
+            <li>The most robust approach for providing text editing functions is to rely on browsers, which  supply them for HTML inputs with type text and for elements with the <code>contenteditable</code> HTML attribute.</li>
```
charmarkk:
```suggestion
            <li>The most robust approach for providing text editing functions is to rely on browsers, which supply them for HTML inputs with type text and for elements with the <code>contenteditable</code> HTML attribute.</li>
```
Remove space between "which supply."

aria-practices.html
```diff
@@ -812,81 +814,90 @@ <h5>Textbox Keyboard Interaction</h5>
...
+            <li><strong>IMPORTANT:</strong> Ensure JavaScript does not interfere with browser-provided text editing functions by capturing key events for the keys used to perform them.</li>
+          </ol>
+        </div>
         <h5>Listbox Popup Keyboard Interaction</h5>
         <p>When focus is in a listbox popup:</p>
         <ul>
-          <li><kbd>Enter</kbd>: Accepts the focused option in the listbox by closing the popup and placing the accepted value in the textbox with the input cursor at the end of the value.</li>
-          <li><kbd>Escape</kbd>: Closes the popup and returns focus to the textbox. Optionally, clears the contents of the textbox.</li>
+          <li><kbd>Enter</kbd>: Accepts the focused option in the listbox  by closing the popup, placing the accepted value in the combobox, and if the combobox is editable,  placing the input cursor at the end of the value.</li>
```
charmarkk:
```suggestion
          <li><kbd>Enter</kbd>: Accepts the focused option in the listbox by closing the popup, placing the accepted value in the combobox, and if the combobox is editable,  placing the input cursor at the end of the value.</li>
```
Remove space between "listbox by."

aria-practices.html
```diff
@@ -812,81 +814,90 @@ <h5>Textbox Keyboard Interaction</h5>
...
-          <li><kbd>Delete</kbd> (Optional): Returns focus to the textbox, removes the selected state if a suggestion was selected, and removes the inline autocomplete string if present.</li>
-          <li><kbd>Down Arrow</kbd>: Moves focus to and selects the next option. If focus is on the last option, either returns focus to the textbox or does nothing.</li>
-          <li><kbd>Up Arrow</kbd>: Moves focus to and selects the previous option. If focus is on the first option, either returns focus to the textbox or does nothing.</li>
-          <li><kbd>Home</kbd> (Optional): Either moves focus to and selects the first option or returns focus to the textbox and places the cursor on the first character.</li>
-          <li><kbd>End</kbd> (Optional): Either moves focus to the last option or returns focus to the textbox and places the cursor after the last character.</li>
+          <li><kbd>Home</kbd> (Optional): Either moves focus to and selects the first option or, if the combobox is editable,  returns focus to the combobox and places the cursor on the first character.</li>
```
charmarkk:
```suggestion
          <li><kbd>Home</kbd> (Optional): Either moves focus to and selects the first option or, if the combobox is editable, returns focus to the combobox and places the cursor on the first character.</li>
```
Remove space after "editable."

carmacleod:
See example bug reported by @LaurenceRLewis in https://github.com/w3c/aria-practices/issues/1261#issuecomment-553581609 ... every instance of `aria-haspopup="true"` in the example code and in the Role, Property, State section of the examples should either be changed to `aria-haspopup="listbox"` or deleted completely.

aria-practices.html
```diff
@@ -984,91 +1003,82 @@ <h5>Tree Popup Keyboard Interaction</h5>
...
           </li>
-          <li>The textbox element has <a href="#aria-autocomplete" class="property-reference">aria-autocomplete</a> set to a value that corresponds to its autocomplete behavior:
+          <li>The combobox element has <a href="#aria-autocomplete" class="property-reference">aria-autocomplete</a> set to a value that corresponds to its autocomplete behavior:
             <ul>
-              <li><code>none</code>: When the popup is displayed, the suggested values it contains are the same regardless of the characters typed in the textbox.</li>
-              <li><code>list</code>: When the popup is triggered, it presents suggested values that complete or logically correspond to the characters typed in the textbox.</li>
+              <li><code>none</code>: When the popup is displayed, the suggested values it contains are the same regardless of the characters typed in the combobox.</li>
+              <li><code>list</code>: When the popup is triggered, it presents suggested values. If the combobox is editable,  the values complete or logically correspond to the characters typed in the combobox.</li>
```
charmarkk:
```suggestion
              <li><code>list</code>: When the popup is triggered, it presents suggested values. If the combobox is editable, the values complete or logically correspond to the characters typed in the combobox.</li>
```
Remove space after "editable."

aria-practices.html
```diff
@@ -6583,12 +6593,12 @@ <h2>Structural Roles</h2>
       These roles express the meaning that is conveyed by the layout and appearance of elements that organize and structure content, such as headings, lists, and tables.
     </p>
     <p>
-      some host languages, such as HTML, include elements that express the same  semantics as an ARIA role. 
+      some host languages, such as HTML, include elements that express the same  semantics as an ARIA role.
```
charmarkk:
```suggestion
      some host languages, such as HTML, include elements that express the same semantics as an ARIA role.
```
Remove space after "same."

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
+        Many implementations also include a third optional element -- a graphical <q>Open</q> button adjacent to the combobox, which indicates availability of the popup.
+        Activating the <q>Open</q> button displays the popup if suggestions are available.
       </p>
-      <p>The popup is hidden by default, and the conditions that trigger its display are specific to each implementation. Some possible popup display conditions include:</p>
+      <p>
+        The combobox pattern supports several optional behaviors.
+        The one that most shapes interaction is text input.
+        Some comboboxes allow users to type and edit text in the combobox and others do not.
+        If a combobox does not support text input, it is referred to as select-only, meaning the only way users can set a value is by selecting a value in the popup.
+        For example, in some browsers, an HTML <code>select</code> element with <code>size="1"</code> is presented to assistive technologies as a combobox.
```
carmacleod: consider changing "as a combobox" to "as a select-only combobox" to tie this sentence in to the flow of the discussion.

charmarkk: Is this not just saying, effectively, that a `select` with `size="1"` is announced as a combobox by AT? In other words, AT doesn't announce it as "select-only," right? That's how I read it at least.

aria-practices.html
carmacleod: add space: "--are"

aria-practices.html
```diff
@@ -710,100 +710,102 @@ <h4>WAI-ARIA Roles, States, and Properties</h4>
...
-        <li>It is displayed as soon as the textbox is focused, even if the textbox is empty.</li>
-        <li>It is displayed when the <kbd>Down Arrow</kbd> key is pressed or the <q>show</q> button is activated, possibly with a dependency on the content of the textbox.</li>
-        <li>It is displayed if the value of the textbox is altered in a way that creates one or more partial matches to a suggested value.</li>
+        <li>It is displayed when the <kbd>Down Arrow</kbd> key is pressed or the <q>Open</q> button is activated. Optionally, if the combobox is editable and contains a string that does not match an allowed value, expansion does not occur.</li>
+        <li>It is displayed when the combobox receives focus even if the combobox is editable and empty.</li>
+        <li>If the combobox is editable, the popup is displayed only if a certain number of characters are typed in the combobox and those characters match some portion of one of the suggested values.</li>
+        <li>If the combobox is editable, it is displayed if the value of the combobox is altered in a way that creates one or more partial matches to a suggested value.</li>
```
carmacleod: This last point (and how it differs from the one preceding it) is hard to parse.
i.e. what does "altered in a way" mean? Typing, pasting, or selecting from the list?
How is this different from the previous point? Is the key difference "only if a certain number of characters are typed"?

aria-practices.html
carmacleod: also, add "the" after "previous" so that it reads, "changing the previous input".

spectranaut:
Tests look good to me! I just notice one test that was marked as failing that has previously been fixed.

aria-practices.html
carmacleod: Agree. Please add thoughts to this issue: https://github.com/w3c/aria-practices/issues/1026

zcorpan:
I've done visual design review in https://github.com/w3c/aria-practices/issues/1263.

> Our examples are not on par with modern design systems/component frameworks.

aria-practices.html
```diff
@@ -812,81 +814,90 @@ <h5>Textbox Keyboard Interaction</h5>
...
           <li>
-            <kbd>Enter</kbd>: If an autocomplete suggestion is automatically selected, accepts the suggestion either by placing the input cursor at the end of the accepted value in the textbox or by performing a default action on the value.
-            For example, in a messaging application, the default action may be to add the accepted value to a list of message recipients and then clear the textbox so the user can add another recipient.
+            <kbd>Enter</kbd>: If the combobox is editable and an autocomplete suggestion is selected in the popup, accepts the suggestion either by placing the input cursor at the end of the accepted value in the combobox or by performing a default action on the value.
+            For example, in a messaging application, the default action may be to add the accepted value to a list of message recipients and then clear the combobox so the user can add another recipient.
+          </li>
+          <li>Printable Characters:
+            <ul>
+              <li>If the combobox is editable, type characters in the combobox. Note that some implementations may regard certain characters as invalid and prevent their input.</li>
```
```suggestion
              <li>If the combobox is editable, types characters in the combobox. Note that some implementations may regard certain characters as invalid and prevent their input.</li>
```

aria-practices.html
```diff
@@ -812,81 +814,90 @@ <h5>Textbox Keyboard Interaction</h5>
...
-          <li>Selection follows focus in the listbox; the listbox allows only one suggested value to be selected at a time for the textbox value.</li>
+          <li>DOM Focus is maintained on the combobox and the assistive technology focus is moved within the listbox using <code>aria-activedescendant</code> as described in <a href="#kbd_focus_activedescendant">Managing Focus in Composites Using aria-activedescendant.</a></li>
+          <li>Selection follows focus in the listbox; the listbox allows only one suggested value to be selected at a time for the combobox value.</li>
         </ol>
         <h5>Grid Popup Keyboard Interaction</h5>
         <p>
           In a grid popup, each suggested value may be represented by either a single cell or an entire row.
           See notes below for how this aspect of grid design effects the keyboard interaction design and the way that selection moves in response to focus movements.
```
```suggestion
          See notes below for how this aspect of grid design affects the keyboard interaction design and the way that selection moves in response to focus movements.
```

aria-practices.html
```diff
@@ -899,21 +910,24 @@ <h5>Grid Popup Keyboard Interaction</h5>
...
         </ul>
         <ol class="note">
-          <li>DOM Focus is maintained on the combobox textbox and the assistive technology focus is moved within the grid using <code>aria-activedescendant</code> as described in <a href="#kbd_focus_activedescendant">Managing Focus in Composites Using aria-activedescendant.</a></li>
-          <li>The grid allows only one suggested value to be selected at a time for the textbox value.</li>
+          <li>DOM Focus is maintained on the combobox and the assistive technology focus is moved within the grid using <code>aria-activedescendant</code> as described in <a href="#kbd_focus_activedescendant">Managing Focus in Composites Using aria-activedescendant.</a></li>
+          <li>The grid allows only one suggested value to be selected at a time for the combobox value.</li>
           <li>In a grid popup, each suggested value may be represented by either a single cell or an entire row. This aspect of design effects focus and selection movement:
```
```suggestion
          <li>In a grid popup, each suggested value may be represented by either a single cell or an entire row. This aspect of design affects focus and selection movement:
```

aria-practices.html
```diff
@@ -899,21 +910,24 @@ <h5>Grid Popup Keyboard Interaction</h5>
...
             </ul>
           </li>
           <li><kbd>End</kbd> (Optional): <strong>Either:</strong>
             <ul>
               <li> Moves focus to the last cell in the row that contains focus. Or, if the grid has fewer than three cells per row or multiple suggested values per row, focus may move to the last cell in the grid.</li>
-              <li>Returns focus to the textbox and places the cursor after the last character.</li>
+              <li>If the combobox is editable, returns focus to the combobox and places the cursor after the last character.</li>
             </ul>
           </li>
           <li><kbd>Control + Home</kbd> (optional): moves focus to the first row.</li>
```
```suggestion
          <li><kbd>Control + Home</kbd> (Optional): moves focus to the first row.</li>
```

aria-practices.html
```diff
@@ -899,21 +910,24 @@ <h5>Grid Popup Keyboard Interaction</h5>
...
           <li><kbd>Control + End</kbd> (Optional): moves focus to the last row.</li>
+          <li>Any printable character: If the combobox is editable, returns the focus to the combobox without closing the popup and types the character.</li>
```
```suggestion
          <li>Any printable character: If the combobox is editable, returns the focus to the combobox without closing the popup and types the character into the combobox.</li>
```

examples/combobox/grid-combo.html
```diff
@@ -77,13 +76,13 @@ <h2 id="ex_label">Example</h2>
       </div>
```
smhigley: Suggest adding `aria-readonly="true"` to the grid, to prevent each individual grid cell announcing it is editable (as happens in NVDA + FF)

examples/combobox/js/combobox-list.js
```diff
@@ -57,7 +57,7 @@ ComboboxList.prototype.init = function () {
```
smhigley: There's an error in this file in how the listbox handles the blur event --

line 349 calls `this.removeVisualFocusAll();`, which sets `listbox.hasFocus` to true, which prevents the listbox from closing. The end result is that the listbox does not close when you click outside of it.

aria-practices.html
mcking65: got it, thanks.

aria-practices.html
mcking65: Mark, correct, screen readers present selects as a plain combobox.

aria-practices.html
```diff
@@ -961,10 +974,16 @@ <h5>Tree Popup Keyboard Interaction</h5>
           <li><kbd>Up Arrow</kbd>: Moves focus to the previous node that is focusable without opening or closing a node and selects it if it is selectable.</li>
           <li><kbd>Home</kbd>: Moves focus to the first node in the tree without opening or closing a node and selects it if it is selectable.</li>
```
```suggestion
          <li><kbd>Home</kbd>: Moves focus to the first focusable node in the tree without opening or closing a node and selects it if it is selectable.</li>
```

aria-practices.html
```diff
@@ -961,10 +974,16 @@ <h5>Tree Popup Keyboard Interaction</h5>
...
           <li><kbd>End</kbd>: Moves focus to the last node in the tree that is focusable without opening a node and selects it if it is selectable.</li>
```
```suggestion
          <li><kbd>End</kbd>: Moves focus to the last focusable node in the tree without opening or closing a node and selects it if it is selectable.</li>
```

The prose for Home/End should be exactly the same except for first/last.

aria-practices.html
```diff
@@ -984,91 +1003,82 @@ <h5>Tree Popup Keyboard Interaction</h5>
...
-            </ul>
-          </li>
-        </ol>
+        <div class="note">
+          <ol>
+            <li>
+              In ARIA 1.0, the combobox referenced its popup with <a href="#aria-owns" class="property-reference">aria-owns</a> instead of <a href="#aria-controls" class="property-reference">aria-controls</a>.
+              While user agents might support comboboxes with <a href="#aria-owns" class="property-reference">aria-owns</a> for backwards compatibility with legacy content, it is strongly recommended that authors use <a href="#aria-controls" class="property-reference">aria-controls</a>.
+            </li>
+            <li>When referring to the roles, states, and properties documentation for the below list of patterns used for popups, keep in mind that a combobox is a single-select widget where selection always follows focus in the popup.</li>
```
```suggestion
            <li>When referring to the roles, states, and properties documentation for the below list of patterns used for popups, keep in mind that a combobox is a single-select widget where selection follows focus in the popup.</li>
```

Maybe delete "always", because selection does not always follow focus in tree popup.

examples/combobox/combobox-autocomplete-both.html
```diff
@@ -196,7 +191,7 @@ <h3 id="kbd_label_listbox">Listbox Popup</h3>
         <strong>NOTE:</strong> When visual focus is in the listbox, DOM focus remains on the textbox and the value of <code>aria-activedescendant</code> on the textbox is set to a value that refers to the listbox option that is visually indicated as focused.
```
carmacleod: For Up arrow, change "last" to "previous" as follows:
> If the listbox is displayed and a suggestion is selected, moves visual focus to the last suggested value.

to:
> If the listbox is displayed and a suggestion is selected, moves visual focus to the previous suggested value.

JAWS-test:
> IE: none of the listbox comboboxes work at all, with or without screen readers running

* IE without screenreader: This was also true for the old examples and seems to be a CSS problem: The list entries are not displayed. I suspect this is easy to fix. 
* IE with JAWS: JAWS + IE 11 does not support any `combobox` pattern (ARIA 1.0, ARIA 1.1., ARIA 1.2). There is no solution for this (https://github.com/FreedomScientific/VFO-standards-support/issues/221). I am in favor of including a clear warning that `role=combobox` must be avoided in IE 11 and other elements should be used (`<select>`, input field with autocomplete etc).

aria-practices.html
mcking65: Agree, deleting the last one.

carmacleod:
Functional issues:
- I agree with @smhigley that Escape should not clear the text input if the popup is open. The user just wants to close the popup at that point - they do not want their previous value to be erased without their consent. Typing Escape with the popup closed can clear the text input. Fixing this includes rewriting the keyboard interaction prose for Escape in the pattern and example.
The [Combobox Keyboard Interaction](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/aria-practices.html#combobox-keyboard-interaction) in the pattern correctly says:
    > Escape: Dismisses the popup if it is visible. Optionally, if the popup is hidden before Escape is pressed, clears the combobox.

    But then the [Listbox Popup Keyboard Interaction](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/aria-practices.html#listbox-popup-keyboard-interaction), [Grid](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/aria-practices.html#grid-popup-keyboard-interaction), and [Tree](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/aria-practices.html#tree-popup-keyboard-interaction) change it for some reason, to say:
    > Escape: Closes the popup and returns focus to the combobox. Optionally, if the combobox is editable, clears the contents of the combobox.

    Then, the [Dialog](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/aria-practices.html#dialog-popup-keyboard-interaction) changes it yet again to be optional:
    > Cancel out of the dialog, e.g., press Escape or activate the cancel button in the dialog. Canceling either returns focus to the text box without changing the combobox value or returns focus to the combobox and clears the combobox.

- Also agree that down arrow should open the popup in the grid example.

Additional functional issues:
- the black hover outline appears when hovering over what looks like empty space off to the right of the button (but is actually the combo's label element). This feels unexpected.

- I somehow managed to cause the popup to show only "Utah" while the combobox value was "Wyoming" (the last value in the list - which I had not selected). This occurred on 2 separate occasions when returning to the open browser tab. I don't think this should be possible. :)
proof of Wyoming-Utah bug

examples/combobox/combobox-autocomplete-both.html
mcking65: That's not what happens when focus is in the textbox ... this table is for textbox. If it is in the textbox, up takes you to last suggestion

aria-practices.html
mcking65: I revised this text to make the actual requirement more clear but specify that it is valid to refer to an element that is not visible.

examples/combobox/grid-combo.html
mcking65: This is an NVDA bug. It would not align with the ARIA spec to add aria-readonly. There is actually no such thing as a editable grid. You can have an editable gridcell in a spreadshett type application where the cell content is directly editable, but the grid itself is never editable. This is explained in the spec.

mcking65:
JAWS-test wrote:

> > IE: none of the listbox comboboxes work at all, with or without screen readers running
> 
> * IE without screenreader: This was also true for the old examples. The list entries are empty (also in the source code). Probably a JS problem.
> * IE with JAWS: JAWS + IE 11 does not support any `combobox` pattern (ARIA 1.0, ARIA 1.1., ARIA 1.2). There is no solution for this ([FreedomScientific/VFO-standards-support#221](https://github.com/FreedomScientific/VFO-standards-support/issues/221)). I am in favor of including a clear warning that `role=combobox` must be avoided in IE 11 and other elements should be used (`<select>`, input field with autocomplete etc).

I believe this is solveable if Freedom Scientific wishes to solve it. I think it is related to aria-activedescendant. @accdc probably knows the root cause.

It is likely APG will soon drop IE support. However, given the popularity of comboboxes, there may be a good case for making these examples operational in IE if FS does want to make them work in IE.

mcking65:
Sarah, thank you for the fabulous review.

@smhigley wrote:

> * NVDA + Firefox: reads the page title four times almost every time I change options using the arrow key with the listbox open (this may be related to the fact that all option elements are mutated every time one changes selection; I'd be happy to investigate the root cause further). This is a code bug, not an issue with the pattern, as it did not repro in a different implementation.
> * JAWS + Firefox: would sporadically re-announce the listbox when arrowing through options. Seems likely related to the above

Opened #1264 and scheduled to resolve in next milestone.

> * IE: none of the listbox comboboxes work at all, with or without screen readers running

I didn't open an issue for this since we previously discussed dropping IE support in the new year.

> * IE: the grid combobox does not work with JAWS; options are not read when arrowed to.

Known bug in JAWS support for grids.

> * macOS + VoiceOver: the suggested "open" keyboard command does not work (Caps + Space), though Option + down arrow does
> * macOS + VoiceOver: although everything but Caps+ Space worked for me in 10.14, @jnurthen reported that 10.15 did not work; someone else with Catalina should probably officially test this.

My system is still 10.4.6, and it works for me too. We will need to continues this testing prior to APG 1.2 R1.

> * iOS + VoiceOver: the "open" button did not work (didn't open the listbox, and sent the cursor back to the text input). I'd suggest either making the open button `aria-hidden="true" tabindex="-1"`, or removing it altogether.

Surprising suggestion. How would touch users expand/collapse the combobox if it is not accessible?

This is a widget that we should make touch compatible before APG 1.2 release. Opened #1267.

> * Windows high contrast mode: the selected option within the listbox is not visible

Added this to issue #1263.

> General functionality notes and issues:
> 
> * The listbox/grid popup does not close when clicking outside of it (detailed comment on why in the `combobox-list.js` file)

Opened #1265 and scheduled for next milestone.

> * Escape clears the text input (this has been mentioned before). It also means that if you have a currently selected item and open the listbox, it's impossible to dismiss the listbox and restore the original selection with escape

Added to #1066.

> * Grid combobox: it seems odd that this combobox (and only this one) does not let you open the popup and see all options without first typing something
> * This isn't fully a bug, but I find it odd that the list scrolls every time you navigate to a new option, even if that option was already in view before the programmatic scroll

Opened #1266 and schedule for next milestone.

mcking65:
Carolyn,thank you for the awesome reviews!!

Except for the 2 where I commented otherwise, I merged all your suggestions.

@carmacleod wrote:

> Functional issues:
> 
> * I agree with @smhigley that Escape should not clear the text input if the popup is open. The user just wants to close the popup at that point - they do not want their previous value to be erased without their consent. Typing Escape with the popup closed can clear the text input. Fixing this includes rewriting the keyboard interaction prose for Escape in the pattern and example.

Added this to #1066.

> * the black hover outline appears when hovering over what looks like empty space off to the right of the button (but is actually the combo's label element). This feels unexpected.

Is this something that could be addressed in #1263? Or added to one of the other open issues in the combobox project?

> * I somehow managed to cause the popup to show only "Utah" while the combobox value was "Wyoming" (the last value in the list - which I had not selected). This occurred on 2 separate occasions when returning to the open browser tab. I don't think this should be possible. :)
>   proof of Wyoming-Utah bug

Opened #1268 and scheduled in next milestone.

JAWS-test:
> I believe this is solveable if Freedom Scientific wishes to solve it. 

I'm afraid that won't happen because the problem has been there for many years and not a single improvement has happened since. 

> I think it is related to aria-activedescendant.

The problem has nothing to do with `aria-activedescendant` (this works fine now, it was just a problem in the past). The problem is that `role=combobox` is not recognized by JAWS. So it's not the list entries that are affected, but the element itself.

mcking65:
@jongund wrote:

> In the ARIA practices section for combobox "WAI-ARIA Roles, States, and Properties"
> 
> Item 1
> 
> > If the combobox element is editable, it has a value for aria-multiline of false. Note that the default value of aria-multiline is false.
> 
> 1. The design pattern mentions the use of `aria-multiline=false` if the combobox is editable, but none of our editable comnbobox examples defines the `aria-multiline` property.  What is even more confusing is that the practices say the default value is `false`, so does it need to be defined or not?  If not why do we say anything.

Removed it in commit dd6f6c0.

> 2. We also do not say what to do if the values in the combobox can only be from the options in the popup.

The very first bullet says that the element that displays the value has role combobox. That's all that is needed. However, I do think that we might eventually benefit from a property something like "aria-valuefrom" that would take an ID and explicitly define the source of the value for the combobox.

> Item 2
> 
> > When a combobox receives focus, DOM focus is placed on the combobox element
> 
> Does this statement make any sense now that the textbox has the combobox role?

Certainly ... we need focus on that element. Obviously, it gets it by default if you use html input type text, but you would need to explicitly make the element focusable if it were not an element that is focusable by default.

mcking65:
@@charmarkk, thank you very much for the thorough review!! I believe I have merged all your suggestions.

mcking65:
@carmacleod wrote:

> See example bug reported by @LaurenceRLewis in [#1261 (comment)](https://github.com/w3c/aria-practices/issues/1261#issuecomment-553581609) ... every instance of `aria-haspopup="true"` in the example code and in the Role, Property, State section of the examples should either be changed to `aria-haspopup="listbox"` or deleted completely.

Removed with commit e51ad04.

@spectranaut, This will need to be changed in the regression tests.

carmacleod:
> > * the black hover outline appears when hovering over what looks like empty space off to the right of the button (but is actually the combo's label element). This feels unexpected.
> 
> Is this something that could be addressed in #1263? Or added to one of the other open issues in the combobox project?

Added to #1263 by @zcorpan.

examples/combobox/combobox-autocomplete-both.html
carmacleod: Nope; with focus in textbox, up arrow goes to previous suggestion in list. :)
Steps:
- focus textbox
- type down arrow 6 times; you will be on California
- type up arrow once; you will be on Arkansas, not Wyoming

carmacleod:
#### VoiceOver+Chrome on macOS Catalina 10.15.1

- focusing the combobox and then typing down arrow: all 3 listbox examples say the list item twice, although the first time is cut short. This is most apparent in the [autocomplete=none example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-none.html) for "w-weather, ch-cheap flights to NY".

- the [autocomplete=both example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-both.html) has some very weird behavior where sometimes only a part of the previously-selected list item is spoken as you arrow through the list. For example: focus the combo and type down arrow, and Alaska is selected and spoken (even though the first item is Alabama); type down arrow again, and VO says "lask", down arrow again to hear "merican Samoa Arizona". If you wait long enough before typing down arrow again, it may synchronize so that the selection matches what is spoken. But then down arrow again results in "District of Columbi". It's pretty unusable.

- `ctrl+alt+space` (i.e. VO+space) does open the popup in the listbox (not grid) examples in Chrome (but does not focus the first item)

#### VoiceOver+Safari on macOS Catalina 10.15.1

- VO does not read the list items in the [autocomplete=list](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-list.html) or [autocomplete=none](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-none.html) or [grid](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/grid-combo.html) examples in Safari. These 3 examples are completely unusable in VO+Safari on Catalina.

- However, VO _does_ read list items in the [autocomplete=both](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-both.html) example

- VO+space does not open the list in Safari (for any example)

mcking65:
@carmacleod wrote:

> #### VoiceOver+Chrome on macOS Catalina 10.15.1
> * focusing the combobox and then typing down arrow: all 3 listbox examples say the list item twice, although the first time is cut short. This is most apparent in the [autocomplete=none example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-none.html) for "w-weather, ch-cheap flights to NY".
> * the [autocomplete=both example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-both.html) has some very weird behavior where sometimes only a part of the previously-selected list item is spoken as you arrow through the list. For example: focus the combo and type down arrow, and Alaska is selected and spoken (even though the first item is Alabama); type down arrow again, and VO says "lask", down arrow again to hear "merican Samoa Arizona". If you wait long enough before typing down arrow again, it may synchronize so that the selection matches what is spoken. But then down arrow again results in "District of Columbi". It's pretty unusable.

Both these behaviors smell like bugs in our code and are also probably related to #1264. @spectranaut is working on #1264 today.

> * `ctrl+alt+space` (i.e. VO+space) does open the popup in the listbox (not grid) examples in Chrome (but does not focus the first item)

Interesting, confusing.

> #### VoiceOver+Safari on macOS Catalina 10.15.1
> * VO does not read the list items in the [autocomplete=list](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-list.html) or [autocomplete=none](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-none.html) or [grid](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/grid-combo.html) examples in Safari. These 3 examples are completely unusable in VO+Safari on Catalina.

The fact that they do work in 10.14.6 does not necessarily implicate Safari, but kind of points in that direction. @accdc has some different implementations of the same pattern we could test with 10.15. Apple may have implemented overly restrictive processing of aria-activedescendant, overlooking the fact that the ARIA spec supports activedescendant on a text input or combobox that controls an element that also supports activedescendant.

> * However, VO _does_ read list items in the [autocomplete=both](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-both.html) example

This is almost certainly because of the inline selection.

> * VO+space does not open the list in Safari (for any example)

VO-space is supposed to be equivalent to a mouse click on the element in the VO cursor. Given that, does VO announcing that VO-space should open the popup make sense?

jongund:
@mcking65 do we distinguish the term "focus" from "DOM focus" in the practices?

> Item 2
> 
> When a combobox receives focus, DOM focus is placed on the combobox element
> 
> Does this statement make any sense now that the textbox has the combobox role?
> 
> Certainly ... we need focus on that element. Obviously, it gets it by default if you use html input type text, but you would need to explicitly make the element focusable if it were not an element that is focusable by default.
>

charmarkk:
> > * VO+space does not open the list in Safari (for any example)
> 
> VO-space is supposed to be equivalent to a mouse click on the element in the VO cursor. Given that, does VO announcing that VO-space should open the popup make sense?

I think that does make sense, Matt.

Issue #1268, opened 2019-11-14
Combobox Examples: Popup inccorectly contains only Utah when Wyoming in input


mcking65:
From @carmacleod in #1255:

> * I somehow managed to cause the popup to show only "Utah" while the combobox value was "Wyoming" (the last value in the list - which I had not selected). This occurred on 2 separate occasions when returning to the open browser tab. I don't think this should be possible. :)
>   ![proof of Wyoming-Utah bug](https://user-images.githubusercontent.com/3331913/68826080-67714d00-066a-11ea-9e0e-ccc0e211a825.png)

Issue #1265, opened 2019-11-14
Combobox examples: Make clicking outside popup close the popup


mcking65:
Per review of the combobox examples by @smhigley in #1255, the listbox/grid popup does not close when clicking outside of it (detailed comment on why in the #1255 review of `combobox-list.js`).

Issue #1263, opened 2019-11-13
Visual design review of combobox PR


zcorpan:
[Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review) of combobox PR https://github.com/w3c/aria-practices/pull/1255

> Review the visual presentation to assess whether look and feel is similar to typical modern implementations of the widget. Allow for adjustments to accommodate accessibility requirements in cases where most modern implementations are not sufficiently accessible.

There are 4 examples:

* [Editable Combobox With Both List and Inline Autocomplete Example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-both.html)
* [Editable Combobox With List Autocomplete Example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-list.html)
* [Editable Combobox without Autocomplete Example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/combobox-autocomplete-none.html)
* [Editable Combobox with Grid Popup Example](https://raw.githack.com/w3c/aria-practices/issue1244-aria1.2-combobox-conversion/examples/combobox/grid-combo.html)

The first 3 have the same visual design, so I'll only review the first one of those ("example 1"). The grid popup example ("example 4") has different visual design.

## Example 1

First, let's compare the default state of the example side by side in Safari (left), Chrome (middle), and Firefox (right).

<img width="742" alt="The borders are different. The spacing between the combobox and the button is different. The combobox and the button don't align in Safari and Chrome." src="https://user-images.githubusercontent.com/244772/68805565-5cb3b980-0664-11ea-97bf-3483454d3d91.png">

The following screenshots are taken with Firefox Nightly on macOS.

Default state
<img width="365" alt="The label State on one line, a gray textfield with a 2px inset border, next to it a gray button with a 2px outset border and a downward-pointing triangle." src="https://user-images.githubusercontent.com/244772/68807368-fd57a880-0667-11ea-89f0-a1d20223995c.png">

Hover on the combobox field
<img width="377" alt="The combobox has a black outline." src="https://user-images.githubusercontent.com/244772/68807428-106a7880-0668-11ea-96e5-1fed21d66e05.png">

Hover on the button
[...]

Issue #785, opened 2018-07-19
Combobox examples: missing description about the id attribute


spectranaut:
The description for row `id="string"` in the textbox `Role, Property, State, and Tabindex Attributes` table is unclear to me.

It says "Referenced by `for` attribute of `label` element to provide an accessible label." 

The second two examples do no have a "for" attribute on their "label" elements, they instead have "aria-labelledby" on their textbox.

mcking65:
Potentially related to issue #542.

carmacleod:
I had a look at this issue, using the preview links from PR https://github.com/w3c/aria-practices/pull/1276:
* [Preview combobox autocomplete=both](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-both.html)
* [Preview combobox autocomplete=list](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-list.html)
* [Preview combobox autocomplete=none](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-none.html)

All 3 examples now use an HTML label with a `for` attribute, so there's no longer any use or mention of `aria-labelledby` (which is good).

However, the "Role, Property, State, and Tabindex Attributes" table no longer lists the `id` for the combobox's text input.

Perhaps it should? Maybe something similar to the `id` row in the [grid-combo example](https://w3c.github.io/aria-practices/examples/combobox/grid-combo.html):


Role | Attribute | Element | Usage
-- | -- | -- | --
combobox | id="string" | input[type="text"] | - Referenced by `for` attribute of `label` element to provide an accessible name. - Recommended naming method for HTML input elements because clicking label focuses input.

@spectranaut Would adding that row help clarify the 3 previews from PR https://github.com/w3c/aria-practices/pull/1276?

zcorpan:
Fixed in https://github.com/w3c/aria-practices/pull/1276

PR #1276, opened 2019-11-22
Combobox Examples with listbox popup: Fix Escape behavior, mouse interaction, visual design, and more 

jongund:
I have updated the combobox autocomplete examples:

* Rewrote the JS code to be a single prototype object, so code is simpler to read
* Improved escape key behavior based on JQuery Combobox Escape key behavior (Issue 1066)
* Updated regression tests for new escape key behavior
* Fixed other bugs related to previous code:
  * Issue #1268: Combobox Examples: Popup incorrectly contains only Utah when Wyoming in input 
  * Issue #1265: Combobox examples: Make clicking outside popup close the popup
  * Issue #1263: Combobox examples: Visual design improved by using SVG for button
  * Issue #785: Combobox examples: missing description about the id attribute

#### Preview links

* [Preview combobox autocomplete=both](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-both.html)
* [Preview combobox autocomplete=list](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-list.html)
* [Preview combobox autocomplete=none](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-none.html)

### Review checklist

* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @ZoeBijl 


test/tests/combobox_autocomplete-both.js
```diff
@@ -349,6 +349,29 @@ ariaTest('Test escape key press with focus on textbox',
 
     // Confirm the listbox is closed and the textboxed is clearedx
```
carmacleod: clearedx -> cleared

test/tests/combobox_autocomplete-both.js
```diff
@@ -349,6 +349,29 @@ ariaTest('Test escape key press with focus on textbox',
...
+  exampleFile, 'textbox-key-escape', async (t) => {
+    t.plan(2);
+
+    // Send key "a", then key ESCAPE to the textbox
+
+    await t.context.session
+      .findElement(By.css(ex.textboxSelector))
+      .sendKeys('a', Key.ESCAPE, Key.ESCAPE);
+
+    // Confirm the listbox is closed and the textboxed is clearedx
```
carmacleod: clearedx -> cleared

test/tests/combobox_autocomplete-list.js
```diff
@@ -356,6 +356,29 @@ ariaTest('Test escape key press with focus on textbox',
 
     // Confirm the listbox is closed and the textboxed is cleared
```
carmacleod: textboxed -> textbox

test/tests/combobox_autocomplete-list.js
```diff
@@ -356,6 +356,29 @@ ariaTest('Test escape key press with focus on textbox',
...
+  exampleFile, 'textbox-key-escape', async (t) => {
+    t.plan(2);
+
+    // Send key "a", then key ESCAPE to the textbox
+
+    await t.context.session
+      .findElement(By.css(ex.textboxSelector))
+      .sendKeys('a', Key.ESCAPE, Key.ESCAPE);
+
+    // Confirm the listbox is closed and the textboxed is cleared
```
carmacleod: textboxed -> textbox

test/tests/combobox_autocomplete-none.js
```diff
@@ -356,6 +356,30 @@ ariaTest('Test escape key press with focus on textbox',
 
     // Confirm the listbox is closed and the textboxed is cleared
```
carmacleod: textboxed -> textbox

test/tests/combobox_autocomplete-none.js
```diff
@@ -356,6 +356,30 @@ ariaTest('Test escape key press with focus on textbox',
...
+    t.plan(2);
+
+    // Send key "a" then key "ARROW_DOWN to put the focus on the listbox,
+    // then key ESCAPE to the textbox
+
+    await t.context.session
+      .findElement(By.css(ex.textboxSelector))
+      .sendKeys('a', Key.ARROW_DOWN, Key.ESCAPE, Key.ESCAPE);
+
+    // Confirm the listbox is closed and the textboxed is cleared
```
carmacleod: textboxed -> textbox

test/tests/combobox_autocomplete-both.js
```diff
@@ -349,6 +349,29 @@ ariaTest('Test escape key press with focus on textbox',
 
     // Confirm the listbox is closed and the textboxed is clearedx
 
+    await assertAttributeValues(t, ex.textboxSelector, 'aria-expanded', 'false');
+    t.is(
+      await t.context.session
+        .findElement(By.css(ex.textboxSelector))
+        .getAttribute('value'),
+      'Alabama',
+      'In key press "ESCAPE" should result in first option in textbox'
```
carmacleod: In key press -> In listbox key press

test/tests/combobox_autocomplete-list.js
```diff
@@ -356,6 +356,29 @@ ariaTest('Test escape key press with focus on textbox',
 
     // Confirm the listbox is closed and the textboxed is cleared
 
+    await assertAttributeValues(t, ex.textboxSelector, 'aria-expanded', 'false');
+    t.is(
+      await t.context.session
+        .findElement(By.css(ex.textboxSelector))
+        .getAttribute('value'),
+      'a',
+      'In key press "ESCAPE" should result in first option in textbox'
```
carmacleod: In key press -> In listbox key press

carmacleod:
@jongund
I updated the preview links because they were going to old copies of the 3 examples.

I tested very quickly and noticed one issue. This can be seen in all 3 comboboxes:
1. tab to the combobox to focus it
2. type down arrow a few times - this will open the listbox and move through the items
3. type enter - this will choose an item and close the listbox
4. now I expected to be able to type down arrow again to reopen the list, but nothing happens
    (you can type character keys, or delete, or even left or right arrow, but down and up arrow are ignored)

I think focus needs to go back to the textbox after an item is selected with Enter.

jongund:
@carmacleod  Carolyn I fixed the problem of the listbox not opening

carmacleod:
Thanks, @jongund !

One more weird case, only in the [autocomplete=both example](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-both.html):
- tab to the combo
- type down arrow a few times
- type Enter
- now type down arrow - nothing happens

I was expecting it to behave the same as the [autocomplete=list example](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-list.html), and open the popup even though there's only one item in it.

jongund:
@carmacleod I made some updates to fix the down arrow bug you found, please test

examples/combobox/combobox-autocomplete-both.html
```diff
@@ -54,7 +52,18 @@ <h2 id="ex_label">Example</h2>
           <div class="group">
             <input id="cb1-input" class="cb_edit" type="text" role="combobox" aria-autocomplete="both"
               aria-expanded="false" aria-controls="lb1">
-            <button type="button" id="cb1-button" aria-label="Open" tabindex="-1">&#9660;</button>
+            <button type="button" id="cb1-button" aria-label="Open" tabindex="-1">
+              <span class="arrow up">
+                <svg width="18" height="18">
```
zcorpan: I think it's better to use a single SVG and instead use CSS `transform: rotate(180deg)` to change between up arrow and down arrow.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -30,14 +30,38 @@
   width: 1.25rem;
   border-left: none;
   outline: none;
-  font-size: 70%;
+  position: relative;
+}
+
+.combobox .group button .arrow {
+  top:;
```
zcorpan: Remove this rule?

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,545 @@
...
+  return option;
+};
+
+ComboboxAutocomplete.prototype.setCurrentOptionStyle = function (option) {
+
+  for (var i = 0; i < this.filteredOptions.length; i++) {
+    var opt = this.filteredOptions[i];
+    if (opt === option) {
+      opt.setAttribute('aria-selected', 'true');
+      this.listboxNode.scrollTop = opt.offsetTop;
```
zcorpan: I think this was pointed out as being unexpected (point 2 in https://github.com/w3c/aria-practices/issues/1266 ). (That can be fixed separately, just noting it here.)

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -58,18 +82,21 @@ ul[role="listbox"] {
 }
 
 ul[role="listbox"] li[role="option"] {
-  border-top: 1px solid transparent;
-  border-bottom: 1px solid transparent;
   display: block;
+  padding-left: 2px;
```
zcorpan: The padding in the dropdown was intended to match the padding in the combobox (0.3em on left and right) so that the text would line up nicely.

zcorpan:
> Fixed other bugs related to previous code:
> Issue #1268: Combobox Examples: Popup incorrectly contains only Utah when Wyoming in input
> Issue #1265: Combobox examples: Make clicking outside popup close the popup

Very nice, the combobox feels much less buggy to use now!

I've reported a separate issue that I noticed while reviewing this: https://github.com/w3c/aria-practices/issues/1278

carmacleod:
> @carmacleod I made some updates to fix the down arrow bug you found, please test

Fixed - thank-you, Jon! Much better.

examples/combobox/combobox-autocomplete-both.html
jongund: I have removed one SVG image and used the transform property on the remaining image.

examples/combobox/js/combobox-autocomplete.js
jongund: @zcorpan Fixed the option scrolling issue

zcorpan:
The arrow doesn't appear centered vertically. It's better when the popup is open, though.

Adding `vertical-align: middle` to the `svg` helps, but then it doesn't look centered when the popup is open. Tweaking the transform slightly helps: `rotate(180deg) translate(0, 1px);`

jongund:
@smhigley  Thanks for finding the onclick bug for the `list` and `none` options, this should be fixed now.

carmacleod:
Hi @jongund !

Please fix up these few typos:
- in tests/combobox_autocomplete-both.js:
  - change 2 instances of "textboxed" to "textbox"
  - in `ariaTest('Test single escape key press with focus on textbox',`
    - change "clearedx" to "not cleared" (notice the "not")
  - in `ariaTest('Test double escape key press with focus on textbox',`
    - change "clearedx" to "cleared" (this time, it is cleared)
  - change "In key press" to "In listbox key press"

- tests/combobox_autocomplete-list.js:
  - change 2 instances of "textboxed" to "textbox"
  - change "In key press" to "In listbox key press"

- tests/combobox_autocomplete-none.js:
  - change 2 instances of "textboxed" to "textbox"

jongund:
@carmacleod I updated the test cases with the documentation changes.

examples/combobox/combobox-autocomplete-both.html
```diff
@@ -271,7 +271,7 @@ <h3 id="kbd_label_listbox">Listbox Popup</h3>
               <ul>
                 <li>Moves visual focus to the textbox.</li>
                 <li>Types the character in the textbox.</li>
-                <li>Filters options based on matching letters in the textbox.</li>
+                <li>Options in the listbox are filtered based on character in the textbox.</li>
```
carmacleod:
```suggestion
                <li>Options in the listbox are filtered based on characters in the textbox.</li>
```

examples/combobox/combobox-autocomplete-list.html
```diff
@@ -267,7 +267,7 @@ <h3 id="kbd_label_listbox">Listbox Popup</h3>
               <ul>
                 <li>Moves visual focus to the textbox.</li>
                 <li>Types the character in the textbox.</li>
-                <li>Filters options based on matching letters in the textbox.</li>
+                <li>Options in the listbox are filtered based on character in the textbox.</li>
```
carmacleod:
```suggestion
                <li>Options in the listbox are filtered based on characters in the textbox.</li>
```

carmacleod:
@jongund The autocomplete=both and autocomplete=list examples work well now!

However, I had a small problem with the autocomplete=none example.
First, a question. The example says:
> Options in the listbox are not filtered based on the characters in the textbox.

But the list filters if you select something, then tab away and tab back.
Is this intentional? I think a non-autocomplete list should always stay unfiltered, so that people never wonder where the rest of the items went?

Here's the detailed steps to see this:
- [autocomplete=none combobox example](https://raw.githack.com/w3c/aria-practices/issue1066-combobox-fix-escape-key-and-other-bugs/examples/combobox/combobox-autocomplete-none.html)
- tab to the combobox
- type **Down arrow** and then **Enter** to select "weather" in the combobox
- type **Tab** and then **Shift+Tab** to blur and then re-focus the combobox
- now type **Down arrow** again, and the list only contains "weather"?

I was expecting to see the complete list.

There's something strange going on, because if you now type **Esc** twice (to close the list and clear the textbox), and then type **Down arrow** again (to open the list), the list will still only contain "weather".

jongund:
@carmacleod Thanks for finding the bug, can you test autocomplete again

carmacleod:
Thanks for the fix @jongund !

I noticed one more thing - sorry I didn't notice before.
There's a sentence at the top of the example page (for autocomplete=none) that says:
> In this implementation, The listbox popup is not automatically triggered; it is displayed only when the user opens it.

However, if I tab to the combobox and type a letter, the popup is automatically opened.

So should the behavior be changed to match the description, or should the description be changed to match the behavior? :)

jongund:
@carmacleod I mad an edit to the description of when the combobox opens, please review.

examples/combobox/combobox-autocomplete-none.html
```diff
@@ -32,7 +32,7 @@ <h1>Editable Combobox without Autocomplete Example</h1>
       The design pattern describes four types of autocomplete behavior.
       This example illustrates the autocomplete behavior known as <q>no autocomplete</q>.
       The terms that appear in the listbox popup are not related to the string that is present in the textbox.
-      In this implementation, The listbox popup is not automatically triggered; it is displayed only when the user opens it.
+      In this implementation, The listbox popup is not automatically triggered when the textbox receives focus; the list is opened when the user types a character into the textbox or through an explicit open command.
```
carmacleod:
```suggestion
      In this implementation, the listbox popup is not automatically triggered when the textbox receives focus; the list is opened when the user types a character into the textbox or through an explicit open command.
```

mcking65:
Does this PR address #1261?

test/tests/combobox_autocomplete-none.js
```diff
@@ -178,6 +200,9 @@ ariaTest('Test down key press with focus on textbox',
     // Check that the active descendent focus is correct
     await assertAriaSelectedAndActivedescendant(t, ex.textboxSelector, ex.optionsSelector, 0);
 
+   // Check if aria-selected is set on first item
+    await assertAttributeValues(t, ex.optionsSelector + ':nth-of-type(1)', 'aria-selected', 'true');
```
spectranaut: @jongund -- aria-selected is actually checked in assertAriaSelectedAndActivedescendant, the line above! So you don't need to add this line.

test/tests/combobox_autocomplete-both.js
```diff
@@ -148,9 +149,27 @@ ariaTest('"aria-selected" attribute on options element', exampleFile, 'option-ar
...
+    await t.context.session
+      .findElement(By.css(ex.textboxSelector))
+      .sendKeys(Key.ALT, Key.ARROW_DOWN);
+
+    // Check that the listbox is displayed
+    t.true(
+      await t.context.session.findElement(By.css(ex.listboxSelector)).isDisplayed(),
+      'In example the list box should display after ALT + ARROW_DOWN keypress'
+    );
+
```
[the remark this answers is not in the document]
jongund: @spectranaut 

I added the `assertAttributeNDE(t, ex.optionsSelector, 'aria-selected')` option and added `assertAttributeValues(t, ex.optionsSelector + ':nth-of-type(1)', 'aria-selected', 'true')` to the tests for `textbox-key-down-arrow`.   The focus styling uses `aria-selected` as a selector.

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+      opt.removeAttribute('aria-selected');
+    }
+  }
+};
+
+ComboboxAutocomplete.prototype.getPreviousOption = function (currentOption) {
+  var index;
+
+  if (currentOption !== this.firstOption) {
+    index = this.filteredOptions.indexOf(currentOption);
```
smhigley: `index` can be declared here, since it's only used within the `if` block

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+    return this.filteredOptions[index - 1];
+  }
+  return this.lastOption;
+};
+
+ComboboxAutocomplete.prototype.getNextOption = function (currentOption) {
+  var index;
+
+  if (currentOption !== this.lastOption) {
+    index = this.filteredOptions.indexOf(currentOption);
```
smhigley: same as above

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+};
+
+ComboboxAutocomplete.prototype.handleComboboxKeyUp = function (event) {
+  var flag = false,
+    option = false,
+    char = event.key;
+
+  function isPrintableCharacter (str) {
```
smhigley: it might make sense to define this outside of the `keyup` function

examples/combobox/combobox-autocomplete-both.html
```diff
@@ -54,7 +52,13 @@ <h2 id="ex_label">Example</h2>
           <div class="group">
             <input id="cb1-input" class="cb_edit" type="text" role="combobox" aria-autocomplete="both"
               aria-expanded="false" aria-controls="lb1">
-            <button type="button" id="cb1-button" aria-label="Open" tabindex="-1">&#9660;</button>
+            <button type="button" id="cb1-button" aria-label="Open" tabindex="-1">
+              <span class="arrow">
+                <svg width="18" height="16">
```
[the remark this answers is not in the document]
zcorpan: Done. I also added `focusable="false"` per https://github.com/w3c/aria-practices/pull/1045#discussion_r320559735 -- though, possibly this SVG should be in the CSS instead, with generated content?

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -30,14 +31,33 @@
   width: 1.25rem;
   border-left: none;
   outline: none;
-  font-size: 70%;
+}
+
+.combobox .group button.open svg {
```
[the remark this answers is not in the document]
zcorpan: Why?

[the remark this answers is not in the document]
smhigley: It's a common best practice for efficiency (browser go right-to-left in interpreting CSS selectors), and to avoid an arms race of over-specific CSS selectors.

for reference:
- https://primer.style/css/principles/scss#lint-rules-wip
- https://csswizardry.com/2011/09/writing-efficient-css-selectors/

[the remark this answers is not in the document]
zcorpan: OK, thanks. For efficiency, I think it's better to use child selector over descendant selector, but that would add even more specificity and result in harder to read selectors. I think the selectors here can remove some middle part without changing what matches, so I'll try that.

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+
+  // Traverse the element children of domNode: configure each with
+  // option role behavior and store reference in.options array.
+  var nodes = this.listboxNode.getElementsByTagName('LI');
+
+  for (var i = 0; i < nodes.length; i++) {
+    var node = nodes[i];
+    this.allOptions.push(node);
+
+    node.textComparison = node.textContent.toLowerCase();
```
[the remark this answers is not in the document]
zcorpan: I agree.

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+  }
+  else {
+    this.comboboxNode.setAttribute('aria-activedescendant', '');
+  }
+};
+
+ComboboxAutocomplete.prototype.setValue = function (value) {
+  this.filter = value;
+  this.comboboxNode.value = this.filter;
+  this.comboboxNode.setSelectionRange(this.filter.length,this.filter.length);
```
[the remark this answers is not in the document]
zcorpan: From what I can tell, it's mostly for when you autocomplete with tab or enter. I don't know why `setValue` is called at all for backspace. Removing that call for backspace appears to remove the odd behavior of moving the cursor, so I've done that.
511a7077773fd358708c664f08a40ba6ea120bdf

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+          if (option) {
+            if (this.isClosed() && this.comboboxNode.value.length) {
+              this.open();
+            }
+
+            if (option.textComparison.indexOf(this.comboboxNode.value.toLowerCase()) === 0) {
+              this.option = option;
+              if (this.isBoth || this.listboxHasVisualFocus) {
+                this.setCurrentOptionStyle(option);
+                if (this.isBoth && isPrintableCharacter(char)) {
```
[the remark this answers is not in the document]
zcorpan: Fixed in https://github.com/w3c/aria-practices/pull/1276/commits/4e42a829d1808f491da0d175bed5a4268c1978ad

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+  var filter = this.filter.toLowerCase();
+
+  this.filteredOptions    = [];
+  this.listboxNode.innerHTML = '';
+
+  for (var i = 0; i < this.allOptions.length; i++) {
+    var option = this.allOptions[i];
+    if (filter.length === 0 || option.textComparison.indexOf(filter) === 0) {
+      this.filteredOptions.push(option);
+      var textContent = option.textContent.trim();
```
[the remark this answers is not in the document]
zcorpan: Fixed in 755416208a698c98e7e0f4857079943ce3f3b561

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  // Use populated options array to initialize firstOption and lastOption.
+  var numItems = this.filteredOptions.length;
+  if (numItems > 0) {
+    this.firstOption = this.filteredOptions[0];
+    this.lastOption  = this.filteredOptions[numItems - 1];
+
+    if (currentOption && this.filteredOptions.indexOf(currentOption) >= 0) {
+      option = currentOption;
```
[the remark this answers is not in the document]
zcorpan: I'll let @jongund speak to the intent.

I have however done as you suggested and declared `option` earlier. Also, the usage of `false` for this and `firstOption`/`lastOption` doesn't make sense to me (and was inconsistently sometimes using `null`), so I changed them from `false` to `null`.
9614aff6968f8dabc932c30fb0b9797d6ac0a9b1

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -0,0 +1,558 @@
...
+  for (var i = 0; i < this.filteredOptions.length; i++) {
+    var opt = this.filteredOptions[i];
+    if (opt === option) {
+      opt.setAttribute('aria-selected', 'true');
+      while ((this.listboxNode.scrollTop + this.listboxNode.offsetHeight) < (opt.offsetTop + opt.offsetHeight)) {
+        this.listboxNode.scrollTop = this.listboxNode.scrollTop + opt.offsetHeight;
+      }
+      while (this.listboxNode.scrollTop > (opt.offsetTop + 2)) {
+        this.listboxNode.scrollTop = this.listboxNode.scrollTop - opt.offsetHeight;
+      }
```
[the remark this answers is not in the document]
zcorpan: Fixed in 28f0ebd24df2e5103d535ee2b08696a2dd33988f

examples/combobox/js/combobox-autocomplete.js
zcorpan: Fixed in 5be02daa88b9c402a9a31bdca5ef95c3b03956ab

examples/combobox/js/combobox-autocomplete.js
zcorpan: Fixed

examples/combobox/js/combobox-autocomplete.js
zcorpan: Fixed in 2a5719e65bbd45bb8ae6ad05ac663051b8ed425f

zcorpan:
Thank you @smhigley, I believe I have addressed all comments.

ZoeBijl:
### Accessibility Review

Notes per screen reader and browser combo.

#### Narrator with Edge 44.1 on Windows 10 (1909)

The comboboxes are announced correctly ("State, combo-edit", "Search, combo-edit, suggestions available"). But I can't select anything past the first suggestion. Not by pressing down arrow only. If there's another way to interact with comboboxes I'd like to learn about that.

### Visual Review

In high contrast mode the drop down arrow disappears when the field is selected:

Selected combobox with missing arrow

sinabahram:
Note that you can use alt+downArrow to expand the combobox on Windows, if it’s expandable. Mac does this automatically, at least visually.

ZoeBijl:
Thanks @sinabahram! This works in both Safari and Edge with VoiceOver and Narrator respectively.

sinabahram:
You’re welcome, but please note, it is very much a work-around/advanced-user feature, and never should be relied upon. Comboboxes are not accessible if that’s the only way to get at their elements. It does come in handy, though, especially when debugging certain behaviors.

ZoeBijl:
Thanks. I've updated my two previous comments quite extensively after your comment. Out of interest: how is GitHub about communicating those changes? If at all.

sinabahram:
Good question. It doesn’t over email, so I’d have to brows on the site. Looking on the site, at the absolute end of the heading announcement, and incorrectly marked up with inaccurate menu semantics (already fixing that with them) is “edited”, so if I know to listen for it, then I can at least check if something is edited, but it’s hardly surfaced as a primary thing. Call it a 3 out of 10 in terms of usefulness 😊.

 

The takeaway is that it’s most appreciated that you pointed out you edited something, because then I know to go look for it, so thank you for that.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -16,28 +16,48 @@
...
   outline-offset: 1px;
 }
 
+.combobox .group.focus input,
+.combobox .group.focus button {
+  background-color: #DEF;
+}
+
+.combobox polygon {
+  fill: gray;
```
ZoeBijl: change this to `fill: currentColor;` to fix High Contrast Mode.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -16,28 +16,48 @@
...
+  stroke: gray;
```
ZoeBijl: change this to `stroke: currentColor;` to fix High Contrast Mode.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -16,28 +16,48 @@
...
+}
+
+.combobox button.open polygon,
+.combobox .group.focus polygon {
+  fill: black;
```
ZoeBijl: change this to `fill: currentColor;` to fix High Contrast Mode.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -16,28 +16,48 @@
...
+  stroke: black;
```
ZoeBijl: change this to `stroke: currentColor;` to fix High Contrast Mode.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -16,28 +16,48 @@
...
+  background-color: #DEF;
+}
+
+.combobox polygon {
+  fill: gray;
+  stroke: gray;
+}
+
+.combobox button.open polygon,
+.combobox .group.focus polygon {
```
ZoeBijl: Add `color: black;` to fix High Contrast Mode.

examples/combobox/css/combobox-autocomplete.css
```diff
@@ -16,28 +16,48 @@
...
+.combobox input {
   width: 10.75rem;
   border-right: none;
   outline: none;
   font-size: 87.5%;
   padding: 0.1em 0.3em;
 }
 
-.combobox .group button {
+.combobox button {
```
ZoeBijl: Add `color: gray;` to fix High Contrast Mode.

sinabahram:
I finally got a chance to play with this ... Jaws 2020, latest, and Chrome latest.

I tried the first preview link "both", and I came across several issues. but not sure if this should be tested in Chrome+Jaws yet?

1: the aria-expanded does not seem to be surfaced? I hit the open button, but it never says it's collapsed nor does it announce being in an expanded state. It does announce the transition between collapsed/expanded, which is super odd, but I didn't spend time to figure out if that was coming from an aria-live message on the page or from some native announcement of aria-expanded state change.

2: the button opens a listbox, which allows for no interaction. Instead, when one focuses it, it just kicks you out of forms mode ... I tried lots of times.

3: In the combobox, several things jumped out at me.
	A: The values wrap. That was unexpected, but I don't think I have a problem with this, especially since it can reduce the time it takes to find something even if all you can do is iterate previous/next, so I'm fine with that, but just noting that it was super-unexpected behavior.

	B: Typing a letter seems to be a one-way path. Once I type 'N", I don't see any way of removing that from my filter. The character model is weirder than that, though, because it doesn't replace the value with 'n', or do what I think almost any screen reader user would expect, which is to be placed on a reduced list of items starting with the letter 'n'. Instead it adds 'n' to the end of whatever string I am on, but then if I use my arrow keys, seems to filter appropriately by starting with 'n'. 

	C: I get zero automatic announcements of any found results or other type of dynamic information when typing or backspacing.

	D: Hitting backspace just erases the last letter typed or the last letter of whatever item is currently focused, but clearly not selected, since if it were selected, it would erase the whole term.

	E: It can lose state sometimes, like I backspaced over a word until I got to a subset of that place's name which I know appears in other states e.g. I backspaced over Northern Mariana Islands, until I had "north", but then when I hit up/down, I didn't get North Carolina.

4: The button says "open", but open what? That feels a bit like "read more" to me, where more specificity is required.

Anyways, I'm going to stop here. There are other issues as well. I'm actually somewhat concerned that I'm going down a rabbit hole because this is not meant to be tested in Chrome+Jaws yet?

If not, sorry for wasting time, but if so, I think a detailed screen reader treatment may be in order here.

mcking65:
While there are still quite a few bugs, the bugs targeted by this PR are resolved. I am merging. Then, I will respond to all the recent feedback and create separate issues as needed.

Thank you everyone for all the work on this!

mcking65:
@sinabahram wrote:

> I finally got a chance to play with this ... Jaws 2020, latest, and Chrome latest.
> 
> I tried the first preview link "both", and I came across several issues. but not sure if this should be tested in Chrome+Jaws yet?

Yes, it is ready for testing. Bugs can fall into 3 buckets: 1) APG example code, 2) browser, and 3) screen reader. If you haven't yet read it, the section of the APG titled [Read Me First](https://w3c.github.io/aria-practices/#read_me_first) explains the scope of practices in this regard. In particular, section 2.2 addresses browser and assistive technology support.

From a W3C perspective, browser bugs are the concern of the ARIA Working Group and screen reader bugs are the concern of the ARIA-AT community group.

> 1: the aria-expanded does not seem to be surfaced? I hit the open button, but it never says it's collapsed nor does it announce being in an expanded state. It does announce the transition between collapsed/expanded, which is super odd, but I didn't spend time to figure out if that was coming from an aria-live message on the page or from some native announcement of aria-expanded state change.

Correct, there is a JAWS bug because JAWS does not announce the state of the combobox when you navigate to it or read it.

When you click open with the virtual cursor, the mouse click event is causing DOM focus to move to the text input. JAWS announces the focus change event, then the expanded state changes, and  JAWS announces the state change event.

When the combobox is expanded, and you move the reading cursor back to the open button and click it again, the DOM focus is still on the text input. Sso, JAWS is announcing the state change from expanded to collapsed.

The APG examples do not use live regions to compensate for anything that could be categorized as lack of screen reader support. Information about the state of the combobox, how many suggested values are available in a combobox popup, which suggestion is automatically selected if any, etc. is all available in the accessibility tree for any comboboxes that correctly implement the pattern. What assistive technologies do with that information is an assistive technology design decision. However, the aria-at project is seeking to build consensus around basic expectations for assistive technology behavior.

I am writing combobox tests for the aria-at project right now. Combobox testing is part of the ARIA-AT project production system design phase, so those tests will be among the first we review with screen reader developers. ARIA-AT will have must have, should have, and nice to have expectations for screen readers. Announcements about suggestion availability may fall into one of the latter 2 buckets.

> 2: the button opens a listbox, which allows for no interaction. Instead, when one focuses it, it just kicks you out of forms mode ... I tried lots of times.

This is an issue that the ARIA-AT project should highlight for Freedom Scientific. To activate forms mode for a listbox, JAWS issues MSAA default action. In this case, that will generate a click event. Clicking an item in the listbox will close the listbox and put the clicked value into the combobox. Ordinarily, JAWS users will navigate the listbox by pressing down arrow from within the combobox. However, ideally, users would be able to operate the combobox in either reading mode or forms mode. I don't think there is anything we can do about this in the example code.

> 3: In the combobox, several things jumped out at me.
> A: The values wrap. That was unexpected, but I don't think I have a problem with this, especially since it can reduce the time it takes to find something even if all you can do is iterate previous/next, so I'm fine with that, but just noting that it was super-unexpected behavior.

We do this because home and end move the caret in the edit field. BTW, do you think the edit field should be in the up and down arrow navigation? Or, should it keep focus in the list?

> B: Typing a letter seems to be a one-way path. Once I type 'N", I don't see any way of removing that from my filter. The character model is weirder than that, though, because it doesn't replace the value with 'n', or do what I think almost any screen reader user would expect, which is to be placed on a reduced list of items starting with the letter 'n'. Instead it adds 'n' to the end of whatever string I am on, but then if I use my arrow keys, seems to filter appropriately by starting with 'n'. 

It looks like we still have some serious bugs related to typing, especially after deleting characters from the input. The suggestion list is borked. Thank you for pointing this out. I raised #1332.

> C: I get zero automatic announcements of any found results or other type of dynamic information when typing or backspacing.

Yap, and this is considered to be a screen reader shortcoming per the above. We will address this in the ARIA-AT project.

> D: Hitting backspace just erases the last letter typed or the last letter of whatever item is currently focused, but clearly not selected, since if it were selected, it would erase the whole term.

In the autocomplete both example, the first press of backspace will erase the autocompletion string -- the automatically suggested portion of the text. The second press would delete the last character typed. Compare tests 1 and 2 for the autocomplete both example with tests 3 and 4 for the autocomplete list example in issue #1332.

> E: It can lose state sometimes, like I backspaced over a word until I got to a subset of that place's name which I know appears in other states e.g. I backspaced over Northern Mariana Islands, until I had "north", but then when I hit up/down, I didn't get North Carolina.

This is addressed by #1332.

> 4: The button says "open", but open what? That feels a bit like "read more" to me, where more specificity is required.

You are right on this. It is worse than that because when the combobox is expanded, nothing about the button changes. For compatibility with touch-based screen readers and reading cursors, this button should be marked up more like a menu button with aria-expanded and aria-controls. I'm surprised that we didn't consider this already. Again, thank you for calling attention to it. I have opened #1333.

Excellent feedback. Thank you!!

sinabahram:
Thanks for this amazing response. Some answers below.

 

You said

Announcements about suggestion availability may fall into one of the latter 2 buckets.

 

SB:

I hope not. That’s super-concerning to me, just to be frank. I hope we don’t start with limited expectations because they will only go down from there in practice, or at least the past 20+ years of having to depend on assistive technologies as informed me. If there are suggestions, this needs to be surfaced, full stop. But, that’s my opinion, so there you have it. My criteria is very simple, can someone who doesn’t depend on assistive technology and someone who does both achieve all the same results in an equitable way. Let me know if that criteria is either weaker than or far stronger than the expectations of the folks working on this, so I can calibrate appropriately. Note, I won’t actually back down from this position, but at least I’ll understand that other people are not there and attempt to kindly and respectfully convince instead of assume.

 

You asked:

We do this because home and end move the caret in the edit field. BTW, do you think the edit field should be in the up and down arrow navigation? Or, should it keep focus in the list?

 

SB:

This is an interesting question. I think we have a mixed metaphor happening here because one can iterate with up/down; one can type but not for type-ahead, instead it does filtering based on a starts-with approach or so it seems to be intended; and one can also backspace to remove the filter once that bug gets fixed, but at that point, the backspacing should only affect what has been typed, not what the value is, right?

 

So, before we talk about up/down, maybe we should talk about left/right? I’ve always been partial to the idea, which is in my opinion almost an unwritten standard approach, which is that you type characters into an edit field or editable combobox, depending on phrasing, and then all that is there is what you have typed in, but when you hit up/down then it highlights what is found, but if you backspace or use left/right, you are only exploring what has been typed so far, because we know with 100% certainty that this will match the user’s expectation of what is there, since they typed it. Once space or enter or what not is used to choose an item, then of course the previous rule can be updated to include that which has been selected.

 

To that end, I definitely don’t like this super-confusing partial selection thing going on, where the user has to pay such close attention to what is selected VS focused. You and I have complete mastery over such things, and we probably still make mistakes on this stuff, but I argue 90% of screen reader users do not at all. I know why it’s there, but the idea that backspace behaves differently the very first time is a huge flag for me on usability.

 

Here’s my proposed approach:

You get this list of states, and then you type ‘North’ or something like that. You are informed that there are two or three options, and you can arrow up/down on them, but if you hit backspace, you just erase the ‘h’ from ‘north’, nothing else. Once you hit enter, you have now selected that state. Allowing people to edit state names seems highly irregular to me, because presumably the whole point of having a dropdown is actually not for the human’s benefit, right? it’s so that we can force people to  choose “North Carolina” in our data model instead of also having NC and nc and “north Carolina” and all those other variants in our eventual DB. So, why allow for the editing of these values at all?

 

Here is how I think of this visually.

 

Imagine an edit field with ‘North” and a blinking cursor after the ‘h’, but also a dropdown below it that filters based on the values of the above, therefore, it is super-clear that backspace, home, end, left, and right, as well as all character-input should affect the edit field, but ‘up/down’ should navigate within the list below.

 

That’s my mental model of such things, anyways.

 

If you absolutely must allow for editing because we also wish to make that pattern accessible, which I consider a completely different pattern, by the way, then I would suggest an affordance of space or something like that to select without closing e.g. space VS enter, and go at it that way, but I feel this would require standard help text to also be augmented, and this introduces interesting edge cases like entries with spaces in them.

 

hope this helps.

smhigley:
@sinabahram there are probably diminishing returns in debating combobox behavior in a merged PR thread :). If you feel like it's worth continued discussion, it would probably be better to open an issue.

That said, here are a couple brief notes in response to a couple of your points:

- I'm generally inclined to agree with some of the usability concerns around inline autocomplete, spacebar to select, and autoselection (I even ran a usability study around those, among other behaviors, with results here: https://www.24a11y.com/2019/select-your-poison-part-2/). That said, inline autocomplete specifically did not present much of a barrier to the keyboard or screen reader users who participated. Both spacebar and autoselect had far greater impacts on the usability of the control.
- Even if autocomplete with filtering performs worse in a usability study, that does not mean that there are no use cases where that behavior would be the right choice, or that those patterns should be forbidden. Given that they are valid use cases, it makes sense for us to provide examples in the ARIA Authoring Practices.

PR #1000, opened 2019-03-14
Guidelines for range related properties

spectranaut:
This PR is to get feedback on range related property section to add to the APG.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/aria-practices/pull/1000.html" title="Last updated on Aug 30, 2019, 1:54 PM UTC (d4a09ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1000/e653a7a...bocoup:d4a09ee.html" title="Last updated on Aug 30, 2019, 1:54 PM UTC (d4a09ee)">Diff</a>

PR #1279, opened 2019-12-03
Guidelines for range related properties

zcorpan:
Replaces https://github.com/w3c/aria-practices/pull/1000, changed to be a branch in this repo so it's easier to collaborate.

#### Preview Link

[View the new section the compare branch using RawGitHack](https://raw.githack.com/w3c/aria-practices/zcorpan/range-related-properties-2/aria-practices.html#range_related_properties)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1279.html" title="Last updated on Jul 7, 2020, 7:32 PM UTC (311e7be)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1279/33cb0ba...311e7be.html" title="Last updated on Jul 7, 2020, 7:32 PM UTC (311e7be)">Diff</a>


zcorpan:
@a11ydoer found the table for `aria-valuemin` and `aria-valuemax` confusing. Related to this, I've filed an issue for the ARIA spec about the `meter` role: https://github.com/w3c/aria/issues/1123

zcorpan:
I have also tried to clarify the table itself, by changing "none" to "no default", and the table heading "Required?" to "Are `aria-valuemin` and `aria-valuemax` required?".

zcorpan:
I'm not certain what to say about the purpose of `aria-valuemin` and `aria-valuemax` when `aria-valuetext` is used. Any suggestions?

jongund:
@mcking65 @zcorpan I like the battery example, but does it need to include `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` if the only information spoken is `aria-valuetext`? 

Also what is the guidance for some thing like a range of costs and aria-valuetext?
For example, a price range between $50-$100, should aria-valuetext be used to include `a aria-value text="75 dollars"`, instead of just `aria-valuenow="75"`?   Is either good? or is one better?

aria-practices.html
```diff
@@ -6279,6 +6279,285 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+  <section id="range_related_properties">
+    <h2>Communicating Value and Limits for Range Widgets</h2>
+    <p>
+      ARIA defines the following roles as range widgets, which means they communicate a value that is typically numeric and constrained to defined limits.
+    </p>
+    <ul>
+      <li><code>meter</code></li>
+      <li><code>progressbar</code></li>
+      <li><code>scrollbar</code></li>
+      <li><code>separator</code> (if the element is focusable)</li>
```
carmacleod: maybe just "(if focusable)"? (fewer words)

aria-practices.html
```diff
@@ -6279,6 +6279,285 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+          <th><code>aria-valuemax</code></th>
+          <td>Defines the maximum value allowed by a range widget.</td>
+        </tr>
+        <tr>
+          <th><code>aria-valuenow</code></th>
+          <td>Defines the current value of a range widget. This value is a number greater than or equal to <code>aria-valuemin</code> and less than or equal to <code>aria-valuemax</code> (if they are specified).</td>
+        </tr>
+        <tr>
+          <th><code>aria-valuetext</code></th>
+          <td>When a numeric value is not sufficiently descriptive, this property can define a description of the current value of a range widget.</td>
```
carmacleod: consider changing "When" to "If", and adding "textual" before "description"
```suggestion
          <td>If a numeric value is not sufficiently descriptive, this property can define a textual description of the current value of a range widget.</td>
```

zcorpan:
Thanks, @carmacleod, I've applied your suggestions.

@jongund I think your question is the topic of https://github.com/w3c/aria/issues/1128

jongund:
I like the table for default and required values, but I think it should also include aria-valuenow, and be formatted with two rows of headers to make it visually easier to read.

|  | Default Values ||| Require ||
Role | aria-valuemin | aria-valuemax | aria-valuenow | aria-valuemin | aria-valuemax | aria-valuenow |
----- | ---------------| ----------------| -----------------| --------------- | --------------- | ---------------- |

aria-practices.html
```diff
@@ -6421,6 +6421,305 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+    </section>
+
+    <section id="range_related_properties_meter_role">
+      <h3>Range properties with meter</h3>
+
+      <p>
+        A meter is a graphical display of a numeric value within a defined range. Therefore, the <code>aria-valuemin</code>, <code>aria-valuenow</code> and <code>aria-valuemax</code> properties are required for the <code>meter</code> role although there are no default values for <code>aria-valuemin</code> and <code>aria-valuemax</code>.
+      </p>
+      
+      <p>This is an example of a meter, which shows the change of Central Processing Unit(CPU) usage. </p>
```
carmacleod: The meter isn't really showing the _change_ of CPU usage; it's showing the _current_ CPU usage.
```suggestion
      <p>This is an example of a meter, which shows the current Central Processing Unit (CPU) usage. </p>
```

aria-practices.html
```diff
@@ -6421,6 +6421,305 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+        aria-valuetext="5%, 18 minutes remaining.">&lt;/span>
+&lt;/p></code></pre>
+      </div>
+    </section>
+
+    <section id="range_related_properties_meter_role">
+      <h3>Range properties with meter</h3>
+
+      <p>
+        A meter is a graphical display of a numeric value within a defined range. Therefore, the <code>aria-valuemin</code>, <code>aria-valuenow</code> and <code>aria-valuemax</code> properties are required for the <code>meter</code> role although there are no default values for <code>aria-valuemin</code> and <code>aria-valuemax</code>.
```
carmacleod: There's no default value for aria-valuenow, because it's required.

So I think we could change this to something like:
```suggestion
        A meter is a graphical display of a numeric value within a defined range. The <code>aria-valuemin</code>, <code>aria-valuenow</code> and <code>aria-valuemax</code> properties are required for the <code>meter</code> role, and they don't have default values.
```

aria-practices.html
```diff
@@ -6421,6 +6421,305 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+      </p>
+      
+      <p>This is an example of a meter, which shows the change of Central Processing Unit(CPU) usage. </p>
+
+      <pre><code>&lt;div role="meter"
```
carmacleod: Maybe this example should include the label, similar to the previous example?
Something like this, right before the div:
`<span id="cpu_usage_label">CPU usage</span>`

mcking65:
@a11ydoer, thank you for the meter section; it is looking good.

@carmacleod, thank you for the suggestions; they are all good as well.

I think the meter and spin button sections should have links back to the pattern like in slider.

I am curious if others would like to see the table of requirements by role expanded to include valuenow as Jon has suggested? I don't recall if there is a default calculation of valuenow for any of these roles. If not, we may not want a default column for valuenow. However, we could certainly have a required yes/no column for it.

carmacleod:
@mcking65 
> I am curious if others would like to see the table of requirements by role expanded to include valuenow as Jon has suggested?

Yes, I think that would make things clearer, and all in one place.

> I don't recall if there is a default calculation of valuenow for any of these roles.

I believe the only role that doesn't require valuenow is [spinbutton](https://w3c.github.io/aria/#spinbutton).

aria-practices.html
carmacleod:
```suggestion
      <pre><code>&lt;span id="cpu_usage_label">CPU usage&lt;/span>
&lt;div role="meter" 
```

carmacleod:
@mcking65 The latest commit does the following:
- restructure the table in 8.1 as suggested by @jongund (add valuenow; group columns under "Default Values" and "Required")
- collapse 8.2 (valuenow) into 8.1 (min & max) to make a single 8.1 for min, max and valuenow
- update `meter` based on PR https://github.com/w3c/aria/pull/1129 (min, max are supported and default to 0, 100; valuenow is required)

[Preview the new section 8.1](https://raw.githack.com/w3c/aria-practices/zcorpan/range-related-properties-2/aria-practices.html#range_related_properties_using_aria-valuemin_aria-valuemax_and_aria-valuenow).

@jnurthen - ping on https://github.com/w3c/aria/pull/1129 for 1.2 so that APG and ARIA 1.2 are in sync

carmacleod:
The table is correct - I triple-checked when I modified it. Agree that the prose does need to be updated in places to match the table.

Looking at the table again now, though, I wonder if it would be clearer to delete the 3 "Required" columns, and instead, just use the word "Required" in the valuenow cells, maybe with an asterisk footnote below the table reminding authors that they need to provide a value. I also wonder if we should have 2 rows for progressbar (one for indeterminate), or maybe use another asterisk footnote to explain that a value is needed unless it's indeterminate? Similarly, spinbutton needs valuenow set if it has a value, so maybe we should point that out?

For example, the "Default Values Table" could be rewritten like this:

Role | aria-valuemin | aria-valuemax | aria-valuenow
-- | -- | -- | --
meter | 0 | 100 | Required *
progressbar | 0 | 100 | None **
scrollbar | 0 | 100 | Required *
separator (if focusable) | 0 | 100 | Required *
slider | 0 | 100 | Required *
spinbutton | None | None | None ***

`*` authors need to provide a value
`**` authors need to provide a value unless the progressbar is indeterminate
`***` authors need to provide a value if the spinbutton has a value

Do people think this is clearer or more complicated? (note that it does contain more information in 4 columns than the current table contains in 7 columns...)

jongund:
How about something like this:


Role | aria-valuemin (default) | aria-valuemax (default) | aria-valuemin (required) | aria-valuemax (required) | aria-valuenow (required) 
-- | -- | -- | -- | -- | -- |
meter | 0 | 100 | When minimum value is not 0 | When maximum value is not 100 | current value
progressbar | 0 | 100  | When minimum value is not 0 | When maximum value is not 100 | current value, unless indeterminate 
scrollbar | 0 | 100  | When minimum value is not 0 | When maximum value is not 100 | current value
separator (if focusable) | 0 | 100  | When minimum value is not 0 | When maximum value is not 100 | current value
slider | 0 | 100  | When minimum value is not 0 | When maximum value is not 100 | current value
spinbutton | None | None  | minimum value if known | maximum value if known | Current value if known

carmacleod:
I like @jongund's latest table, except that I think the columns for aria-valuemin (required) and aria-valuemax (required) add complexity that is not strictly necessary, and could be confusing. Can we delete those two columns? I think authors know that if they don't want the attribute's default value, they need to set it to something else.

Not sure if "current value" for valuenow might be confusing? i.e. could be confused as follows:
"it's required, but it has a default of 'current value'?" Maybe use "authors must provide value" instead?

mcking65:
To wrap this up, we need commits to this branch that:

- [ ] Update table in section 8.1 to match [table in Jon's comment above](https://github.com/w3c/aria-practices/pull/1279#issuecomment-596641576)
- [ ] Edit text for min/max/now in sections 8.3 through 8.7 where the text does not match the requirements in the table.
- [ ] In the meter section (8.3), add a sentence with a link to the meter pattern similar to the sentence in the slider section that links to the slider pattern.
- [ ] In the spin button section (8.7), add a sentence with a link to the spin button pattern similar to the sentence in the slider section that links to the slider pattern.

a11ydoer:
@jongund @mcking65 @carmacleod I created PR for the range related property. I appreciate if you can have a look for final feeback. Thanks, Jemma

a11ydoer:
@jongund Thanks for your feedback. This branch was updated with your suggestion. Change 1 is very clearer than mine. 

The only modification I made is pH meter example. I used Alkaline pH level example so that it can have min value 7 and max value 14. I also change aria value now to 8 (ex:baking soda).  Value 1 (ex:lemon) is within acid pH level, which  1-7 min and max range. (a bit nerdy chemsitry detail ;-)

Thanks again.


```
<span id="**ph_alkaline_meter_label**">**Alkaline pH(Power of Hydrogen) Level**</span>
<div role="meter"
      **aria-valuenow="8"**
      aria-valuemin="7"
      aria-valuemax="14"
      aria-labelledby="ph_alkaline_meter_label">
</div>
```

a11ydoer:
any final suggestions, @carmacleod and @mcking65 ?

a11ydoer:
I hear you @jongund  Your edit is also consistent throughout examples. 

Let me just wait to hear more confirmation regarding **"Remove aria-valuemin and ariavaluemax from the source code example, since they use the default values."**
Thanks, 
Jemma

a11ydoer:
@jongund 

1. I reflected your editorial feedback to PR, [Preview](https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1279.html#range_related_properties_using_aria-valuemin_aria-valuemax_and_aria-valuenow)

2. One thing I did not change is removing default value of 100 for temperature(F) slider exmaple.  
I considered this as specific value for the temperature(F), not default value, 100.
>Changes for "Range properties with sliders"
NOTE: Remove ariavaluemax from the source code example, since it uses the default value of 100. 

@carmacleod 
Can you quickly check the modified table under [8.1](https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1279.html#range_related_properties_using_aria-valuemin_aria-valuemax_and_aria-valuenow)?

aria-practices.html
```diff
@@ -6424,6 +6424,327 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+            <td>No</td>
+            <td>Yes</td>
+          </tr>
+          <tr>
+            <th scope="row"><code>progressbar</code></th>
+            <td>0</td>
+            <td>100</td>
+            <td>No</td>
+            <td>No</td>
+            <td>Yes, if it is not exactly known.</td>
```
carmacleod, line 6520: aria-valuenow is required on progressbar if it is known. (otherwise, it's an indeterminate progressbar).

```suggestion
            <td>Yes, if known</td>
```

aria-practices.html
```diff
@@ -6424,6 +6424,327 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+&lt;/div&gt;</code></pre>
+
+
+    </section>
+
+    <section id="range_related_properties_slider_role">
+      <h3>Range properties with sliders</h3>
+      <p>
+        The <code>aria-valuemin</code> and <code>aria-valuemax</code> properties only need to be set for the <code>slider</code> role when the slider's minimum is not 0 or the maximum value is not 100.
+        The <code>aria-valuenow property is required for <code>slider</code> role and the author needs to make sure it is within the minimum and maximum values.
```
carmacleod, line 6684: Missing `</code>`.

```suggestion
        The <code>aria-valuenow</code> property is required for <code>slider</code> role and the author needs to make sure it is within the minimum and maximum values.
```

aria-practices.html
```diff
@@ -6424,6 +6424,327 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+      <p>
+        When the element's values are contained within a range but those values are not numeric (such as "small", "medium" and "large"),
+        or they are numeric but there is value in communicating more information than just a number,
+        <code>aria-valuetext</code> is used to surface the text value to assistive technologies.
+        Only use <code>aria-valuetext</code> when <code>aria-valuenow</code> is not sufficiently meaningful for users because using <code>aria-valuetext</code> will prevent assistive technologies from communicating <code>aria-valuenow</code>.
+      </p>
+
+      <div class="example">
+        <p>
+          For example, for a battery indicator, it would be useful to know how many minutes are remaining, in addition to how many percent charge is remaining.
```
carmacleod, line 6570:
```suggestion
          For example, for a battery indicator, it would be useful to know how many minutes are remaining, in addition to the percentage of charge remaining.
```

aria-practices.html
```diff
@@ -6424,6 +6424,327 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+        <code>aria-valuetext</code> is not be used as the number value in <code>aria-valuenow</code> is meaningful to the user.
+      </p>
+      <pre><code>&lt;div class="rail"&gt;
+  &lt;div id="thumb" role="slider" aria-valuemin="50" aria-valuenow="68" aria-valuemax="100"
+       aria-label="Temperature (F)" tabindex="0"&gt;
+  &lt;/div&gt;
+&lt;/div&gt;</code></pre>
+      <p>The slider example above can be made using the HTML <code>&lt;input type="range"></code> element.</p>
+      <pre><code>&lt;input type="range" min="50" value="68" max="100" aria-label="Temperature (F)"&gt;
+      </code></pre><p>The following example is a fan control. The <code>aria-valuenow</code> value is "1", which is not meaningful to the user. The assistive technology will surface the value of <code>aria-valuetext</code> ("low") instead.</p>
```
carmacleod, line 6698:
```suggestion
      </code></pre><p>The following example is a fan control. The <code>aria-valuenow</code> value is "1", which is not meaningful to the user. The <code>aria-valuetext</code> property is used so that assistive technology will surface its value ("low") instead.</p>
```

aria-practices.html
```diff
@@ -6424,6 +6424,327 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+&lt;/div&gt;</code></pre>
+    </section>
+
+    <section id="range_related_properties_spinbutton_role">
+      <h3>Range properties with spin buttons</h3>
+
+      <p>
+        The <code>aria-valuenow</code>, <code>aria-valuemin</code> and <code>aria-valuemax</code> properties are not required for the <code>spinbutton</code> role.
+        <code>aria-valuetext</code> can be used when appropriate. However, <code>aria-valuenow</code>, <code>aria-valuemin</code> and <code>aria-valuemax</code> are required if those values are known.
+        Detailed description of the <code>spin buttons</code> role can be found in the <a href="#spinbutton">spinbutton design pattern</a>.
```
carmacleod, line 6716:
```suggestion
        A detailed description of the <code>spinbutton</code> role can be found in the <a href="#spinbutton">spinbutton design pattern</a>.
```

aria-practices.html
```diff
@@ -6424,6 +6424,327 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+  Price per paperclip: $&lt;span id="price"&gt;0.50&lt;/span&gt;
+&lt;/div&gt;</code></pre>
+
+      <p>
+        The spin button example above can be made using the native HTML <code>&lt;input type="number"></code> element.
+      </p>
+
+      <pre><code>&lt;input type="number" min="0.01" value="0.5" max="2" aria-labelledby="paperclip-label"&gt;
+&lt;span id="paperclip-label"&gt;Price per paperclip&lt;/span&gt;:
+$&lt;output id="price" aria-labelledby="paperclip-label"&gt;0.50&lt;/output&gt;</code></pre>
```
carmacleod, line 6743: ^^ This example is a duplicate of the one below, except that this one is a bit buggy (it needs step, and having an output is a bit strange). I think this example should just be deleted, and the one below kept. Perhaps that was the intent.

carmacleod:
@a11ydoer 
> Can you quickly check the modified table under 8.1?

I checked the table, and pushed a fix for progressbar aria-valuenow (required if known).

I decided to push in a bunch of editorial changes - sorry about the email spam.

There are now only 5 "example-related" comments left in my review that I didn't want to push in without asking what was wanted:
1. Three html examples have a comment with `//` syntax, but html comment syntax is `<!-- ... -->`. Also, they're in the middle of a start tag, which I don't think is allowed (browsers probably parse each word in the comment as an empty attribute).
2. I'm not sure if the native progressbar example was intended to be the same as the aria example preceding it (i.e. indeterminate). It's not the same, so it either needs to be made the same, or the wording needs to be clearer that it's a different example that happens to be native.
3. There are 2 `input type="number"` examples, and I believe the first one was likely an experiment? It should probably just be deleted?

a11ydoer:
@carmacleod Your reviews are always welecomed.  @mcking65 and I talked about this issue two weeks ago and @mcking65 has the changes to add. I will also read through your comment tonight and respond as much as I can. (This afternoon is full of work meetings.)

mcking65:
@jongund, @a11ydoer, @carmacleod, I've completed final editorial review and revisions. I believe this is ready to merge. However, I'd like to have some approving reviews, especially from Jon since your last review was requesting changes before merge.

aria-practices.html
```diff
@@ -6425,6 +6425,313 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+        A detailed description of the <code>spinbutton</code> role can be found in the <a href="#spinbutton">spinbutton design pattern</a>.
+      </p>
+
+      <p>The following example sets the price of paperclips in cents.</p>
+
+      <pre><code>&lt;div role="spinbutton" aria-valuenow="50" tabindex="0"&gt;
+  &lt;button id="lower-price"&gt;Lower&lt;/button&gt;
+  &lt;button id="raise-price"&gt;Raise&lt;/button&gt;
+  Price per paperclip: $&lt;span id="price"&gt;0.50&lt;/span&gt;
+&lt;/div&gt;</code></pre>
```
carmacleod, line 6719: Spinbutton requires a name, so this example should show that.
The following suggested code wraps the "Price per paperclip: $" text in a span and points to the span's id with aria-labelledby from the spinbutton.

```suggestion
      <pre><code>&lt;div role="spinbutton" aria-labelledby="price-label" aria-valuenow="50" tabindex="0"&gt;
  &lt;button id="lower-price"&gt;Lower&lt;/button&gt;
  &lt;button id="raise-price"&gt;Raise&lt;/button&gt;
  &lt;span id="price-label"&gt;Price per paperclip: $&lt;/span&gt;&lt;span id="price"&gt;0.50&lt;/span&gt;
&lt;/div&gt;</code></pre>
```

aria-practices.html
```diff
@@ -6425,6 +6425,313 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+
+      <p>If there are minimum and maximum allowed values, set the <code>aria-valuemin</code> and <code>aria-valuemax</code> properties.</p>
+
+      <pre><code>&lt;div role="spinbutton" aria-valuemin="1" aria-valuenow="50" aria-valuemax="200" tabindex="0"&gt;
+  &lt;button id="lower-price"&gt;Lower&lt;/button&gt;
+  &lt;button id="raise-price"&gt;Raise&lt;/button&gt;
+  Price per paperclip: $&lt;span id="price"&gt;0.50&lt;/span&gt;
+&lt;/div&gt;</code></pre>
```
carmacleod, line 6727: Same as above - spinbutton requires a label.

```suggestion
      <pre><code>&lt;div role="spinbutton" aria-labelledby="price-label"
      aria-valuemin="1" aria-valuenow="50" aria-valuemax="200" tabindex="0"&gt;
  &lt;button id="lower-price"&gt;Lower&lt;/button&gt;
  &lt;button id="raise-price"&gt;Raise&lt;/button&gt;
  &lt;span id="price-label"&gt;Price per paperclip: $&lt;/span&gt;&lt;span id="price"&gt;0.50&lt;/span&gt;
&lt;/div&gt;</code></pre>
```

jongund:
@mcking65 

Reading the Section 8.1 on range guidance there seems to be a disconnect for ranges that are indeterminate and using default values for range attributes.  The guidance for indeterminate is to not include `aria-valuemax` or `aria-valuemin`, but we also say when one or both of these attributed are not included that the default values of 0 and 100 will be used.   So which is it, default values or indeterminate?    Maybe it should be when the `aria-valuenow` attribute is missing or empty it should be the way to author indeterminate.

aria-practices.html
jongund: @carmacleod Good catch!

mcking65:
@jongund wrote:

> In section 8.7 there is the following statement:
> "If there are minimum and maximum allowed values, set the aria-valuemin and aria-valuemax properties."
> 
> What should be done if there is no minimum or maximum values?
> Omitting the attributes gives you the default values of 0 and/or 100.

Section 8.7 says:

>The aria-valuemin and aria-valuemax properties are used only when a spinbutton has a defined range. They do not have default values, so if they are not specified, range limits will not be exposed to assistive technologies.

@jongund wrote:

> Are progress bars the only widget that can be indeterminate, it is the only one mentioned in the ARIA Spec?
> 
> If so maybe all the information related to indeterminate state should be moved to the progress bar section and removed form the other sections of the range widget.

Section 8.4 is about progressbar, and it says the following:

>The aria-valuemin and aria-valuemax properties only need to be set for the progressbar role when the progress bar's minimum is not 0 or the maximum value is not 100. The aria-valuenow property needs to be set for a progressbar if its value is known (e.g. not indeterminate). If the aria-valuenow property is set, the author needs to make sure it is within the minimum and maximum values.

So, while the current value may be indeterminate, the limits of the range are never indeterminate.

There are only 2 uses of the word "indeterminate", and both are in the progressbar section.

mcking65:
@jongund wrote:

> Reading the Section 8.1 on range guidance there seems to be a disconnect for ranges that are indeterminate and using default values for range attributes. The guidance for indeterminate is to not include aria-valuemax or aria-valuemin, but we also say when one or both of these attributed are not included that the default values of 0 and 100 will be used. So which is it, default values or indeterminate? Maybe it should be when the aria-valuenow attribute is missing or empty it should be the way to author identifies an indeterminate state.

The only widget that allows for unknown range is spinbutton. Even a progressbar that is indeterminate must have known limits. I reworded section 8.1 in commit 1985db4 to clarify. I hope this addresses your concern.

aria-practices.html
```diff
@@ -6425,6 +6425,316 @@ <h3>Indicating sort order with <code>aria-sort</code></h3>
...
+        For some widgets, assistive technologies use this information to present the current value as a percentage.
+        When using these properties, set <code>aria-valuemin</code> to the lowest value allowed for the widget and <code>aria-valuemax</code> to the highest allowed value.
+        If values for <code>aria-valuemin</code> and <code>aria-valuemax</code> are not set, default values are exposed to assistive technologies unless the widget is a <code>spinbutton</code>, which is the only range widget that does not have default limits.
+      </p>
+      <p>
+        The <code>aria-valuenow</code> property is used to inform assistive technologies of the current value of a range widget.
+        Set <code>aria-valuenow</code> to the current value of the widget, ensuring the value of <code>aria-valuenow</code> falls within the range defined by <code>aria-valuemin</code> and <code>aria-valuemax</code>.
+        If the current value of a <code>progressbar</code> or <code>spinbutton</code> is indeterminate or unknown, omit The <code>aria-valuenow</code> property.
```
carmacleod, line 6487: change capitalized The to lowercase the
```suggestion
        If the current value of a <code>progressbar</code> or <code>spinbutton</code> is indeterminate or unknown, omit the <code>aria-valuenow</code> property.
```

PR #1285, opened 2019-12-14
Windows Regression Test Fix

sh0ji:
This just moves the test globbing to [the Ava configuration](https://github.com/avajs/ava/blob/master/docs/06-configuration.md), which should solve issues where Windows wasn't able to find files.

Some background: npm `scripts` will run command line operations, leveraging whatever syntax features the operating system has in place, which results in ` test/tests/*.js does not exist.` on Windows. This change forces Ava to use [its own internal globbing algorithm](https://github.com/avajs/ava/blob/master/package.json#L91) ([globby](https://github.com/sindresorhus/globby)), which has cross-platform support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "lint:es": "eslint .",
     "lint:js": "npm run lint:es",
     "lint:html": "npm run vnu-jar",
-    "regression": "ava test/tests/*.js",
+    "regression": "ava",
     "regression-report": "node test/util/report",
     "test": "npm run lint && npm run regression",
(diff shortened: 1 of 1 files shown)


jongund:
Will this change force running all the tests?
Can you sill run individual test files like to just run the tests for carousel-2-tablist example:

`npm run regression -- --match *carousel-2*`

sh0ji:
That should still work, though I haven't tested it. You can also target specific files if you like, now that the npm script doesn't contain the file glob.

```sh
# windows path
npm run regression -- test\tests\carousel_carousel-1.js
```

jongund:
@spectranaut @sh0ji Evan thanks for this fix, it is working in Windows now.

Issue #760, opened 2018-07-09
File directory treeview example doesn't populate edit field correctly in FF


jkshapiro:
Steps to replicate:
1. Visit https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1b.html, using Firefox on Windows
1. Note that the "File or Folder Selected" field is blank.
1. Select the "Projects" folder in the example

Expected:
The "File or Folder Selected" field contains "Projects"

Observed
The "File or Folder Selected" field contains the following:
[...]

PR #1186, opened 2019-09-24
Slider patterns: Provide warning about inability to operate with touch-based assistive technologies due to lack of needed APIs

charmarkk:
Partially resolves issue #1150 by adding a warning to the slider patterns.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1186.html" title="Last updated on Jan 27, 2020, 6:30 AM UTC (766ce61)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1186/1a1ba9f...766ce61.html" title="Last updated on Jan 27, 2020, 6:30 AM UTC (766ce61)">Diff</a>

PR #1306, opened 2020-01-24
Tree Examples: Correct result of selecting parent node

carmacleod:
Fixes TreeView problem pointed out in PR https://github.com/w3c/aria-practices/pull/1186#issuecomment-577840796 :

> I have tested https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html with Talkback (Chrome) and VoiceOver (Safari) and the tree works well with both screenreaders on the smartphone without keyboard. The only problem: The hierarchy of the treeitems is not output.
>
> Note: The following small problem occurs in every browser: in the input field "File or Folder Selected" after selecting a folder not the name of the folder is displayed, but the entire HTML code including all child elements (`<span class="focus">Reports</span><ul role="group">`...)


mcking65:
This PR resolves #760.

Issue #1307, opened 2020-01-24
Navigation menubar example: Clicking menu item scrolls page to top in Chrome


JAWS-test:
When I click on a [main menu item](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html) in Chrome, the page is scrolled to the top of the page. Depending on the display size of my device, the menu is no longer visible. 

With Chrome on mobile device (Android) there is also the problem that after clicking on a menu item the font size of the menu items changes (they get much bigger).

Both problems do not occur with Safari, IE, Firefox.

Unfortunately I have not been able to find out the cause of the problems.

carmacleod:
The "scroll to top" sounds like a missing event.preventDefault() on a repurposed anchor element.
Does the problem happen for you with [menubar2](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-2/menubar-2.html)?

carmacleod:
I was able to replicate the "scroll to top" behavior and provide a fix.
Here's a [preview link](https://raw.githack.com/w3c/aria-practices/issue1307-scroll-to-top/examples/menubar/menubar-1/menubar-1.html) if you would like to try it.

JAWS-test:
> Does the problem happen for you with menubar2?

No. Thanks for fixing!

a11ydoer:
@mcking65 I think we can close tihs issue.

carmacleod:
@a11ydoer 
Is the Android problem fixed, too? I don't have an Android device. Here's what @JAWS-test said:
> With Chrome on mobile device (Android) there is also the problem that after clicking on a menu item the font size of the menu items changes (they get much bigger).

a11ydoer:
[Navigation Menubar Test URL](https://htmlpreview.github.io/?https://github.com/w3c/aria-practices/blob/master/examples/menubar/menubar-1/menubar-1.html)

jongund:
@a11ydoer 
This is the latest preview link that should be used for testing:
https://raw.githack.com/w3c/aria-practices/issue1357-navigation-menubar-js/examples/menubar/menubar-navigation.html 

I believe this issue was fixed along time ago, I don't see the reported behavior in the current example.

PR #1308, opened 2020-01-24
Navigation menubar: Fix scroll-to-top problem reported in issue #1307

carmacleod:
Issue #1307 reports 2 issues with the navigation menubar example. This PR resolves the scroll to top issue by preventing `menuitem` anchor elements from jumping to `href="#"` (i.e. top of page in Chrome) if they have a menu/submenu.

Here's a [preview link](https://raw.githack.com/w3c/aria-practices/issue1307-scroll-to-top/examples/menubar/menubar-1/menubar-1.html).

### Review checklist

* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) Sarah (@@smhigley)
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by Valerie (@spectranaut)
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) Valerie (@spectranaut)

### DIFF
--- a/examples/menubar/menubar-1/js/MenubarItemLinks.js
+++ b/examples/menubar/menubar-1/js/MenubarItemLinks.js
@@ -33,6 +33,7 @@ MenubarItem.prototype.init = function () {
   this.domNode.tabIndex = -1;
 
   this.domNode.addEventListener('keydown', this.handleKeydown.bind(this));
+  this.domNode.addEventListener('click', this.handleClick.bind(this));
   this.domNode.addEventListener('focus', this.handleFocus.bind(this));
   this.domNode.addEventListener('blur', this.handleBlur.bind(this));
   this.domNode.addEventListener('mouseover', this.handleMouseover.bind(this));
@@ -118,6 +119,13 @@ MenubarItem.prototype.handleKeydown = function (event) {
 
   if (flag) {
     event.stopPropagation();
+    event.preventDefault();
+  }
+};
+
+MenubarItem.prototype.handleClick = function (event) {
+  if (this.popupMenu) {
+    // prevent scroll to top of page when anchor element is clicked
     event.preventDefault();
   }
 };
--- a/examples/menubar/menubar-1/js/MenubarItemLinks.js
+++ b/examples/menubar/menubar-1/js/MenubarItemLinks.js
@@ -125,7 +125,8 @@ MenubarItem.prototype.handleKeydown = function (event) {
 
 MenubarItem.prototype.handleClick = function (event) {
   if (this.popupMenu) {
-    // prevent scroll to top of page when anchor element is clicked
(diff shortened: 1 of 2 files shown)


examples/menubar/menubar-1/js/MenubarItemLinks.js
```diff
@@ -122,6 +123,13 @@ MenubarItem.prototype.handleKeydown = function (event) {
   }
 };
 
+MenubarItem.prototype.handleClick = function (event) {
+  if (this.popupMenu) {
```
spectranaut: Curious why this file is necessary?

examples/menubar/menubar-1/js/PopupMenuItemLinks.js
```diff
@@ -173,6 +173,10 @@ MenuItem.prototype.setExpanded = function (value) {
 MenuItem.prototype.handleClick = function (event) {
   this.menu.setFocusToController();
   this.menu.close(true);
+  if (this.popupMenu) {
```
spectranaut: same q?

examples/menubar/menubar-1/js/MenubarItemLinks.js
carmacleod: Hi @spectranaut !

If the selected menuitem has a submenu, then `this.popupMenu` will be true.
Those menuitems are implemented with `<a>`, but we just want them to open their submenu (aka popupMenu) when they are clicked, so we need to prevent default anchor behavior on them.

The other type of menuitem in the navigation menubar really is an anchor (semantically as well as being implemented with `<a>`), so for those, we need to keep the default anchor behavior so that it will take the user to a new page (or to a fragment, or wherever) when clicked.

examples/menubar/menubar-1/js/PopupMenuItemLinks.js
carmacleod: If the selected menuitem opens a (sub) menu, then this.popupMenu will be true.
We want to suppress default link behavior for these guys.

So, basically the same answer as above, except that the menuitems here (in file PopupMenuItemLinks.js) are in a menu (as opposed to the ones in MenubarItemLinks.js, above, which are in the menubar).

carmacleod:
@spectranaut 
I improved the comment a bit in the 2 places with `if (this.popupMenu)`.
Please let me know if you think it's clearer now.

examples/menubar/menubar-1/js/PopupMenuItemLinks.js
spectranaut: got it! And thanks for adding the comments.

PR #1303, opened 2020-01-20
Landmark Examples: Correct tabs structure (issue #1301)

ZoeBijl:
Fixes #1301.

PR #1315, opened 2020-01-28
Infrastructure: Skip regression testing against landmark example files when they are edited

spectranaut:
This fixes the CI error in: https://github.com/w3c/aria-practices/pull/1303

The problem was that AVA is looking for landmark tests because the landmark directory is in the examples directory, and [this script](https://github.com/w3c/aria-practices/blob/master/scripts/regression-tests.sh) tries to run only the tests for the examples that have been edited.

I tested these changes on top of #1303 in this PR, where you can see the CI now passing.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/aria-practices/pull/1315.html" title="Last updated on Feb 3, 2020, 7:08 AM UTC (0486008)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1315/d75c28b...bocoup:0486008.html" title="Last updated on Feb 3, 2020, 7:08 AM UTC (0486008)">Diff</a>

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -2344,7 +2344,10 @@
         A slider is an input where the user selects a value from within a given range.
         Sliders typically have a slider thumb that can be moved along a bar or track to change the value of the slider.
       </p>
-
+      <p class="warning">
+        Users of touch-based assistive technologies may not be able to fully operate widgets that implement this pattern because APIs for capturing the necessary gestures and commands from assistive technologies are not yet available.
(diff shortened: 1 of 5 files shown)


spectranaut:
@nschonni can you review?

spectranaut:
@nschonni yeah I try to make bash scripts as simple as possible because even when they are simple, I find them hard to read and edit. If it becomes a list of things we have to ignore maybe we should revisit!

PR #1321, opened 2020-02-04
Spin Button Date Picker Example: Fix February bug

spectranaut:
This missing reference was uncovered because this test only fails in February -- for which the function getDaysInMonth had some special handling that lead to an error. Fixed it so the function doesn't error!

#### Preview Link

[View the spin button example in the compare branch](https://raw.githack.com/bocoup/aria-practices/spinbutton-fix/examples/spinbutton/datepicker-spinbuttons.html)

### DIFF
--- a/examples/spinbutton/js/datepicker-spinbuttons.js
+++ b/examples/spinbutton/js/datepicker-spinbuttons.js
@@ -52,42 +52,7 @@ DatePickerSpinButtons.prototype.init = function () {
 };
 
 DatePickerSpinButtons.prototype.getDaysInMonth = function (year, month) {
-
-  if (typeof year !== 'number') {
-    year = this.year;
-  }
-
-  if (typeof month !== 'number') {
-    month = this.month;
-  }
-
-  switch (month) {
-
-    case 0:
-    case 2:
-    case 4:
-    case 6:
-    case 7:
-    case 9:
-    case 11:
-      return 31;
-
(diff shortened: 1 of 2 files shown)


mcking65:
@spectranaut, I'm now able to set the date to february 31, 2020 in the [example](https://raw.githack.com/bocoup/aria-practices/spinbutton-fix/examples/spinbutton/datepicker-spinbuttons.html).

Was using Chrome on Windows.

Is ECMA creating new days for only those of who know about them to use ... I could use a few extra. If this is not a valid secret plot by them, then we should do something about the code. I hope it is a valid plot but fear it is a bug.

spectranaut:
@mcking65 sorry to disappoint, Ecma didn't create any extra days in February, it was only an off by one error on my part! However, today is the EcmaScript meeting, so maybe we should introduce "extra days in February" as a feature request.

I just force pushed and tested the fix locally. Thanks for catching :)

mcking65:
OK, odd, odd, a few minutes ago this compare branch still had the Feb 31 problem. I refreshed my cache and even tried in two other browsers -- still had the problem. So, I tested locally by pulling directly from Valerie's fork. It tested correctly locally. I pushed that to a new branch, tested there and it worked. Now, I come back to this PR and test in its compare branch and the date picker is working fine Oh well.... happy for the fix!!! Time to merge!

PR #1328, opened 2020-02-13
#1325 rebased (Update stylelint to the latest version 🚀)

zcorpan:
This is #1325 rebased on latest master. I resolved the conflicts in `package.json` manually, and then ran `npm install` to re-generate `package-lock.json`. However, this appears to have caused some versions to decrease in `package-lock.json` -- for example, `browserlist` changed from 4.8.7 to 4.8.6.

I have the latest npm as far as I can tell (6.13.7).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -868,14 +868,14 @@
       }
     },
     "browserslist": {
-      "version": "4.8.7",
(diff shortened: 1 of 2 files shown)


zcorpan:
On the other hand, `package-lock.json` will get overridden on the next package change anyway, so maybe it doesn't matter much if there's a discrepancy here compared to #1325...

Issue #1266, opened 2019-11-14
Grid Combobox Example: Improve popup trigger and scrolling


mcking65:
Per review of combobox examples by @smhigley in #1255, the grid combobox has the following two undesireable behaviors:

1. It does not let you open the popup and see all options without first typing something.
2. The list scrolls every time you navigate to a new option, even if that option was already in view before the programmatic scroll.

Issue #860, opened 2018-08-27
 Bug in example combobox/aria1.1pattern/grid-combo.html: key "escape" does not close popup


spectranaut:
On the grid-combo example, the "Grid Popup" Keyboard Support table describes the "escape" key interaction as:
* Closes the grid popup.
* Sets focus on the textbox.
* Clears the textbox.

Presently, escape (while focus is on a gridcell) puts focus back on the textbox but does not close the popup or clear the textbox.

PR #1334, opened 2020-02-16
Combobox with grid popup: Make escape key behavior consistent with pattern

mcking65:
Fix issue #1066 for the grid combobox.

- [x] Update keyboard documentation to reflect expected behavior
- [x] Change JavaScript so Escape functions as documented
- [x] Revise tests to test documented behavior

#### Preview Link

[View the combobox with grid popup in compare branch of this PR](https://raw.githack.com/w3c/aria-practices/issue1066-for-grid/examples/combobox/grid-combo.html)

### Review checklist

* [x] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @carmacleod 
* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by Matt (@mcking65)
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by Simon (@zcorpan)
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by Simon (@zcorpan)


jongund:
@mcking65 

I have updated the escape key behavior and escape key regression tests to match the pattern.

zcorpan:
I've fixed the logic so that escape always hides the popup if the popup is visible, regardless of focus.

We were testing this case correctly. It's just that the test was marked as expected to fail, since this was a known issue: https://github.com/w3c/aria-practices/issues/860 

I've changed the test so it instead expects to pass, and it does.

zcorpan:
@spectranaut can you review my changes?

examples/combobox/js/grid-combo.js
```diff
@@ -94,12 +94,16 @@ aria.GridCombobox.prototype.handleInputKeyDown = function (evt) {
       );
     }
     else {
+      if (!this.shown) {
+        setTimeout((function () {
+          // On Firefox, input does not get cleared here unless wrapped in
```
zcorpan: Do we know why this happens? Is it a bug in Firefox?

zcorpan:
Looks like we're missing some tests. But that was the case before this PR, too, so I've filed an issue to track this: https://github.com/w3c/aria-practices/issues/1349

mcking65:
OK, now just waiting on a review by @spectranaut of the test changes made by @zcorpan. I will merge if @spectranaut approves.

examples/combobox/js/grid-combo.js
jongund: I think this is an issue of using keydown, instead of the keyup event.   I was only asked to fix the ESC behavior, so I didn't think it would be good to make more changes than requested.

carmacleod:
Should down arrow (and/or alt+down arrow) open the grid? (It currently doesn't, but I think it should).
It currently feels strange that you have to type an appropriate letter key to open the grid.

examples/combobox/js/grid-combo.js
mcking65: Thank you @jongund, that's good call. let's address that as a separate issue if there is a better way to resolve it than what is currently implemented.

mcking65:
@carmacleod commented:

> Should down arrow (and/or alt+down arrow) open the grid? (It currently doesn't, but I think it should).
> It currently feels strange that you have to type an appropriate letter key to open the grid.
> (If this comment would be better as a new issue, please let me know).

That is issue #1266.

carmacleod:
@mcking65 
> That is issue #1266.

Ah, right. Sorry about that. :) Thanks!

Issue #1332, opened 2020-02-16
Combobox examples with both and list autocomplete: Editing value in text input does not correctly update suggestion list


mcking65:
Editing the value in the combobox input is not correctly updating the listbox to have the correct set of suggested values in the combobox examples with list and both autocomplete:
1. [autocomplete = both example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-both.html)
2. [autocomplete = list example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html)

The following tests are failing.

#### Test 1
1.  Load the [combobox with autocomplete = both example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-both.html)
2. Focus the combobox and type "a".
3. Type backspace to delete the auto-highlighted text "labama"
4. Press backspace to delete the "a".
5. Press down arrow.

Expected result: There are 56 suggested values in the listbox and "Alabama" is selected.

Actual result: there are 5 suggested values in the listbox and "Alabama" is selected.

#### Test 2
1.  Load the [combobox with autocomplete = both example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-both.html)
2. Focus the combobox and type "no".
3. Press Backspace to delete the auto-highlighted text "rth Carolina"
4. Press backspace to delete the "o".
5. Press down arrow.

Expected result: There are 9 suggested values in the listbox and "Nebraska" is selected.

Actual result: there are 3 suggested values in the listbox and "North Carolina" is selected.

#### Test 3
1.  Load the [combobox with list autocomplete example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html).
2. Focus the combobox and type "a".
3. Press backspace to delete the "a".
4. Press down arrow.

Expected result: There are 56 suggested values in the listbox and "Alabama" is selected.

Actual result: there are 5 suggested values in the listbox and "Alabama" is selected.

#### Test 4
1.  Load the [combobox with list autocomplete example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html).
2. Focus the combobox and type "no".
3. Press backspace to delete the "o".
4. Press down arrow.

Expected result: There are 9 suggested values in the listbox and "Nebraska" is selected.

Actual result: there are 3 suggested values in the listbox and "North Carolina" is selected.

jongund:
@mcking65 

Just made a pull request to fix this bug.

PR #1335, opened 2020-02-16
Issue1332 listbox not updating

jongund:
@mcking65 

[Preview link for fix for listbox not properly updating using backspace (Issue 1332)](https://raw.githack.com/w3c/aria-practices/issue1332-listbox-not-updating/examples/combobox/combobox-autocomplete-both.html)

I added calls to the code that filters the options after the filter property is updated when backspace is pressed or if the length of the textbox value is less than the length of the filter property.

### DIFF
--- a/examples/combobox/js/combobox-autocomplete.js
+++ b/examples/combobox/js/combobox-autocomplete.js
@@ -391,6 +391,7 @@ ComboboxAutocomplete.prototype.handleComboboxKeyUp = function (event) {
   if (this.comboboxNode.value.length < this.filter.length) {
     this.filter = this.comboboxNode.value;
     this.option = null;
+    this.filterOptions()
   }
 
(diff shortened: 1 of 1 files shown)


mcking65:
@spectranaut, @smhigley, seems like it would be a good idea to have tests for this behavior as specified in issue #1332 and associated with data-test-id="standard-single-line-editing-keys". What do you think?

mcking65:
@smhigley, are you available for code and test review?  If we are going to change/add tests, we could break that out into a separate issue and PR to keep this one simple.

smhigley:
@mcking65 I agree, I think a test should probably be added for backspace correctly updating the filter, but that can be a separate issue/PR

mcking65:
Created #1345 to add regression testing for backspace.

PR #1343, opened 2020-02-24
Add queryElements test helper, remove t.plan

### LINKED ISSUE #1312: Helper function to eliminate need for t.plan()
See this comment for background: https://github.com/w3c/aria-practices/pull/1120/files#r334756143

In summary, `t.plan()` is annoying to maintain but helps prevent false positives during loops, Sarah suggested:

> we create a helper to get elements that also has an assertion that the returned elements array is not empty.
> 
> It would have two benefits: we don't need to worry about people forgetting to add t.plan(), and we also aren't testing the number of elements that exist in the DOM, as long as there's at least one. So, as an example, if we end up updating the tabs example to have one more tab, we don't then need to go update a bunch of t.plan()s in the tests.

smhigley:
Resolves #1312 

Adds `t.context.queryElements` helper function that wraps `t.context.session.findElements`, and fails the test if no elements are found.

Removes `t.plan()`, now that it isn't needed to ensure that tests are run.

Removes some unrelated unused imports caught while updating the above.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,10 @@
   },
   "rules": {
     "no-unused-vars": 0,
-    "no-undef": 0
+    "no-undef": 0,
+    "no-restricted-properties": [2, {
+      "property": "findElements",
+      "message": "Please use t.context.queryElements()."
+    }]
   }
 }
--- a/test/index.js
+++ b/test/index.js
@@ -88,7 +88,7 @@ const _ariaTest = (desc, page, testId, body, failing) => {
     if (testId !== 'test-additional-behavior') {
       const assert = require('assert');
       assert(
-        (await t.context.session.findElements(By.css(selector))).length,
+        (await t.context.queryElements(t, selector)).length,
         'Cannot find behavior description for this test in example page:' + testId
       );
     }
--- a/test/index.js
+++ b/test/index.js
@@ -6,6 +6,7 @@ const webdriver = require('selenium-webdriver');
 const { By } = require('selenium-webdriver');
 
 const startGeckodriver = require('./util/start-geckodriver');
+const queryElements = require('./util/queryElements');
 
 let session, geckodriver;
 const firefoxArgs = process.env.CI ? [ '-headless' ] : [];
@@ -30,6 +31,7 @@ if (!coverageReportRun) {
   test.beforeEach((t) => {
     t.context.session = session;
     t.context.waitTime = testWaitTime;
+    t.context.queryElements = queryElements;
   });
 
   test.after.always(() => {
--- a/test/index.js
+++ b/test/index.js
@@ -1,3 +1,5 @@
+/* eslint no-restricted-properties: 0 */
+
 'use strict';
 
 const path = require('path');
--- a/test/index.js
+++ b/test/index.js
@@ -1,5 +1,3 @@
-/* eslint no-restricted-properties: 0 */
-
 'use strict';
 
 const path = require('path');
--- a/test/tests/checkbox_checkbox-2.js
+++ b/test/tests/checkbox_checkbox-2.js
@@ -1,3 +1,5 @@
+/* eslint no-restricted-properties: 0 */
+
 'use strict';
 
 const { ariaTest } = require('..');
@@ -40,7 +42,7 @@ ariaTest('"tabindex" on checkbox element', exampleFile, 'checkbox-tabindex', asy
 });
 
 ariaTest('"aria-controls" ', exampleFile, 'checkbox-aria-controls', async (t) => {
-  
+
   const checkbox = await t.context.session.findElement(By.css(ex.checkboxSelector));
   const controls = (await checkbox.getAttribute('aria-controls')).split(' ');
 
@@ -60,7 +62,7 @@ ariaTest('"aria-controls" ', exampleFile, 'checkbox-aria-controls', async (t) =>
 });
 
 ariaTest('"aria-checked" on checkbox element', exampleFile, 'checkbox-aria-checked-false', async (t) => {
-  
+
   await uncheckAllConds(t);
   let checkbox = t.context.session.findElement(By.css(ex.checkboxSelector));
 
@@ -90,7 +92,7 @@ ariaTest('"aria-checked" on checkbox element', exampleFile, 'checkbox-aria-check
 });
 
 ariaTest('"aria-checked" on checkbox element', exampleFile, 'checkbox-aria-checked-mixed', async (t) => {
-  
+
   await uncheckAllConds(t);
   let checkbox = t.context.session.findElement(By.css(ex.checkboxSelector));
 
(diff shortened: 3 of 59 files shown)


test/util/queryElements.js
```diff
@@ -0,0 +1,23 @@
...
+module.exports = async function queryElements(t, selector, context) {
+  context = context || t.context.session;
+  const result = await context.findElements(By.css(selector));
+  if (result.length === 0) {
+    t.fail(`Element query returned no results: ${selector}`);
+    return [];
+  }
+  else {
+    return result;
+  }
```
zcorpan: `result` can be returned unconditionally, right?
```suggestion
  }
  return result;
```

test/util/queryElements.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const { By } = require('selenium-webdriver');
+
+/**
+ * Return an array of elements by selector. Wraps Selenium's findElements, but with a failing text for empty queries
+ *
+ * @param {String} selector - CSS selector string
```
zcorpan: `t` is not documented

zcorpan:
I added a lint check in c16088f -- there were still 12 uses (not counting the one in `queryElement.js`). Do you think those should be changed to use `queryElements`?

```
/Users/zcorpan/git/w3c/aria-practices/test/index.js
  89:16  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/checkbox_checkbox-2.js
   55:14  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties
   85:12  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties
  201:12  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/combobox_autocomplete-both.js
  60:31  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/combobox_autocomplete-list.js
  61:31  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/combobox_autocomplete-none.js
  61:31  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/combobox_grid-combo.js
  137:31  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/grid_LayoutGrids.js
  125:16  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/tabs_tabs-1.js
  355:12  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/tests/tabs_tabs-2.js
  363:12  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

/Users/zcorpan/git/w3c/aria-practices/test/util/assertAriaDescribedby.js
  51:36  error  'findElements' is restricted from being used. Please use t.context.queryElements()  no-restricted-properties

✖ 12 problems (12 errors, 0 warnings)
```

zcorpan:
Documentation in the wiki should be updated. I don't want to do this now since my changes haven't been reviewed and I don't want the documentation to be out of sync with what's in `master`.

* [ ] The eslint for `findElements`, and the per-file override `/* eslint no-restricted-properties: 0 */`
* [ ] [How to declare a test](https://github.com/w3c/aria-practices/wiki/Introduction-to-APG-Regression-Tests#how-to-declare-a-test)
* [ ] [Example test](https://github.com/w3c/aria-practices/wiki/Regression-Tests-for-APG-Example-Pages#example-test)
* [ ] [Selenium API](https://github.com/w3c/aria-practices/wiki/Regression-Tests-for-APG-Example-Pages#selenium-api)

mcking65:
@smhigley, could you please look at Simon's changes?

It would be good to have a review of the entire PR from @sh0ji or @spectranaut as well.

Given the number of files touched, I'd like to get this merged as soon as we can.

.eslintrc.json
```diff
@@ -5,6 +5,10 @@
   },
   "rules": {
     "no-unused-vars": 0,
-    "no-undef": 0
+    "no-undef": 0,
+    "no-restricted-properties": [2, {
```
nschonni: Should `plan` be added if you're trying to discourage it too?

test/index.js
```diff
@@ -1,3 +1,5 @@
+/* eslint no-restricted-properties: 0 */
```
nschonni: Might makes sense to use a single line ignore instead of a file level one

smhigley:
I left those 12 in there, because they dealt with testing the number returned, rather than just assuming within the test that there would be a non-empty array. I think that could still be a valid use case for `findElements` instead of `queryElements`.

What if we add a lint rule for `t.plan` instead, as @nschonni recommended? If that seems like it increases complexity too much for people writing tests, I also like @spectranaut's suggestion of a helper for that specific scenario. Perhaps something like `assertNotEmpty` or `assertNumberOfElements` (I'm bad at naming things... 😝)

zcorpan:
I considered linting `plan` but wasn't sure if there are still cases where it's needed. If there aren't then we can lint it.

For the cases where `findElements` is still used, we want to assert that the length is 0? A separate helper function for that which fails if the length is not 0 seems cleanest to me. `assertNoElement()` maybe?

mcking65:
What is the best approach to arriving at a decision for how to resolve the remaining issue here? The technical aspect of this discussion is beyond my expertise.

One thing I hope we optimize for is clarity of code, especially for new comers to test writing.

smhigley:
I like @zcorpan's suggestion of `assertNoElements`. I'll push up a commit with it in just a bit :)

smhigley:
I rebased and removed all the remaining uses of `findElements`, apart from the two in the `queryElements` and `assertNoElements` helpers.

@spectranaut / @zcorpan / @nschonni, how do you feel about the PR after those changes?

test/index.js
nschonni: Looks like it can be re-enabled now

test/tests/checkbox_checkbox-2.js
```diff
@@ -1,10 +1,13 @@
+/* eslint no-restricted-properties: 0 */
```
nschonni: Think this can be re-enabled again

smhigley:
@nschonni I removed the now-unnecessary `eslint no-restricted-properties`. `eslint . --fix` doesn't seem to be catching or changing anything, though

nschonni:
@smhigley don't worry about it then :)
Looks like some mixed 2/4 spaces in some functions, but if the lint isn't catching it then the rules aren't enforcing it right now and shouldn't matter here

smhigley:
@spectranaut are you ok with the PR now/do you have any outstanding concerns?

zcorpan:
(I didn't get to resolving conflicts today.)

smhigley:
@zcorpan conflicts resolved :)

PR #1344, opened 2020-02-24
Listboxes with Rearrangeable Options: Add keyboard commands for multiple selection

### LINKED ISSUE #919: Bugs in example: listbox/listbox-rearrangeable.html (multiselect keyboard interactions)
In chrome on linux, the following key commands do not work in the second example implementation on the example page [listbox/listbox-rearrangeable.html](https://w3c.github.io/aria-practices/examples/listbox/listbox-rearrangeable.html):
- Shift + Down Arrow
  - Instead, the behavior is the same as down arrow
- Shift + Up Arrow
  - Instead, the behavior is the same as up arrow
- Control + Shift + Home
  - Instead, the behavior is the same as home
- Control + Shift + End
  - Instead, the behavior is the same as end
- Control + A
  - Instead, all of the text on the web page is selected

smhigley:
Resolves #919 

Adds the following functionality:
- Shift + arrow keys
- Control + Shift + home and end
- Control + A
- Shift + Click

Includes tests :)

[Preview link](https://raw.githack.com/w3c/aria-practices/919-listbox-multiselect-keys/examples/listbox/listbox-rearrangeable.html)

### Review checklist

* [x] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by Matt (@mcking65)
* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by Carolyn (@carmacleod) and Matt (@mcking65)
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by Carolyn (@carmacleod) and Simon (zcorpan)
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by Simon (@zcorpan)

### DIFF
--- a/examples/listbox/js/listbox.js
+++ b/examples/listbox/js/listbox.js
@@ -22,6 +22,7 @@ aria.Listbox = function (listboxNode) {
   this.multiselectable = this.listboxNode.hasAttribute('aria-multiselectable');
   this.moveUpDownEnabled = false;
   this.siblingList = null;
+  this.startRangeIndex = 0;
   this.upButton = null;
   this.downButton = null;
   this.moveButton = null;
@@ -39,6 +40,10 @@ aria.Listbox.prototype.registerEvents = function () {
   this.listboxNode.addEventListener('focus', this.setupFocus.bind(this));
   this.listboxNode.addEventListener('keydown', this.checkKeyPress.bind(this));
   this.listboxNode.addEventListener('click', this.checkClickItem.bind(this));
+
+  if (this.multiselectable) {
+    this.listboxNode.addEventListener('mousedown', this.checkMouseDown.bind(this));
+  }
 };
 
 /**
@@ -86,10 +91,11 @@ aria.Listbox.prototype.focusLastItem = function () {
 aria.Listbox.prototype.checkKeyPress = function (evt) {
   var key = evt.which || evt.keyCode;
   var lastActiveId = this.activeDescendant;
-  var firstItem = this.listboxNode.querySelector('[role="option"]');
-  var nextItem = document.getElementById(this.activeDescendant) || firstItem;
+  var allOptions = this.listboxNode.querySelectorAll('[role="option"]');
+  var currentItem = document.getElementById(this.activeDescendant) || allOptions[0];
+  var nextItem = currentItem;
 
-  if (!nextItem) {
+  if (!currentItem) {
     return;
   }
 
@@ -113,7 +119,7 @@ aria.Listbox.prototype.checkKeyPress = function (evt) {
 
       if (!this.activeDescendant) {
         // focus first option if no option was previously focused, and perform no other actions
-        this.focusItem(nextItem);
+        this.focusItem(currentItem);
         break;
       }
 
@@ -129,10 +135,14 @@ aria.Listbox.prototype.checkKeyPress = function (evt) {
       }
 
       if (key === aria.KeyCode.UP) {
-        nextItem = this.findPreviousOption(nextItem);
+        nextItem = this.findPreviousOption(currentItem);
       }
       else {
-        nextItem = this.findNextOption(nextItem);
+        nextItem = this.findNextOption(currentItem);
+      }
+
+      if (nextItem && this.multiselectable && event.shiftKey) {
+        this.selectRange(this.startRangeIndex, nextItem);
       }
 
       if (nextItem) {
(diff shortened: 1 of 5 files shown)


examples/listbox/js/listbox.js
```diff
@@ -196,6 +218,13 @@ aria.Listbox.prototype.checkKeyPress = function (evt) {
         this.focusItem(nextUnselected);
       }
       break;
+    case 65:
+      // handle control + A
+      if (evt.ctrlKey) {
```
carmacleod:
```suggestion
      if (this.multiselectable && evt.ctrlKey) {
```

smhigley: Good catch!

Fixed this + the failing test :)

smhigley:
@carmacleod I added a couple more tests for the control + a behavior, including the one you suggested :)

Let me know if that addresses everything, and thanks for the review!

zcorpan:
So on macOS, the usual shortcut for select all is Command+A, not Control+A. How should we handle this? Support both Control+A and Command+A? Detect macOS and change the shortcut from Control+A to Command+A?

examples/listbox/js/listbox.js
```diff
@@ -239,6 +269,14 @@ aria.Listbox.prototype.findItemToFocus = function (key) {
   return nextMatch;
 };
 
+/* Return the index of the passed elemnt within the passed array, or null if not found */
```
zcorpan:
```suggestion
/* Return the index of the passed element within the passed array, or null if not found */
```

examples/listbox/js/listbox.js
```diff
@@ -196,6 +218,14 @@ aria.Listbox.prototype.checkKeyPress = function (evt) {
         this.focusItem(nextUnselected);
       }
       break;
+    case 65:
+      // handle control + A
+      if (this.multiselectable && evt.ctrlKey) {
```
carmacleod: To answer @zcorpan 's question:
> So on macOS, the usual shortcut for select all is Command+A, not Control+A. How should we handle this?

Detect macOS and change the shortcut from Control+A to Command+A

Something like this should do it:
```suggestion
      var isMac = window.navigator.platform.indexOf("Mac") !== -1;
      var mod = isMac ? evt.metaKey : evt.ctrlKey;
      if (this.multiselectable && mod) {
```

smhigley: I don't think we need to detect Mac. I just added `(evt.ctrlKey || evt.metaKey)`. That should work for Mac, and doesn't cause issues with the Windows key, since Windows + A is eaten by the OS anyway.

@zcorpan let me know if it's working for you with that change

carmacleod: > Windows + A is eaten by the OS anyway.

Ha! I did not know that. :)
Seems to open up the notification sidebar.
Whaddaya know.
Thanks for the lesson!

mcking65: What will this logic end up doing on Linux where I believe we want it to be ctrl-a? There could be other platforms, even some mobile with BT keyboards that will allow these commands to be used. I just pushed a change to the keyboard documentation, but I don't believe what I wrote is entirely accurate using this logic.

carmacleod: Hmmm... now that I'm looking at the logic, I believe it makes both `Ctrl+A` and `Cmd+A` work on macOS. Do we want that?

carmacleod: @mcking65 
I don't think there's a [KeyboardEvent.metaKey](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey) on Linux. Hopefully someone can confirm.

The doc says the only other `metaKey` is the Windows key on Windows keyboards, and @smhigley mentioned that Windows + A is eaten by the OS.

The code within the case for the 'A' (65) key being pressed is:
`if (this.multiselectable && (evt.ctrlKey || evt.metaKey))`

so the logic would work as follows on Linux:
"if the list is multiselectable and the control key was pressed (with the 'A')"

So the code `|| evt.metaKey` would never be evaluated.

So I think the logic is ok for Linux, but I don't think it's ok for macOS, unless we want both ctrl+A and command-A to select all.

carmacleod: @mcking65 
You are correct about the keyboards! I googled a bit and discovered that some older Linux keyboards and even some recent bluetooth android keyboards have a meta key (apparently for the mobile ones, the "search" key maps to metaKey). So even though those keyboards are likely rare, we don't want meta+A to select all. We only want that on Mac.

@smhigley I think we need to go with the code that detects Mac:
```
var isMac = window.navigator.platform.indexOf("Mac") !== -1;
var mod = isMac ? evt.metaKey : evt.ctrlKey;
if (this.multiselectable && mod) {
```

examples/listbox/listbox-rearrangeable.html
```diff
@@ -290,7 +290,10 @@ <h4 id="kbd_label_multiselect">Multiple selection keys supported in example 2</h
           <td>Selects from the focused option to the end of the list.</td>
         </tr>
         <tr data-test-id="key-control-a">
-          <th><kbd>Control + A</kbd></th>
+          <th>
+            <kbd>Control + A</kbd> (except on MacOS)<br>
+            <kbd>Command-A</kbd> (Mac OS)
```
carmacleod:
```suggestion
            <kbd>Control + A</kbd> (except on macOS)<br>
            <kbd>Command-A</kbd> (macOS)
```

PR #1365, opened 2020-04-03
Introduce a command to regenerate examples/index.html and lint that it is run

### LINKED ISSUE #1079: Autogenerated and check in examples/index.html
When merging into master, we should auto-generate and check `examples/index.html` into master. Can this happen in the travis deploy script?

zcorpan:
The command to regenerate is: npm run build

The new lint command is: npm run lint:build

Document this in the README. Run the new lint in Travis CI to ensure that
PRs that change something under examples/ also regenerate the index.

Fixes #1079.


zcorpan:
See #1366 for a demonstration of the new linter failing in Travis CI.

zcorpan:
@smhigley can you review this?

mcking65:
Thank you @zcorpan and @smhigley

I was starting to have second thoughts, wondering if this could sometimes result in conflicts within the examples/index.html file. Perhaps it could if we were to rename an example file. It would mean hand-editing a generated file to resolve conflicts during a rebase.

However, such circumstances should be very rare. In fact, the examples/index.html doesn't change very often. So, maybe we should just give this a try and see how it goes.

This is just one of those situations where the creakiness of our overall versioning/build process is manifest. We could simply do this during CI to push changes to examples;/index.html to gh-pages, but that would then create the possibility of the index.html getting out-of-date in our protected publication branches, which are separate from master and gh-pages. This all gets easier when dwe get to a place where we can stop versioning the APG.

smhigley:
@mcking65 I'm not sure I follow; why would this change have any effect on git conflicts? People should already be updating `examples/index.html` when necessary, and this change just checks (and makes the build error) if they've forgotten to.

It looks like it correctly caught an error in master after your merge, as it looks like `examples/index.html` was not regenerated after the tabbed carousel example was added.

spectranaut:
Closing in favor of  #1524

PR #1370, opened 2020-04-10
Use local index variables for loops

pepelsbey:
Here’s my story:

I took [tabs.js](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-2/js/tabs.js) file from the [tabs implementation](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel) and it worked just fine… Until I transpiled it with [babel/preset-env](https://babeljs.io/docs/en/babel-preset-env) and minified with [terser](https://github.com/terser/terser). This is a typical thing to do with any modern JS code.

```js
"use strict";!function(){var e,t,n=document…
```

Babel adds `"use strict";` by default and this is what broke it for me. Turns out there are for loops using global index variables:

```js
for (i = 0; i < tabs.length; ++i) {
    addListeners(i);
}
```

Which works fine until you add `"use strict";`.

I went through all the sample JS code and added `var` in front of every index variable (not defined previously) to make it compatible with the strict mode.

Also there are some trailing spaces are gone according to your EditorConfig, sorry about that. I could probably revert that and keep only `var` in this PR.


pepelsbey:
Done: reverted aria.js

examples/checkbox/checkbox-2/checkbox-2.html
```diff
@@ -72,12 +72,12 @@ <h2 id="ex_label">Example</h2>
...
               var i;
               var mixed = document.querySelectorAll('[role="checkbox"]');
-              for (i = 0; i < mixed.length; i++) {
+              for (var i = 0; i < mixed.length; i++) {
                 var m = new CheckboxMixed(mixed[i]);
                 m.init();
               }
               var checkboxes = document.querySelectorAll('input[type="checkbox"]');
-              for (i = 0; i < checkboxes.length; i++) {
+              for (var i = 0; i < checkboxes.length; i++) {
```
zcorpan, line 80: `i` is declared at the top of the function. I think this should be reverted.

zcorpan:
I wonder if we should add `"use strict"` everywhere, and enforce it with a lint check. I can report an issue to discuss it.

zcorpan:
https://github.com/w3c/aria-practices/issues/1379

zcorpan:
> Also there are some trailing spaces are gone according to your EditorConfig, sorry about that.

I think that's OK, and I wonder why that happens. I suppose some contributors use editors that don't apply the EditorConfig. Maybe this should be enforced with a lint check, too? https://github.com/w3c/aria-practices/issues/1380

examples/checkbox/checkbox-2/checkbox-2.html
pepelsbey: Done

mcking65:
@zcorpan, I see your approval so I assume this is good to go now?

@jongund, just a heads up that we'll be merging this, and it will have a very small effect on 2 of your PRs.

zcorpan:
@mcking65 Yes, this is ok to merge.

examples/landmarks/js/visua11y.js
```diff
@@ -825,7 +825,7 @@ function nameFromAttributeIdRefs(element, attribute) {
   if (value.length) {
     idRefs = value.split(' ');
 
-    for (i = 0; i < idRefs.length; i++) {
+    for (var i = 0; i < idRefs.length; i++) {
```
smhigley, line 828: `var i` is already declared with the other variables in this function block on line `820`. Could you remove that declaration (on 820) and keep the one you added in the for block?

pepelsbey: @smhigley done!

PR #1381, opened 2020-04-18
Fix nested modal dialogs not working on Safari/VoiceOver

### LINKED ISSUE #1241: Modal Dialog Example: Nested modal dialogs don't work on Safari/VoiceOver
The nested modal dialog example on https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html doesn't work with Safari/VoiceOver (both on MacOS and iOS).

The VoiceOver cursor ring remains on the parent modal.

Nov-02-2019 23-18-42

I believe that this is because VoiceOver doesn't accept nested elements with `aria-modal="true"` that aren't actual descendants in the DOM. It doesn't work even using `aria-owns` on the parent modal.

The solution, I guess, is to put the nested dialog as a child of the parent dialog.

diegohaz:
Fixes #1241

Some tests failed locally, but I couldn't understand how they could be related to these changes.

**Before**

Example showing the broken nested modal dialog page being used with Safari and VoiceOver where VO doesn't move focus into the nested dialog

**After**

Example showing the fixed nested modal dialog page being used with Safari and VoiceOver where VO properly move focus into the nested dialog

### DIFF
--- a/examples/dialog-modal/js/dialog.js
+++ b/examples/dialog-modal/js/dialog.js
@@ -199,7 +199,11 @@ aria.Utils = aria.Utils || {};
     // If this modal is opening on top of one that is already open,
     // get rid of the document focus listener of the open dialog.
     if (aria.OpenDialogList.length > 0) {
-      aria.getCurrentDialog().removeListeners();
+      var topDialog = aria.getCurrentDialog();
+      topDialog.removeListeners();
+      // Safari and VoiceOver don't support multiple aria-modal="true" elements
+      // on the page, so we temporarily remove the parent dialog attribute.
+      topDialog.removeAriaModalAttribute();
     }
 
     this.addListeners();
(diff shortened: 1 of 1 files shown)


smhigley:
@diegohaz this looks great, thanks for doing it! The `aria-modal` addition would need a test added as well; is that something you'd be comfortable doing in this PR? If not, we can create a separate issue/PR for that, or I can add a PR to your branch for tests.

diegohaz:
Sure! I'll take a look into it over the weekend. :)

diegohaz:
@smhigley Just added a test for `aria-modal`.

test/tests/dialog-modal_dialog.js
```diff
@@ -203,6 +204,13 @@ ariaTest('"aria-modal" attribute on role="dialog"', exampleFile, 'aria-modal', a
     await assertAttributeValues(t, ex.dialogSelector, 'aria-modal', 'true');
 });
 
+ariaTest('"aria-modal" attribute on nested role="dialog"', exampleFile, 'aria-modal', async (t) => {
+  t.plan(2);
```
smhigley, line 208:
```suggestion
```

We've updated the tests slightly to no longer need `t.plan`, probably caught in your rebase :)

Issue #1267, opened 2019-11-14
Combobox Examples: Make compatible with touch-based screen readers


mcking65:
In review of #1255, issues with iOS VO were discovered.

Test with iOS VO and resolve issues.

Edit: Specifically, from Sarah's review in #1255:
> iOS + VoiceOver: the "open" button did not work (didn't open the listbox, and sent the cursor back to the text input). I'd suggest either making the open button aria-hidden="true" tabindex="-1", or removing it altogether.

Reply from Matt:
> Surprising suggestion. How would touch users expand/collapse the combobox if it is not accessible?
> This is a widget that we should make touch compatible before APG 1.2 release. Opened #1267.

a11ydoer:
using pointer events has been working well for touch issue. thanks, @jongund

a11ydoer:
@jongund can you provide the example link? I could not find the example link from the issue 1255. Thanks in advance.

jongund:
@a11ydoer There is a linked pull request #2086 to this issue, it contains the preview links for the changes.

a11ydoer:
Example links
[Combobox Autocomplete: Both](https://raw.githack.com/w3c/aria-practices/combobox-pointer-events/examples/combobox/combobox-autocomplete-both.html)
[Combobox Autocomplete: List](https://raw.githack.com/w3c/aria-practices/combobox-pointer-events/examples/combobox/combobox-autocomplete-list.html)
[Combobox Autocomplete: None](https://raw.githack.com/w3c/aria-practices/combobox-pointer-events/examples/combobox/combobox-autocomplete-none.html)

PR #1396, opened 2020-05-04
Select-only combobox

### LINKED ISSUE #1026: Develop example of select-only combobox

smhigley:
Resolves #1026 by adding a new select-only combobox example.

NOTE: The build is currently expected to fail; this depends on the es6 PR (#1395) for linting to pass.

[Preview link](https://raw.githack.com/w3c/aria-practices/select-only-combo/examples/combobox/combobox-select-only.html)

### Review checklist

* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @charmarkk 
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @zcorpan
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @zcorpan

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1396.html" title="Last updated on Jun 30, 2020, 8:58 PM UTC (6695e84)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1396/162a945...6695e84.html" title="Last updated on Jun 30, 2020, 8:58 PM UTC (6695e84)">Diff</a>

### DIFF
--- a/examples/combobox/combobox-select-only.html
+++ b/examples/combobox/combobox-select-only.html
@@ -1,0 +1,449 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="utf-8" />
+<title>Select-Only Combobox Example | WAI-ARIA Authoring Practices 1.2</title>
+
+<!--  Core js and css shared by all examples; do not modify when using this template. -->
+<link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/base.css">
+<link rel="stylesheet" href="../css/core.css">
+<script src="../js/examples.js"></script>
+<script src="../js/highlight.pack.js"></script>
+<script src="../js/app.js"></script>
+
+<!--  js and css for this example. -->
+<link href="css/select-only.css" rel="stylesheet">
+<script src="js/select-only.js"  type="text/javascript"></script>
+</head>
+<body>
+  <nav aria-label="Related Links" class="feedback">
+    <ul>
+      <li><a href="../../#browser_and_AT_support">Browser and Assistive Technology Support</a></li>
+      <li><a href="https://github.com/w3c/aria-practices/issues/new">Report Issue</a></li>
+      <li><a href="https://github.com/w3c/aria-practices/projects/7">Related Issues</a></li>
+      <li><a href="../../#combobox">Design Pattern</a></li>
+    </ul>
+  </nav>
+  <main>
+    <h1>Select-Only Combobox Example</h1>
+    <p>The following example implementation of the <a href="../../#combobox">ARIA design pattern for combobox</a>
+    demonstrates a single-select combobox widget that is functionally similar to an HTML <code>select</code> element.
+    Contrary to other combobox examples, this combobox has no <code>&lt;input&gt;</code> element, and does not take freeform user input. Like a <code>&lt;select&gt;</code> a user can still type characters to auto-select matching options, however.
+    </p>
+    <p>Similar examples include: </p>
+    <ul>
+      <li><a href="combobox-autocomplete-both.html">Editable Combobox with Both List and Inline Autocomplete</a>: An editable combobox that demonstrates the autocomplete behavior known as <q>list with inline autocomplete</q>.</li>
+      <li><a href="combobox-autocomplete-list.html">Editable Combobox with List Autocomplete</a>: An editable combobox that demonstrates the autocomplete behavior known as <q>list with manual selection</q>.</li>
+      <li><a href="combobox-autocomplete-none.html">Editable Combobox Without Autocomplete</a>: An editable combobox that demonstrates the behavior associated with <code>aria-autocomplete=none</code>.</li>
+      <li><a href="grid-combo.html">Editable Combobox with Grid Popup</a>: An editable combobox that presents suggestions in a grid, enabling users to navigate descriptive information about each suggestion.</li>
+    </ul>
+
+    <section>
+      <h2 id="ex_label">Example</h2>
+      <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
+      <div id="ex1">
+        <label id="combo1-label" class="combo-label">Read-only Select</label>
+        <div class="combo js-select">
+        <div
+            aria-autocomplete="none"
+            aria-controls="listbox1"
+            aria-expanded="false"
+            aria-haspopup="listbox"
+            aria-labelledby="combo1-label"
+            id="combo1"
+            class="combo-input"
+            role="combobox"
+            tabindex="0"
+            >
+            <span class="combo1-value"></span>
+        </div>
+        <div class="combo-menu" role="listbox" id="listbox1">
+            <!-- options are inserted here -->
+            <!-- e.g. <div role="option" id="op1">option text</div>  -->
+        </div>
+        </div>
+      </div>
+      <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
+    </section>
+
+    <section>
+      <h2 id="kbd_label">Keyboard Support</h2>
+      <p>
+      The example combobox on this page implements the following keyboard interface.
+        Other variations and options for the keyboard interface are described in the
+        <a href="../../#combobox_kbd_interaction">Keyboard Interaction section of the combobox design pattern.</a>
+      </p>
+      <h3 id="kbd_label_textbox">Textbox</h3>
+      <table aria-labelledby="kbd_label_textbox kbd_label" class="def">
+        <thead>
+          <tr>
+            <th>Key</th>
+            <th>Function</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr data-test-id="textbox-key-down-arrow">
+            <th><kbd>Down Arrow</kbd></th>
+            <td>
+              <ul>
+                <li>First opens the listbox if it is not already displayed and then moves visual focus to the first option.</li>
+                <li>DOM focus remains on the textbox.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="textbox-key-alt-down-arrow">
+            <th><kbd>Alt + Down Arrow</kbd></th>
+            <td>
+                Opens the listbox without moving focus or changing selection.
+            </td>
+          </tr>
+          <tr data-test-id="textbox-key-up-arrow">
+            <th><kbd>Up Arrow</kbd></th>
+            <td>
+              <ul>
+                <li>First opens the listbox if it is not already displayed and then moves visual focus to the last option.</li>
+                <li>DOM focus remains on the textbox.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="textbox-key-enter">
+            <th><kbd>Enter</kbd></th>
+            <td>
+              <ul>
+                <li>Sets the textbox value to the content of the selected option.</li>
+                <li>Closes the listbox if it is displayed.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="standard-single-line-editing-keys">
+            <th>Standard single line text editing keys</th>
+            <td>
+              <ul>
+                <li>Keys used for cursor movement and text manipulation, such as <kbd>Delete</kbd> and <kbd>Shift + Right Arrow</kbd>.</li>
+                <li>An HTML <code>input</code> with <code>type="text"</code> is used for the textbox so the browser will provide platform-specific editing keys.</li>
+              </ul>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+      <h3 id="kbd_label_listbox">Listbox Popup</h3>
+      <p>
+        <strong>NOTE:</strong> When visual focus is in the listbox, DOM focus remains on the textbox and the value of <code>aria-activedescendant</code> on the textbox is set to a value that refers to the listbox option that is visually indicated as focused.
+        Where the following descriptions of keyboard commands mention focus, they are referring to the visual focus indicator.
+        For more information about this focus management technique, see
+        <a href="../../#kbd_focus_activedescendant">Using aria-activedescendant to Manage Focus.</a>
+      </p>
+      <table aria-labelledby="kbd_label_listbox kbd_label" class="def">
+        <thead>
+          <tr>
+            <th>Key</th>
+            <th>Function</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr data-test-id="listbox-key-enter">
+            <th><kbd>Enter</kbd></th>
+            <td>
+              <ul>
+                <li>Sets the textbox value to the content of the focused option in the listbox.</li>
+                <li>Closes the listbox.</li>
+                <li>Sets visual focus on the textbox.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="listbox-key-escape">
+            <th><kbd>Escape</kbd></th>
+            <td>
+              <ul>
+                <li>Closes the listbox.</li>
+                <li>Sets visual focus on the textbox.</li>
+               </ul>
+            </td>
+          </tr>
+          <tr data-test-id="listbox-key-down-arrow">
+            <th><kbd>Down Arrow</kbd></th>
+            <td>
+              <ul>
+                <li>Moves visual focus to the next option.</li>
+                <li>If visual focus is on the last option, moves visual focus to the first option.</li>
+                <li>Note: This wrapping behavior is useful when <kbd>Home</kbd> and <kbd>End</kbd> move the editing cursor as described below.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="listbox-key-up-arrow">
+            <th><kbd>Up Arrow</kbd></th>
+            <td>
+              <ul>
+                <li>Moves visual focus to the previous option.</li>
+                <li>If visual focus is on the first option, moves visual focus to the last option.</li>
+                <li>Note: This wrapping behavior is useful when <kbd>Home</kbd> and <kbd>End</kbd> move the editing cursor as described below.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="listbox-key-right-arrow">
+            <th><kbd>Right Arrow</kbd></th>
+            <td>Moves visual focus to the textbox and moves the editing cursor one character to the right.</td>
+          </tr>
+          <tr data-test-id="listbox-key-left-arrow">
+            <th><kbd>Left Arrow</kbd></th>
+            <td>Moves visual focus to the textbox and moves the editing cursor one character to the left.</td>
+          </tr>
+          <tr data-test-id="listbox-key-home">
+            <th><kbd>Home</kbd></th>
+            <td>Moves visual focus to the textbox and places the editing cursor at the beginning of the field.</td>
+          </tr>
+          <tr data-test-id="listbox-key-end">
+            <th><kbd>End</kbd></th>
+            <td>Moves visual focus to the textbox and places the editing cursor at the end of the field.</td>
+          </tr>
+          <tr data-test-id="listbox-key-char">
+            <th>Printable Characters</th>
+            <td>
+              <ul>
+                <li>Moves visual focus to the textbox.</li>
+                <li>Types the character in the textbox.</li>
+                <li>Options in the listbox are <em>not</em> filtered based on the characters in the textbox.</li>
+              </ul>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+      <h3 id="kbd_label_button">Button</h3>
+      <p>
+        The button has been removed from the tab sequence of the page, but is still important to assistive technologies for mobile devices that use touch events to open the list of options.
+      </p>
+    </section>
+
+    <section>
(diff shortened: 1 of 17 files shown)


a11ydoer:
@mcking65 @charmarkk @carmacleod @zcorpan 
I apprecite your review by Tuesday morning so that we can merge this soon.

charmarkk:
A11y Review
---
Environments: 
- Windows 10 1809, Edge 81.0.416.72, Firefox 76.0.1, Chrome 81.0.4044.138 using NVDA 2020.1 and JAWS 2019.1912.1
- MacOS Mojave 10.14.6, Safari 13.1 using VoiceOver 9

In all Windows environments it seemed that screen readers behaved as expected and according to the provided documentation with the example. Functionally speaking, I think there are some issues with the Up Arrow and Alt + Down Arrow combinations working with regard to the closed combobox. They don't seem to behave any differently for me than the Down Arrow on its own. Unless I'm misreading or misunderstanding something. That said, I am able to use Home and End appropriately, type characters in quick succession (B O) to skip past items (blueberry, in this case), and Escape out all as expected.

VoiceOver did not want to seem to use Ctrl-Option Space to open the combobox, nor would it announce entries inside the listbox dropdown no matter what I did. It does seem to work as detailed in documentation otherwise, with the same caveat of the Windows side - Up Arrow and Alt + Down Arrow didn't seem to do anything different for me (again, I may be misunderstanding).

carmacleod:
This is really nice! 🚀 
I would be happy for it to go in as-is.
I do have a few notes and questions, but they don't need to hold up merging this PR.

- should we just delete `aria-autocomplete="none"` from the code? The property doesn't make sense on a select-only combobox. (and it's not listed in the example's attributes table)
- should the listbox have a label? (i.e. we use aria-label[ledby] on all of the other combobox examples)
- should the listbox use ul/li instead of divs? (I think the "progressive enhancement" folks would say yes, but I actually like divs here, because it feels like it's acknowledging that the popup could have been a tree, grid, or dialog instead...)
- should we mention the places where this diverges from html select behavior? Perhaps in an Accessibility Features section? Yes, the first paragraph does say, "functionally _similar_ to an HTML select element" so we never said it was exactly the same, but maybe we want to explain/justify some of the choices? i.e.
  - select doesn't open the listbox when the user types characters
  - select selects immediately when using arrow keys in listbox, but this only selects on space, enter, or loss of focus, and can therefore be cancelled with esc if someone is just using arrow keys to look at possible values (whereas select value cannot be reverted to previous)
  - select closes the listbox if the combobox is clicked with a mouse (should we do this? i.e. behaves as if listbox lost focus. It feels a bit weird to mouse click on the dropdown arrow twice and have the listbox stay open)
- down arrow and alt+down arrow seem to work exactly the same, but the keyboard support notes for them are different?
- maybe the listbox needs to be a couple of pixels longer to accommodate the bottom of the focus indicator when the last item has visual focus. (only tried in Chrome)
- a couple of sentences in the attributes table mention `input`:
  - "Identifies the input as a combobox" --> maybe change `input` to element?
  - "Enables assistive technologies to know which element the application regards as focused while DOM focus remains on the input element." --> maybe change `input` to combobox?
- there are 4 instances of the word "textbox" - should they be changed to "combobox"?

Note: I have only tested in Windows/Chrome so far. I plan to finish testing on other browsers later today, but as I am out of time for now, I thought I'd better get these comments out sooner rather than later.

a11ydoer:
We need to confirm about 'active decendant' would work in VO and Google Chrome Canary.
@carmacleod will check with VO and Safari in recent iOS version.

carmacleod:
Tested in VoiceOver on macOS Catalina 10.15.4 (latest) in Safari 13.1.

I agree with @charmarkk that the combobox is not very usable in VO/Safari. It doesn't open with VO+Space, and the items are not read on up or down arrow.

I can force it to work if I open the combobox with down arrow (or alt+down arrow), then use VO+right arrow to navigate to the listbox and VO+Shift+down arrow to go into the listbox; then I can use VO+right/left to go through the combobox items, and select an item with VO+Space. But I doubt many VO users would fight with it that long, particularly as it is silent when the listbox is first opened.

Looking back at the testing notes for the editable combobox examples, I see the following:
#### From @smhigley in https://github.com/w3c/aria-practices/pull/1255#pullrequestreview-316639448
- macOS + VoiceOver: the suggested "open" keyboard command does not work (Caps + Space), though Option + down arrow does
- macOS + VoiceOver: although everything but Caps+ Space worked for me in 10.14, @jnurthen reported that 10.15 did not work; someone else with Catalina should probably officially test this.

#### Reply from @mcking65 in https://github.com/w3c/aria-practices/pull/1255#issuecomment-553793452
My system is still 10.4.6, and it works for me too. We will need to continue this testing prior to APG 1.2 R1.

#### From @carmacleod in https://github.com/w3c/aria-practices/pull/1255#issuecomment-553966226
VoiceOver+Safari on macOS Catalina 10.15.1
- VO does not read the list items in the autocomplete=list or autocomplete=none or grid examples in Safari. These 3 examples are completely unusable in VO+Safari on Catalina.
- However, VO does read list items in the autocomplete=both example
- VO+space does not open the list in Safari (for any example)

#### Reply from: @mcking65 in https://github.com/w3c/aria-practices/pull/1255#issuecomment-553976921
> - VO does not read the list items in the autocomplete=list or autocomplete=none or grid examples in Safari. These 3 examples are completely unusable in VO+Safari on Catalina.

The fact that they do work in 10.14.6 does not necessarily implicate Safari, but kind of points in that direction. @accdc has some different implementations of the same pattern we could test with 10.15. Apple may have implemented overly restrictive processing of aria-activedescendant, overlooking the fact that the ARIA spec supports activedescendant on a text input or combobox that controls an element that also supports activedescendant.

> - However, VO does read list items in the autocomplete=both example

This is almost certainly because of the inline selection.

> - VO+space does not open the list in Safari (for any example)

VO-space is supposed to be equivalent to a mouse click on the element in the VO cursor. Given that, does VO announcing that VO-space should open the popup make sense?

#### Reply from @carmacleod (today):

I don't know whether VO announcing that VO+Space opens an editable combobox popup makes sense or not, but note that VO+Space doesn't open the select-only combobox either, and mouse click does.

Aside: We should also test select-only combobox with iOS VoiceOver, because I notice that we opened https://github.com/w3c/aria-practices/issues/1267 for the editable comboboxes.

----

@cookiecrook We seem to be having trouble getting our ARIA 1.2 comboboxes to work well in macOS Catalina VoiceOver in Safari. They work well in the usual Windows pairings, and in VO/Chrome, so Safari would seem to be the most likely suspect... :)

There's a possibly-relevant [webkit bug for the option items not being read](https://bugs.webkit.org/show_bug.cgi?id=167671).
There's also an ancient but possibly-relevant [webkit bug about VO+Space not working in select](https://bugs.webkit.org/show_bug.cgi?id=139539).


Here a couple of ARIA 1.2 comboboxes (i.e. aria-controls points to listbox; aria-owns is not used):
1. [editable combobox with autocomplete=list](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html)
2. [select-only combobox](https://raw.githack.com/w3c/aria-practices/select-only-combo/examples/combobox/combobox-select-only.html)

Does anything stand out to you?

carmacleod:
#### Notes from today's call:

(I won't commit anything, because @smhigley has uncommitted changes and I don't want to create conflicts... so I'll create a "to do list")

- [ ] delete `aria-autocomplete="none"` from the code
  - the definition of [aria-autocomplete="none"](https://w3c.github.io/aria/#aria-autocomplete) in the spec doesn't really fit (i.e. we'd need a new value token or something):
    > When a user is providing input, an automatic suggestion that attempts to predict how the user intends to complete the input is not displayed.

- [ ] Add `aria-label="Fruits"` to the `listbox`
  - also add to attributes table
  - the other combobox examples all have `aria-label[ledby]`
  - example: in the case of the [autocomplete none example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-none.html), the combobox label is "Search", and the listbox label is "Previous Searches".
<details><summary>NVDA reads the listbox label</summary>
<br>Search  combo box  collapsed  editable
<br>blank
<br>[down arrow]
<br>expanded
<br>Previous Searches  list
<br>weather  1 of 11
<br>[down arrow]
<br>salsa recipes  2 of 11
</details>
<details><summary>JAWS reads the listbox label</summary>
<br>Search edit combo collapsed 
<br>To set the value use the Arrow keys or type the value.
<br>[down arrow]
<br>Previous Searches List box 
<br>weather 
<br>1 of 11
<br>To move to an item press the Arrow keys.
<br>[down arrow]
<br>salsa recipes 
<br>2 of 11
</details>

- [x] should the listbox use ul/li instead of divs?
  - no, don't bother... we have a philosophy of "use native semantics that most closely matches", but there's no real advantage here.

- [ ] mention the places where this diverges from html select behavior in an Accessibility Features section:
  - select doesn't open the listbox when the user types characters
    - beneficial to see what is in list box
  - select selects immediately when using arrow keys in listbox, but this only selects on space, enter, or loss of focus, and can therefore be cancelled with esc if someone is just using arrow keys to look at possible values (whereas select value cannot be reverted to previous)
    - the ability to cancel is a great accessibility feature

- [ ] close the listbox if it is open and the combobox is clicked with a mouse (i.e. clicking twice on combobox opens and then closes. This is what select does, and it is expected behavior).

- [ ] down arrow and alt+down arrow work exactly the same, but the keyboard support notes for them are different
  - update notes so that both go to current value, not first value

- [ ] listbox needs to be a couple of pixels longer to accommodate the bottom of the focus indicator when the last item has visual focus. (only seen in Chrome)

- [ ] "Identifies the input as a combobox" --> change "input" to "element"

- [ ] "Enables assistive technologies to know which element the application regards as focused while DOM focus remains on the input element." --> change "input" to "combobox"

- [ ] change 4 instances of the word "textbox" to "combobox"

- [ ] document that alt+up arrow works like enter - saves value

- [ ] document pageup and pagedown

a11ydoer:
@carmacleod 
Your through research on previous comments and issues are awesome! 
Also thanks so much for notes from yesterday's meeting. To be somewhat helpful to all your great work, I added a minor info. 

>update notes so that both go to current value, not first value

This refer to the case when a user closes select and reopen it. The focus goes back to last selected value, not the first item in the list.

smhigley:
@carmacleod thanks for the thorough notes! My latest update added tests and fixed a few of the items you had listed.

I can't repro the Chrome outline/pixel thing you mentioned, so may need to ask you about that separately. I also did some digging on listbox labeling that I'll bring to the next meeting :)

I added some keyboard docs so I could target them with tests, but they likely still need wordsmithing. I'm taking off the "WIP" designation though.

Thanks again!

examples/combobox/combobox-select-only.html
```diff
@@ -0,0 +1,413 @@
...
+                Opens the listbox and moves visual focus to the first option.
+            </td>
+          </tr>
+          <tr data-test-id="combobox-key-end">
+            <th><kbd>End</kbd></th>
+            <td>
+                Opens the listbox and moves visual focus to the last option.
+            </td>
+          </tr>
+          <tr data-test-id="standard-single-line-editing-keys">
```
carmacleod, line 131: might want to change this test id to "printable-characters"?

carmacleod:
@smhigley In case it's helpful, here's a screen snap showing what I mean about the focus indicator being clipped on the bottom. I can get this just typing down arrows until the list scrolls. Chrome 83.0.4103.61:

side-by-side screen snaps showing both complete and clipped focus indicator

cookiecrook:
> @cookiecrook We seem to be having trouble getting our ARIA 1.2 comboboxes to work well in macOS Catalina VoiceOver in Safari. They work well in the usual Windows pairings, and in VO/Chrome, so Safari would seem to be the most likely suspect... :)
> 
> There's a possibly-relevant webkit bug for the option items not being read.

cookiecrook:
> @cookiecrook We seem to be having trouble getting our ARIA 1.2 comboboxes to work well in macOS Catalina VoiceOver in Safari. They work well in the usual Windows pairings, and in VO/Chrome, so Safari would seem to be the most likely suspect... :)
> 
> There's a possibly-relevant webkit bug for the option items not being read.

cookiecrook:
> @cookiecrook We seem to be having trouble getting our ARIA 1.2 comboboxes to work well in macOS Catalina VoiceOver in Safari. They work well in the usual Windows pairings, and in VO/Chrome, so Safari would seem to be the most likely suspect... :)
> 
> There's a possibly-relevant webkit bug for the option items not being read.

Those examples work most of the time for me... I was able to find a timing edge case that might be causing the trouble you're experiencing and linked it to the first bug report you mentioned. I don't think the second bug report is related. 

The gist of the remaining bug is that is you arrow down right as a different speech announcement is being spoken (e.g. verbose hints or word echo, which most VO users turn off), the combobox selection change announcement is getting trampled. I'll escalate that one. Thanks!

carmacleod:
@smhigley I noticed one more functional issue. (sorry!)

Steps:
- Windows Chrome, Windows FF or macOS FF (issue does not occur in other browsers)
- go to the [select-only combobox preview](https://raw.githack.com/w3c/aria-practices/select-only-combo/examples/combobox/combobox-select-only.html)
- tab to the combo
- drop down the combo
- type tab again
- combo closes (good) but focus (document.activeElement) goes to body (not good)
- issue: from here, you can't drop down the combo again

I think this should be fixed, because I'm guessing that some users will try using the tab key to navigate to options in the dropdown, and those users would trip over this issue, and get confused when they can't reopen the combobox.

I believe the desired behavior is to focus the combo after closing the dropdown (because that's what most selects do), however, I suppose it would be ok to focus the next element in the tab order.

Note that shift+tab in the example focuses the previous element in the tab order after closing, whereas selects focus the select after closing (i.e. consistent with their forward tab behavior).

The native select behavior mentioned above (i.e. tab closes and then refocuses select) was observed in Windows Chrome, Edge, and Windows & macOS Firefox.

Safari and macOS Chrome do nothing when the user types tab (or shift+tab) in an open select (which feels a bit "trappy", but I guess the user can type enter or escape to close, or even... space?!).

smhigley:
@carmacleod never apologize for finding a functional bug! 😄

I had actually thought I'd already included a fix for that one, but it looks like I forgot to commit & push 🤦‍♀️.

The root cause is Firefox puts any region with scroll overflow into the tab order, so tabbing sends you into the listbox immediately before it is hidden. The fix I did was to simply make the listbox un-tabbable (since it seems like it really shouldn't be), which means that tab would take you to the next item in the focus order after the combobox.

Are you OK with that solution? It seems kinda more intuitive to me, even though the native Windows select keeps you on the `<select>`. Keeping focus on the same element after a tab just feels kinda weird to me somehow 🤷‍♀️

carmacleod:
> tab would take you to the next item in the focus order after the combobox. ... Keeping focus on the same element after a tab just feels kinda weird to me somehow 

I think that would be ok... but we have to wonder why the native controls have the "stay in place" behavior. Maybe some users got confused about whether or not the current item in the list (i.e. at the time they typed tab) became the select's value? i.e. Keeping them on the select would allow them to review the select's value?

Just bouncing thoughts around, I'm guessing that staying on the combo could be important in a multi-select combobox if the user wants to review the chosen values after closing the popup?

Yet another subtle behavior where real user data would be so valuable. :)

Anyhow, the current behavior is broken, so yes, please push the fix to go to the next (previous) item in the tab order on tab (shift+tab), and we can discuss with the group whether we want to keep the "traverse onward" behavior or change it to "stay in place" behavior.

smhigley:
@carmacleod I figured out what happened with the tabbing thing and tests -- somehow I managed to not push up either commit, which is what I get for working on the same PR across two computers 😆🤦

@mcking65 the note about tabbing has been added to the keyboard behavior section now :)

carmacleod:
> working on the same PR across two computers

Heh heh - yep. Our old "IDE in the Cloud" would've fixed that problem, but it was before its time. ;)

examples/combobox/combobox-select-only.html
```diff
@@ -0,0 +1,415 @@
...
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="combobox-key-alt-down-arrow">
+            <th><kbd>Alt + Down Arrow</kbd></th>
+            <td>
+                Opens the listbox without moving focus or changing selection.
+            </td>
+          </tr>
+          <tr data-test-id="combobox-key-up-arrow">
```
spectranaut, line 98: There is not test for this data-test-id

spectranaut, line 98: There is no test for this data-test-id

examples/combobox/combobox-select-only.html
spectranaut: actually it looks like this keyboard interaction is covered by tests with the `data-test-id="listbox-key-char"`. Right now, the regression report says there are no tests for this data-test-id but there are. @smhigley, maybe you should give both of these the same data-test-id, such as `printable-characters`, and update the test reference as well?

examples/combobox/combobox-select-only.html
```diff
@@ -0,0 +1,415 @@
...
+          <tr data-test-id="listbox-key-up-arrow">
+            <th><kbd>Up Arrow</kbd></th>
+            <td>
+              <ul>
+                <li>Moves visual focus to the previous option.</li>
+                <li>If visual focus is on the first option, visual focus does not move.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="listbox-key-alt-up-arrow">
```
spectranaut, line 217: this data-test-id is also missing a test

spectranaut:
> This is an approval of the tests requesting three changes, left in comments! Two tests are missing and two data-test-ids should be updated.

Heads up @smhigley -- I'm going to add these test now :)

examples/combobox/combobox-select-only.html
```diff
@@ -0,0 +1,417 @@
...
+      <li><a href="https://github.com/w3c/aria-practices/projects/7">Related Issues</a></li>
+      <li><a href="../../#combobox">Design Pattern</a></li>
+    </ul>
+  </nav>
+  <main>
+    <h1>Select-Only Combobox Example</h1>
+    <p>
+      The following example implementation of the <a href="../../#combobox">ARIA design pattern for combobox</a>
+      demonstrates a single-select combobox widget that is functionally similar to an HTML <code>select</code> element.
+      Unlike the editable combobox examples, this select-only combobox is not made with a <code>&lt;input&gt;</code> element, and it does not accept freeform user input.
```
carmacleod, line 32:
```suggestion
      Unlike the editable combobox examples, this select-only combobox is not made with an <code>&lt;input&gt;</code> element, and it does not accept freeform user input.
```

JAWS-test:
https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html: 

JAWS and NVDA do not output the select-only combobox value in Firefox and Chrome. This is because the value is not passed to the Accessibility API, as shown by inspection with inspect.exe or aViewer.

mcking65:
@JAWS-test wrote:

> https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html:
> 
> JAWS and NVDA do not output the select-only combobox value in Firefox and Chrome. This is because the value is not passed to the Accessibility API, as shown by inspection with inspect.exe or aViewer. The problem cannot be solved with the current ARIA specification. To do so, we would have to allow aria-valuetext on more elements, as I have been suggesting for a long time ([w3c/aria#711](https://github.com/w3c/aria/issues/711)). The only solution that is currently possible would not be to use a div, but an input for the select-only combobox.
> 
> Further test results with JAWS 2020: [FreedomScientific/VFO-standards-support#399](https://github.com/FreedomScientific/VFO-standards-support/issues/399)

Correct, this is a browser issue. The requirement to expose the value is in ARIA 1.2. It is implemented in Chrome Canary. I think the WebKit fix is also in progress. Not sure if Gecko is started yet.

See [ARIA PR 1225](https://github.com/w3c/aria/pull/1225)

carmacleod:
@JAWS-test 

We need the APG examples to follow the spec. If someone needs this to work in older browsers, they can implement a work-around using aria-labelledby with multiple ids, something like the following (also would need a little bit of tweaking of the javascript to ensure that the value content is being set in the combo1-value element):
```html
<div id="combo1" class="combo-input" tabindex="0"
        role="combobox" aria-haspopup="listbox" aria-controls="listbox1"
        aria-labelledby="combo1-label combo1-value" 
        aria-expanded="false" aria-activedescendant="">
                <div id="combo1-value">Apple</div>
</div>
```

JAWS-test:
@mcking65  Sorry, I forgot ARIA PR 1225
@carmacleod This workaround is already used by many frameworks. Whenever I see it in a WCAG test, I do not accept it and notice a violation of the WCAG. Reason: Label and value are different things and must be perceived individually with AT, because otherwise the AT user cannot know what label and value are

PR #1404, opened 2020-05-14
All Examples: Add browser and AT support notice

### LINKED ISSUE #1228: Stronger support / testing / best practices text on every example
Because these patterns are often referenced via direct links to the example pages, users do not encounter the warnings that come at the start of the overall document. While there is a link at the start of each pattern, in my experience users are not clicking (or reading) before copying patterns into their own projects.

I recommend placing a very obvious warning at the start of each pattern that it is not tested with users, not tested across all browser/AT combinations (linking to [2.2](https://www.w3.org/TR/wai-aria-practices-1.1/#browser_and_AT_support) and [2.3](https://www.w3.org/TR/wai-aria-practices-1.1/#mobile_and_touch_support)), intentionally ignores the First Rule of ARIA (linking to the [First Rule of ARIA](https://www.w3.org/TR/using-aria/#rule1)), and is only a hypothetical application of ARIA (linking to [2.1](https://www.w3.org/TR/wai-aria-practices-1.1/#no_aria_better_bad_aria)).

I see from the [24 September call](https://www.w3.org/2019/09/24-aria-apg-minutes.html#item04) that some language has been discussed, but I feel it is not strong enough and reads only as a standard disclaimer.

This has overlap with #1150, but I feel goes further.

Proposed text:

> Warning: This widget may have support gaps that can only be filled by the development of new technologies. This widget is not tested across all [browser and AT combinations](https://www.w3.org/TR/wai-aria-practices-1.1/#browser_and_AT_support), in particular [mobile/touch devices](https://www.w3.org/TR/wai-aria-practices-1.1/#mobile_and_touch_support), and represents [hypothetical uses of ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#no_aria_better_bad_aria), not necessarily best practices. This widget is not production-ready. Additionally, consult the [First Rule of ARIA](https://www.w3.org/TR/using-aria/#rule1) prior to choosing a pattern.

mcking65:
Resolves #1228

#### Preview link

[View in compare branch for this PR](https://raw.githack.com/w3c/aria-practices/1228-support-notice/aria-practices.html)
[View how it appears in example page](https://raw.githack.com/w3c/aria-practices/1228-support-notice/examples/checkbox/checkbox-1/checkbox-1.html)


a11ydoer:
APG Group decision - Move the notice under H1

DavidMacDonald:
+1 to text

PR #1413, opened 2020-05-26
Add combobox date picker example for issue 34 

mcking65:
Adds the combobox datepicker example described in issue #34. Incorporates work from #1017.

#### Preview Link

[Preview of date picker combobox example](https://raw.githack.com/w3c/aria-practices/issue34-add-combobox-datepicker/examples/combobox/combobox-datepicker.html)

### Review checklist

* [x] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @mcking65 @carmacleod 
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @curtbellew 
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @spectranaut)
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @zcorpan)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1413.html" title="Last updated on Jul 7, 2020, 2:48 PM UTC (5774347)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1413/390302c...5774347.html" title="Last updated on Jul 7, 2020, 2:48 PM UTC (5774347)">Diff</a>


curtbellew:
I ran through some testing tools, screen reader, keyboard and high contrast tests.  Everything looks fantastic but I had one question.   The use of the abbr attribute in the column header TH's.  For instance the demo uses abbr="Sunday" when the TH text is "Su".  I was thinking that abbr in the TH had been deprecated in HTML5.   Also, I have always thought of abbr as the abbreviation of the long content so abbr="Su" for the TH text "Sunday" instead.  I usually recommend using aria-label for the long versions of the column header in cases where we want longer text to be read out by screen readers.

jscholes:
I'm glad to see a functional datepicker example be included for the combo box pattern.  There are a few usability concerns that I'm perceiving based on testing the preview link.

#### Date Input

1. The date format uses a placeholder attribute.  Most screen reader users are going to find it difficult (or impossible) to review it character by character, for example to determine whether the collection of lower case Y's represents a 2, 3 or 4 digit year entry.  Plus the other accessibility concerns with placeholder.  This is a common problem with date fields and can be fixed either by:
    * including the date format inline on the page and referencing it via `aria-describedby`; or
    * including it in the label ("Date ("mm/dd/yyyy").
2. The example page states: "A live region provides information on how to open the date picker dialog when textbox receives focus (e.g. using down arrow key)".  However, this doesn't work for me with the latest stable builds of both NVDA and Chrome, and seems like a suboptimal way of providing hint text.  Again, `aria-describedby` would be the ideal candidate here, allowing JAWS users to turn off hints for example.

#### Modal

1. A live region is used to communicate the new month when crossing a boundary.  However, as the grid and modal containers are also aria-labelledby the heading, NVDA speaks the information twice and JAWS speaks it three times (latest stable builds of each):
    * JAWS: "April 2020 Modal dialog. April 2020 Grid. Th 30. April 2020"
    * NVDA: "April 2020. 30  Th. April 2020"
2. Because this follows the grid pattern rather than using an application role with a table of buttons, NVDA users must press Escape twice to dismiss the modal.  Once to exit focus mode, and once more to actually pass the keystroke through to the page.
3. While navigating between the cells using any of the arrow keys, it consistently takes an entire second for JAWS to announce the new destination cell.
4. The use of abbreviated day names is a horrible UX for screen reader users.  As @curtbellew pointed out, having the full day name inside the `abbr` attribute on each `<th>` seems backwards, but screen readers seem to ignore it regardless.  From a design perspective, I predict that devs and designers will want abbreviated column headers, but we should probably find a reliable way to present the full text to a screen reader so that the pattern can do the right thing out of the box.
5. The current date isn't marked with `aria-current="date"`.  Granted, it doesn't seem to be made visually apparent either, but again this is something designers will probably patch in without adding the appropriate programmatic indication.
6. The "Cancel" button is positioned before "OK", which users will only expect if they're accustomed to working on MacOS or iOS.
7. The help text, "Cursor keys can navigate dates", doesn't communicate any of the other supported keystrokes such as Page Up/Down and Home/End.  As with the help message for the date input, I also wonder if there's a reason not to apply this to the grid via `aria-describedby`?

JAWS-test:
>  I was thinking that abbr in the TH had been deprecated in HTML5.
> Also, I have always thought of abbr as the abbreviation of the long content so abbr="Su" for the TH text "Sunday" instead

Both is wrong, see https://html.spec.whatwg.org/multipage/tables.html#attr-th-abbr:

> It is typically an abbreviated form of the full header cell, but can also be an expansion, or merely a different phrasing

jscholes:
Some further observations:

1. The example page states: "The date picker dialog is opened by ... moving keyboard focus to the combobox and pressing Enter".  Enter is not listed as having this function as part of the Combo Box design pattern, not even as an optional affordance.  Encouraging web authors to implement it in this way will prevent default, quick form submission for users who prefer to type a date manually and then hit Enter.
2. When swiping through dates in the table with iOS, the column headers aren't announced at all.  This makes it seem like just a grid of numbers, which has long been an accessibility antipattern for calendars.  When combined with the apparent lack of support for the `abbr` attribute on `<th>` by screen readers, I think the cells must have more descriptive accessible names.

curtbellew:
> > I was thinking that abbr in the TH had been deprecated in HTML5.
> > Also, I have always thought of abbr as the abbreviation of the long content so abbr="Su" for the TH text "Sunday" instead
> 
> Both is wrong, see https://html.spec.whatwg.org/multipage/tables.html#attr-th-abbr:
> 
> > It is typically an abbreviated form of the full header cell, but can also be an expansion, or merely a different phrasing

I think I've looked at an older version of the standard when I've researched this in the past -- https://www.w3.org/TR/2010/WD-html-markup-20100624/th.html .   It looks like updated versions indicate that it's not been deprecated.  Sorry about that.  The second part of your quote is pretty much what I said above so I'm not following.  In any case, I don't see this as any sort of violation since abbr hasn't been deprecated.   I'd still recommend using aria-label.  Thanks!

carmacleod:
If you click on the drop-down arrow with the mouse, focus stays on the input.
Should it instead go to the grid so that arrow keys can be used to navigate?

jongund:
Made the following changes:

* Removed placeholder attribute from combobox
* Added description to combobox for date format
* Removed information from the documentation about a live region describing how to open the dialog from the combobox
* Added a test case for the `aria-describedby` attribute on the combobox

examples/combobox/css/combobox-datepicker.css
```diff
@@ -0,0 +1,257 @@
...
+
+.combobox-datepicker .group.focus {
+  outline: 2px solid black;
+  outline-offset: 1px;
+}
+
+.combobox-datepicker .group.focus input,
+.combobox-datepicker .group.focus button {
+  background-color: #DEF;
+}
```
Cerberooo, line 70:
```suggestion
```

Same as at line 55.

examples/combobox/css/combobox-datepicker.css
```diff
@@ -0,0 +1,257 @@
...
+
+.combobox-datepicker .group button[aria-expanded="true"] svg {
+  transform: rotate(180deg) translate(0, -1px);
+}
+
+.combobox-datepicker .group.focus {
+  outline: 2px solid black;
+  outline-offset: 1px;
+}
+
```
Cerberooo, line 66:
```suggestion
```

Same as at line 48.

examples/combobox/js/combobox-datepicker.js
```diff
@@ -0,0 +1,843 @@
...
+
+var ComboboxDatePicker = function (cdp) {
+  this.buttonLabel = 'Date';
+  this.dayLabels = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+  this.monthLabels = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+
+  this.messageCursorKeys = 'Cursor keys can navigate dates';
+  this.lastMessage = '';
+
+  this.comboboxNode   = cdp.querySelector('input[type="text"');
```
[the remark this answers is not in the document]
jongund: fixed

examples/combobox/js/combobox-datepicker.js
```diff
@@ -0,0 +1,843 @@
+/*
+*   This content is licensed according to the W3C Software License at
+*   https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
+*
+*   File:   ComboboxDatePicker.js
+*/
+
+var ComboboxDatePicker = function (cdp) {
+  this.buttonLabel = 'Date';
+  this.dayLabels = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
```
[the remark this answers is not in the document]
jongund: removed

examples/combobox/css/combobox-datepicker.css
jongund: removed

examples/combobox/js/combobox-datepicker.js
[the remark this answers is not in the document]
jongund: Thank you for the feedback, I have made the changes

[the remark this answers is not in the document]
Cerberooo: No problem, thanks for your work! :)

examples/dialog-modal/datepicker-dialog.html
```diff
@@ -127,15 +128,6 @@ <h2 id="ex_label">Example</h2>
...
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">2</button></td>
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">3</button></td>
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">4</button></td>
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">5</button></td>
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">6</button></td>
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">7</button></td>
-                <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">8</button></td>
-              </tr>
               <tr>
                 <td class="dateCell"><button type="button" class="dateButton" tabindex="-1">9</button></td>
```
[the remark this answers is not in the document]
jongund: @spectranaut 

PR 1362 "Update menu button datepicker dialog example"  branch for updating the modal dialog datepicker, this PR 1413 is only updating Combobox Date Picker example.

jongund:
@spectranaut 
@mcking65 

I have fixed the spacing issue with combobox-datepicker.html and removed datepicker-dialog.html from being updated in this pull request, it is updated in a different pull request.

mcking65:
@jongund, this is close. There are some changes I think we need to make before merging, and some that can wait and be addressed as separate issues after merge.

I think we can merge by than July 10. We will use the July 7 meeting to resolve anything that we have not resolved asynchronously this week.

After reviewing the experience and the feedback above, here are some changes I think we need to make in this branch. I explain the reasons below.
1. Change the down arrow button (line 61 in HTML) as follows:
   * Change aria-label from "Date" to "Choose Date"
   * Remove aria-expanded, aria-controls, and aria-haspopup and adjust tests and documentation of states and props accordingly.
   * Remove outdated documentation about dynamic changes to the button label.
2. Move the `div` with id `cb-description` to be between the input and the button in the reading order.
3. Remove the Enter key behavior from the combobox and adjust documentation accordingly.
4. Make clicking the down arrow icon place focus inside the dialog in the same way that pressing down arrow does.
5. Instead of having the dialog labeled by the current month, put `aria-label="Choose Date"` on the dialog.

Why these changes?

First, the button that opens the dialog, other than the fact that it is not keyboard focusable, should be consistent with other buttons that open dialogs. We do not use aria-haspopup because that changes the role of the button to menubutton, even if the value of haspopup is dialog. That is a limitation of certain accessibility APIs. The value of `dialog` for haspopup is intended only for the combobox element itself.

This button is different from the down arrow buttons for comboboxes that popup a listbox or grid because those popup elements are not modal. When the popup is not modal, or when a popup can be open without containing visual focus, a touch user can collapse the combobox by activating the down arrow icon. Because of that, we use aria-expanded on the button so that touch screen reader users can readily know the state of the popup from the button and get feedback from activating the button. In the case of a dialog, screen readers inhibit access to the button when the dialog is open, and the dialog itself includes a cancel button for closing via touch.

The label change on the button is to give screen reader users better understanding of the purpose of the button, and it matches the name I proposed for the dialog.

The accessible name change on the dialog is for three reasons. First, changing the title of the dialog dynamically is one of the reasons for the excess speech described in earlier feedback. The second reason is that a label of "Choose Date" on the dialog gives better feedback to the user about where they are, keeping in mind that the grid is labeled by the month and year. Finally, it makes it possible for the button that opens the dialog and the dialog to have the same accessible name, which is consistent with what we recommend for dialogs.

I hope the change to reading order for the combobox description div does not complicate visual positioning. In general, I find that it is a much better screen reader experience, especially with touch, when such instructions are immediately adjacent to the input. This is especially true if we later add some error checking and error messages could be displayed there.

Finally, moving the focus into the dialog when the down arrow icon is clicked as Carolyn suggested is important. That makes it consistent with other buttons that open dialogs. This is particularly important for screen reader users who activate the button with a reading cursor or via touch on a mobile device.

From an editorial copy perspective, I plan to make changes to substantially trim the accessibility features section so it is consistent with how we draft that section in other examples. It will cover only accessibility features with explainations of the ways in which they improve accessibility. However, I WILL NOT commit those changes in this branch. I'll make a separate branch with a PR requesting merge into this branch so we can iron out the content of the accessibility features section without cluttering this PR. So, as you make the above changes, please don't touch the accessibility features section; only change documentation in the tables after it.

There is one other topic that I'd like to discuss in the July 7 meeting -- revealing keyboard functionality in the UI. Please do not make any changes in this branch related to this topic. If the group arrives at some consensus, we will land those changes separately. Basically, I think that we should reveal the keyboard assignments for page up/down and shift+pageUp/Down on their respective buttons. Perhaps we do it in a manner similar to how we reveal the labels on the icons in the editor toolbar example. Or, possibly it is in a help field that appears at the bottom of the dialog when one of those buttons has focus. Related to this, I'm still not a fan of the live region that announces the cursor key instruction. If we keep that, we need to improve the wording. However, I think those types of instructions are the types of instructions that screen readers could choose to provide for beginner users, just like they tell users what keys to press in a menu or listbox. We could have aria-at assertions related to the provision of such instructions.

jongund:
@mcking65 

I have checked in the changes base on your last comments.
Did you want to use `aria-current="date"` instead of `aria-selected="true"` to indicate the currently selected date in the grid?

mcking65:
No, we don't want to use aria-current. We discussed that as a group, and current is not the same as selected.

In the case of a day, current could be understanding as meaning this is the day that represents "today". So, if it can move around, that could be thought of as a bug.

Selected is consistent with the combobox pattern, so that is what we are using to indicate the value in the field.

mcking65:
@jongund the changes are making a huge difference, thank you!!

jscholes:
@mcking65 Any thoughts on the abbreviated day names?  I agree with all of the changes you outlined, but this still remains a sticking point in my mind and a significant usability issue.

mcking65:
@jscholes wrote:

>Any thoughts on the abbreviated day names? I agree with all of the changes you outlined, but this still remains a sticking point in my mind and a significant usability issue.

Day names are frequently abbreviated in practice so demonstrating how to improve accessibility when they are abbreviated is important.

Using aria-label to override visible content is a pernitious practice that we do not want to encourage.

The abbr method for providing a more friendly column title when reading inside tables has been the preferred method for years and is consistent with standards. Screen readers should support it when users are interacting inside of grids just as they support it when reading in static tables.

The APG aims to promote practices that are consistent with the standards that, when properly supported by browsers and assistive technologies, will yield the best results for users. Part of the goal is to provide the reference examples that can be used by both browser and assistive technology developers.

The [ARIA-AT project](https://aria-at.w3.org) is measuring the quality of assistive technology support. It will have assertions related to supporting abbr in this example. So, I'm confident that this usability concern will soon be resolved by improved screen reader support.

In the meantime, whether or not authors abbreviate the column titles, I am really excited to have this example, and the date picker dialog example, available to promote.

mcking65:
@jscholes, thank you, BTW, for the detailed feedback. It was helpful!


> 1. The date format uses a placeholder attribute.  Most screen reader users are going to find it difficult (or impossible) to review it character by character, for example to determine whether the collection of lower case Y's represents a 2, 3 or 4 digit year entry.  Plus the other accessibility concerns with placeholder.  This is a common problem with date fields and can be fixed either by:
>    
>    * including the date format inline on the page and referencing it via `aria-describedby`; or
>    * including it in the label ("Date ("mm/dd/yyyy").

Good call out here. There is broad consensus on the problems caused by placeholders, and we should have caught that sooner. We don't recommend putting instructions in labels. We have adopted the aria-describedby alternative.


> 2. The example page states: "A live region provides information on how to open the date picker dialog when textbox receives focus (e.g. using down arrow key)".  However, this doesn't work for me with the latest stable builds of both NVDA and Chrome, and seems like a suboptimal way of providing hint text.  Again, `aria-describedby` would be the ideal candidate here, allowing JAWS users to turn off hints for example.

That documentation should have been deleted. We removed that live region a while ago.
> 

> 1. A live region is used to communicate the new month when crossing a boundary.  However, as the grid and modal containers are also aria-labelledby the heading, NVDA speaks the information twice and JAWS speaks it three times

Excellent point. The dialog is no longer labeled by the live content; it is now named "Choose Date", which is now also the name of the button that opens it. This fixed the annoying excess verbosity.

> 2. Because this follows the grid pattern rather than using an application role with a table of buttons, NVDA users must press Escape twice to dismiss the modal.  Once to exit focus mode, and once more to actually pass the keystroke through to the page.

This is exactly what we want. NVDA users should be able to switch to browse mode and read inside the dialog.

BTW, please, please never use the application role for that purpose. We rewrote the application role for ARIA 1.1, so that would no longer be a legitimate use.

> 3. While navigating between the cells using any of the arrow keys, it consistently takes an entire second for JAWS to announce the new destination cell.

I can't reproduce this. If this were reproduceable and unique to JAWS, I would consider it a JAWS problem.

> 4. The use of abbreviated day names is a horrible UX for screen reader users.  As @curtbellew pointed out, having the full day name inside the `abbr` attribute on each `<th>` seems backwards, but screen readers seem to ignore it regardless.  From a design perspective, I predict that devs and designers will want abbreviated column headers, but we should probably find a reliable way to present the full text to a screen reader so that the pattern can do the right thing out of the box.

As explained above, we expect this to be resolved by screen readers providing the same support for abbr on column headers and row headers that they do for `th` in tables. We will drive this through the aria-at project.

> 5. The current date isn't marked with `aria-current="date"`.  Granted, it doesn't seem to be made visually apparent either, but again this is something designers will probably patch in without adding the appropriate programmatic indication.

If there is a date in the input, that date is marked as selected in the grid. The task force does not recommend using aria-current for this purpose; that would not be consistent with the spec for aria-current, especially given that this grid is opened from a combobox.

> 6. The "Cancel" button is positioned before "OK", which users will only expect if they're accustomed to working on MacOS or iOS.

True enough ... the APG is not opinionated about such matters.

> 7. The help text, "Cursor keys can navigate dates", doesn't communicate any of the other supported keystrokes such as Page Up/Down and Home/End.  As with the help message for the date input, I also wonder if there's a reason not to apply this to the grid via `aria-describedby`?

Good point. We will improve keyboard command discoverability after merge. This is tracked by issue #1434.

Again, many thanks for all the time and effort you put into providing feedback!! The  example is noticeably better as a result.

examples/combobox/js/combobox-datepicker.js
```diff
@@ -0,0 +1,842 @@
...
+
+  if (str !== this.lastMessage) {
+    setTimeout(setMessageDelayed.bind(this), 200);
+    this.lastMessage = str;
+  }
+};
+
+ComboboxDatePicker.prototype.setFocusCombobox = function () {
+  this.comboboxNode.focus();
+};
```
Cerberooo, line 727: No longer needed since @mcking65's notes.

jongund:
@mcking65 
The example does not use the touchstart event to suppress they onscreen keyboard on mobile devices.  I am not sure we want to do this anyway, since right now we do allow text entry into the textbox.

examples/combobox/js/combobox-datepicker.js
```diff
@@ -0,0 +1,842 @@
...
+
+ComboboxDatePicker.prototype.updateDay = function (day) {
+  var d = this.focusDay;
+  this.focusDay = day;
+  if (this.isNotSameMonth(d, day)) {
+    this.updateGrid();
+    this.setFocusDay();
+  }
+};
```
Cerberooo, line 174: Isn't used anywhere.

examples/combobox/js/combobox-datepicker.js
jongund: Thanks, I removed the unused methods

examples/combobox/js/combobox-datepicker.js
jongund: Thanks, I removed the unused methods

examples/combobox/combobox-datepicker.html
```diff
@@ -214,14 +214,15 @@ <h3>Date Textbox</h3>
     <ul>
       <li>Contains the date value.</li>
       <li>Opens and closes the date picker dialog box through keyboard, click and touch events.</li>
-      <li>To support people who are sighted in understanding the <kbd>down arrow</kbd> key opens the date picker dialog box, a down arrow icon appears in the <code>textbox</code> when the <code>textbox</code> receives keyboard focus and the down arrow icon is removed when the <code>textbox</code> does not have keyboard focus.</li>
+      <li>A down arrow icon in the <code>textbox</code> provides a visual affordance that the <kbd>down arrow</kbd> key opens the date picker dialog box.</li>
```
mcking65, line 217: @jongund 
These lines are going away. If you want to comment on or push changes to the accessibility features section, please do so in pr #1435

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,781 @@
...
+          <input type="text"
+                id="cb-textbox-1"
+                aria-autocomplete="none"
+                role="combobox"
+                aria-expanded="false"
+                aria-haspopup="dialog"
+                aria-controls="cb-dialog-1"
+                aria-describedby="cb-description-1">
+
+          <span class="desc" id="cb-description-1">(date format: mm/dd/yy)</span>
```
carmacleod, line 71: Date format needs to be 4-digit date, because if you type a 2-digit date, you'll end up back in the 20th century (i.e. in the 1900's). 😄 
```suggestion
          <span class="desc" id="cb-description-1">(date format: mm/dd/yyyy)</span>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,781 @@
...
+      <li>A down arrow icon in the <code>textbox</code> provides a visual affordance that the <kbd>down arrow</kbd> key opens the date picker dialog box.</li>
+      <li>A description provides information on the date format to assistive technologies.</li>
+    </ul>
+
+    <h3>Choose Date Button</h3>
+
+    <ul>
+      <li>The down arrow icon is a button to open and close the date picker dialog box through click events.</li>
+      <li>The arrow points down when the dialog box is closed and it points up when the dialog box is open.</li>
+      <li>Excluded from the <kbd>tab</kbd> sequence of the page (eg. <code>tabindex=-1</code>), since the user can use the <kbd>down arrow</kbd> or <kbd>enter</kbd> key to open the date picker dialog box from the <code>textbox</code>.</li>
```
carmacleod, line 236: use e.g. instead of eg.

```suggestion
      <li>Excluded from the <kbd>tab</kbd> sequence of the page (e.g. <code>tabindex=-1</code>), since the user can use the <kbd>down arrow</kbd> or <kbd>enter</kbd> key to open the date picker dialog box from the <code>textbox</code>.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,781 @@
...
+      <li>The date picker dialog is a modal dialog box for selecting a date from a grid of dates for a particular month and year.</li>
+      <li>Additional buttons and keyboard commands are used to change month and year.</li>
+      <li>A live region announces changes in the month and year.</li>
+      <li>A live region announces the use of cursor keys when a date in the grid of dates gets focus.  The message is also visible at the bottom of the dialog box to provide the same information to keyboard users.</li>
+      <li>The dialog is opened through keyboard commands in the textbox or by clicking on the choose date button.</li>
+      <li>High contrast support for focus and hover styling of the controls in the dialog box use the CSS border property:
+        <ul>
+          <li>When a button or date receives focus a border is added.</li>
+          <li>When hovering over a button or date with a pointing device a border is added.</li>
+          <li>By default buttons and dates do not have a border, padding is used as a placeholder for the added border for focus and hover styling.</li>
```
carmacleod, line 251: Add comma, and split into 2 sentences.

```suggestion
          <li>By default, buttons and dates do not have a border. Padding is used as a placeholder for the added border for focus and hover styling.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,781 @@
...
+            <th><kbd>Right Arrow</kbd></th>
+            <td>Moves focus to the next day.</td>
+          </tr>
+          <tr data-test-id="grid-left-arrow">
+            <th><kbd>Left Arrow</kbd></th>
+            <td>Moves focus to the previous day.</td>
+          </tr>
+          <tr data-test-id="grid-home">
+            <th><kbd>Home</kbd></th>
+            <td>Moves focus to the first day (e.g Sunday) of the current week.</td>
```
carmacleod, line 398: add missing .

```suggestion
            <td>Moves focus to the first day (e.g. Sunday) of the current week.</td>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,781 @@
...
+              </tr>
+            </thead>
+            <tbody>
+              <tr data-test-id="dialog-role">
+                <th scope="row"><code>dialog</code></th>
+                <td></td>
+                <td>
+                  <code>div</code>
+                </td>
+                <td>Identifies the element as a dialog .</td>
```
carmacleod, line 597: delete space
```suggestion
                <td>Identifies the element as a dialog.</td>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,781 @@
...
+                        <li>Only set on the gridcell representing the currently selected date, no other gridcells have <code>aria-selected</code> specified.</li>
+                      </ul>
+                    </td>
+                  </tr>
+              </tbody>
+            </table>
+            <p>
+              <strong>Note:</strong> Since the names of the days of the week in the column headers are abbreviated to two characters, they may be difficult to understand when announced by a screen reader.
+              An alternative column header name can be provided to screen readers by applying the <code>abbr</code> attribute to the <code>th</code> elements.
+              So, each <code>th</code> element includes a <code>abbr</code> attribute containing the full spelling of the name of the day for that column.
```
carmacleod, line 747: a -> an
```suggestion
              So, each <code>th</code> element includes an <code>abbr</code> attribute containing the full spelling of the name of the day for that column.
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,782 @@
...
+                </tr>
+                <tr data-test-id="calendar-navigation-aria-live">
+                    <td></td>
+                    <th scope="row"><code>aria-live=<q>polite</q></code></th>
+                    <td>
+                      <code>h2</code>
+                    </td>
+                    <td>
+                      <ul>
+                        <li>When the month and/or year changes the content of the <code>h2</code> element is updated.</li>
```
carmacleod, line 655: Add comma
```suggestion
                        <li>When the month and/or year changes the content of the <code>h2</code> element is updated.</li>
```

examples/combobox/combobox-datepicker.html
```diff
@@ -0,0 +1,782 @@
...
+                  <tr data-test-id="grid-role">
+                    <th scope="row"><code>grid</code></th>
+                    <td></td>
+                    <td>
+                      <code>table</code>
+                    </td>
+                    <td>
+                     <ul>
+                        <li>Identifies the <code>table</code> element as a <code>grid</code> widget.</li>
+                        <li>Since the <code>grid</code> role is applied to a <code>table</code> element, the <code>row</code>, <code>colheader</code>, and <code>gridcell</code> roles do not need to be specified because they are implied by <code>tr</code>, <code>th</code>, and <code>td</code> tags.</li>
```
carmacleod, line 685: change colheader to columnheader
```suggestion
                        <li>Since the <code>grid</code> role is applied to a <code>table</code> element, the <code>row</code>, <code>columnheader</code>, and <code>gridcell</code> roles do not need to be specified because they are implied by <code>tr</code>, <code>th</code>, and <code>td</code> tags.</li>
```

a11ydoer:
Other than a space key issue in “date picker dialog: date grid” – it comes with the behavior of scrolling and jumping down, the example looks like a well-thought-out implementaiton.

jongund:
@mcking65 
Removed test for keyboard support regression tests for the SPACE and ENTER key for the "Choose Date" Button, since the reference in the example has been removed.

PR #1418, opened 2020-06-03
Add file inputs to focusable elements

MartijnCuppens:
Not sure why this was ever skipped. I've tested this by changing the type of the first input field here: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

Before: 
image

After:
image

### DIFF
--- a/examples/js/utils.js
+++ b/examples/js/utils.js
@@ -75,7 +75,7 @@ aria.Utils.isFocusable = function (element) {
     case 'A':
       return !!element.href && element.rel != 'ignore';
     case 'INPUT':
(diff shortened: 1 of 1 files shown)


mcking65:
@tatermelon Could you review this? Unclear why file inputs were treated like hidden elements.

tatermelon:
Strange, I don't see any good reason why file inputs should be skipped. This fix looks right to me!

mcking65:
thank you @tatermelon

mcking65:
Merging via #1470

PR #1419, opened 2020-06-03
Prevent false positives for elements with a positive tabindex

MartijnCuppens:
Elements with a positive tabIndex were always considered to be focusable while `disabled` could prevent this. This PR just checks if the tabIndex is positive.

Since [CanIUse](https://caniuse.com/#feat=mdn-api_htmlelement_tabindex) mentions something about issues with tabIndex, I did some additional tests for IE with https://jsbin.com/jayokocena/edit?html,js,output:

<details>
<summary>
Tests in IE
</summary>

Test in IE11:
image

Test in IE10:
image

Test in IE9:
image
</details>


mcking65:
@MartijnCuppens

Sorry, I don't understand this change. I can't see what is in the images you included, but looking at the code, this would cause the function to return false for elements with tabindex of 0 or greater, which is the opposite of what we want.

BTW, APG is no longer supporting IE, so we do not add code specifically and only for the purpose of creating IE compatibility.

MartijnCuppens:
> APG is no longer supporting IE

In that case, ignore everything I said about IE.

> this would cause the function to return false for elements with tabindex of 0 or greater

Nope, exactly the opposite. Notice I also changed the return value.


With the current implementation, `isFocusable()` would return true if the element has a tabindex greater than 0, which could be false because we don't check the disabled state etc.

PR #1429, opened 2020-06-25
Linter: Catch "no undef" rules

spectranaut:
I'm just curious if there is support for this change!  Now that we are using "no strict" it would be nice to catch undefs in the linter, rather than in edge cases in the application.

This PR will fail in the CI for two reasons (for now):
1. There are some undef's fixed in #1416 (unmerged)
2. The js files in the example reference globals defined in other files. For eslint to pass, we will have to warn eslint about these variables using [this method](https://eslint.org/docs/user-guide/configuring#specifying-globals). I wanted to make sure there was support for the change before doing the work.


.eslintrc.json
```diff
@@ -1,14 +1,16 @@
...
+    "node": true,
+    "es6": true
   },
   "parserOptions": {
     "ecmaVersion": 6
   },
   "rules": {
     "no-unused-vars": 0,
-    "no-undef": 0,
+    "no-undef": "error",
```
nschonni, line 13: Think this line can just get deleted since it's in the "recommended" rules https://eslint.org/docs/rules/no-undef

.eslintrc.json
```diff
@@ -1,14 +1,16 @@
 {
   "extends": "eslint:recommended",
   "env": {
-    "browser": true
+    "browser": true,
+    "node": true,
```
nschonni, line 5: Don't think `node` should be on, since it might not flag things the examples. The `scripts` folder already has it one since those are nodejs scripts

spectranaut:
Whoops I forgot about this PR @nschonni just got to your comments!

nschonni (approved):

test/.eslintrc.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../.eslintrc.json",
+  "env": {
```
nschonni, line 3: Alternately, this file could be deleted, and the config hoisted to an `overrides` section in the root config like https://eslint.org/docs/user-guide/configuring#specifying-processor

spectranaut:
Hi @mcking65 you asked to understand the options a bit better here: If we want to merge this js lint error (catching any new undefined variables), we would have to edit 30 or so files to add a eslint comment specifically listing which globals they should not error on. A lot of those undefs come from the old way APG example code was structured, in multiple files, but Jon has been changing that example by example. It will take a while for all those rewrites to be completed, which is why I want to just go through and add the ignore comment to all these files for now!

spectranaut:
Closing this PR in favor of @nschonni's: https://github.com/w3c/aria-practices/pull/1610/files

PR #1432, opened 2020-07-01
Select-only Combobox Example: Add Accessibility Features section

carmacleod:
Ticks the checkbox in https://github.com/w3c/aria-practices/pull/1396#issuecomment-627738460 regarding mentioning the places where this diverges from html select behavior in an Accessibility Features section.

[Preview in the compare branch for this PR](https://raw.githack.com/w3c/aria-practices/pr1396-accessibility-features/examples/combobox/combobox-select-only.html)

PR #1439, opened 2020-07-08
PR1432 alternative wording

mcking65:
@carmacleod 

Thank you for #1432. Here is an idea for wording it to present the divergence as part of the accessibility features of the example. I'm not sure if I have the description of how mouse users cancel quite right.

Just an idea ... incorporate as you think best.

### DIFF
--- a/examples/combobox/combobox-select-only.html
+++ b/examples/combobox/combobox-select-only.html
@@ -69,14 +69,17 @@
 
     <section>
       <h2>Accessibility Features</h2>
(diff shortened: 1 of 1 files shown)


carmacleod:
@mcking65 
You are correct that mouse users can close and cancel by clicking back on the input.
Your new words are good. I'll merge them. 
I was hoping someone would wordsmith the original - I just wanted to get it out there before the thoughts were gone. 😄 
Do I "squash & merge"? Or "rebase and merge"?

Issue #1450, opened 2020-07-15
Use parallelization for running AVA tests


spectranaut:
When we used AVA's internal parallelization (the tests are run in separate process) we got too many "NoSuchSessionErrors" from selenium, so we can to set concurrency to 1.

I thought we should try using Travis's parallelization instead, which AVA has a hook build into for. In this case, we can run sections of the AVA tests in separate VMs instead of separate process. I start to experiment with this in a PR on the Bocoup branch: https://github.com/bocoup/aria-practices/pull/21

I'd like to test it more before opening a PR, but the tests ran in 6 minutes instead of 20, I'm confused why the speed up is so great and afraid not all tests are running. This is the main change to `.travis.ci`:

```
    - stage: Test
      name: AVA Regression Tests
      addons:
        firefox: latest
      script: scripts/regression-tests.sh
      env:
        - TEST_WAIT_TIME=1000
        - CI_NODE_TOTAL=2
        - CI_NODE_INDEX=0

    - stage: Test
      name: AVA Regression Tests
      addons:
        firefox: latest
      script: scripts/regression-tests.sh
      env:
        - TEST_WAIT_TIME=1000
        - CI_NODE_TOTAL=2
        - CI_NODE_INDEX=1
```

mcking65:
Wow, that would be a huge help to have that kind of improvement.

Could some of the performance difference be due to some differences in configurations or settings at the organization (w3c vs Bocoup) or account level? I have 0 knowledge of what is configurable in the GitHub/Travis integration, if anything. Also, when an organization utilizes Travis, are there different levels or types of service that may be purchased?

Just wondering if testing a repo in the Bocoup account creates an apples to apples comparison.

spectranaut:
We can run the same test on w3c pretty easily! Just edit the travis file with the code I put above and also `tests/index.js` to trigger all the tests to run and open a PR.

Issue #1449, opened 2020-07-15
Bugs with regression test script: tests running when they should not, and not running when they should


spectranaut:
I noticed two weird problems with [this script](https://github.com/w3c/aria-practices/blob/master/scripts/regression-tests.sh) that handles running regression tests in travis:
1. In PR #1441, the regression tests are running although they should not. There is logic specifically to avoid running the regression tests if no examples with tests have been edited.
2. In PR #1424 no regression tests are run because the script errors before getting to the ava command, and the regression check passes. We should have the check fail in this case, but we also need to debug why it errors.

spectranaut:
I've looked into both issues and have fixes, hopefully will open a PR tomorrow!

For bug part 1: There simply needs to be a little extra logic for when there is change in the landmark directory only -- this should be straight forward.

For bug part 2: The issue turns out to be related to force pushing a branch and using the environment variable `TRAVIS_COMMIT_RANGE`, which no longer represents a valid commit range after a force push (see https://github.com/travis-ci/travis-ci/issues/2668). I think we can replace this variable with `master...$TRAVIS_BRANCH` but I'll need to test that this works on the PR and branch builds.

PR #1465, opened 2020-07-23
Run regression tests in parallel in TravisCI and other regression tests improvements

spectranaut:
With this PR, the regression tests run in 6 minutes instead of 20!

I had to do three things in this PR to accomplish the parallelization discussed in issue https://github.com/w3c/aria-practices/issues/1450:
1. Updates `.travis.yml` to use the [Ava feature](https://github.com/avajs/ava#parallel-runs-in-ci) that allows tests in parallel across three VMs in the CI
2. Updates Ava version to include [this bug fix](https://github.com/avajs/ava/issues/2461) -- the fix prevents a false error when Ava tries to parallelize less than three test files (it parallelizes by files, so one test file is not parallelized).
3. Updates the script that runs the regression tests in the CI to only the test files for the example or test file you are editing to take advantage of the parallelization option. Before, it filtered tests by using a regex on the name of the test. Now, it explicitly lists all the test FILES that should be run.

I also pushed a change to fix the two bugs discussed in https://github.com/w3c/aria-practices/issues/1449 because it also involves changes to the regression test CI script.
1. You can confirm the fix for the first bug by seeing that the regression tests are not run on [the branch build for the b4d3f75 commit](https://travis-ci.com/github/w3c/aria-practices/jobs/364415251) (this is because I have force pushed to this branch so $TRAVIS_COMMIT_RANGE is not a valid value), and then see that they are run in both [the branch build](https://travis-ci.com/github/w3c/aria-practices/jobs/364728657) and [the PR build](https://travis-ci.com/github/w3c/aria-practices/jobs/364728685) for the latest commit 4d01c09 (which contains the bug fix).
2. You can confirm the fix for the second bug by seeing that the tests are not run in [this test PR in the Bocoup fork](https://github.com/bocoup/aria-practices/pull/25), where only an landmark file is edited.

Also I added one other thing to make the regression test script easier to test locally. If you run `bash regression-tests.sh` from a branch that is not master, it will run the script as if your current branch is a pull request against master. So you can test locally that file changes in your branch will trigger the correct tests. The logic is as follows:
1. If `package.json` has been edited, or if test utility files have been edited, or if example utility files have been edited, then all regression tests will be run in the CI.
2. Otherwise, if an example is edited, then the regression tests for all of the examples contained in the same example directory are run. For example: if you edit one of the two checkbox examples, then both the checkbox-1 and checkbox-2 example regression tests will be run in the CI.
3. All test files that have been edited will be run in the CI.


spectranaut:
There is an unrelated HTML error: https://github.com/w3c/aria-practices/issues/1467

PR #1354, opened 2020-03-17
chore: Update vnu-jar to 20.3.16

nschonni:
Fixes #1346
The message for img looks like it changed in the release

Issue #1467, opened 2020-07-23
New HTML lint error causes CI to fail


spectranaut:
The HTML linter seems to have changed the errors it throws related to missing "alt" tags on "img" elements. This fails on the [multithumb slider example](https://w3c.github.io/aria-practices/examples/slider/multithumb-slider.html). The lint error are:
-  `An “img” element with no “alt” attribute must not have a “role” attribute.`
-  `An “img” element with no “alt” attribute must not have any “aria-*” attributes other than “aria-hidden”.`
- `An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.`

The version of vnu-jar I'm running is: 20.6.30 (which is only a little newer then the last time this string changed in 20.3.16 -- see #1354

```
> aria-practices@0.0.0 vnu-jar /home/spectranaut/w3c/aria-practices
> java -jar node_modules/vnu-jar/build/dist/vnu.jar --errors-only --filterfile .vnurc --no-langdetect --skip-non-html *.html examples/

"file:/home/spectranaut/w3c/aria-practices/examples/slider/multithumb-slider.html":57.19-66.54: error: An “img” element with no “alt” attribute must not have a “role” attribute.
"file:/home/spectranaut/w3c/aria-practices/examples/slider/multithumb-slider.html":57.19-66.54: error: An “img” element with no “alt” attribute must not have any “aria-*” attributes other than “aria-hidden”.
"file:/home/spectranaut/w3c/aria-practices/examples/slider/multithumb-slider.html":57.19-66.54: error: An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.
"file:/home/spectranaut/w3c/aria-practices/examples/slider/multithumb-slider.html":67.19-76.54: error: An “img” element with no “alt” attribute must not have a “role” attribute.
"file:/home/spectranaut/w3c/aria-practices/examples/slider/multithumb-slider.html":67.19-76.54: error: An “img” element with no “alt” attribute must not have any “aria-*” attributes other than “aria-hidden”.
"file:/home/spectranaut/w3c/aria-practices/examples/slider/multithumb-slider.html":67.19-76.54: error: An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.
[...]

PR #1468, opened 2020-07-23
Update vnujar filtering to ignore no alt on role=img, again

spectranaut:
There are more errors no on "multithumb-slider" related to the alt text, I update the regex to filter out all of the alt+img related errors. I wonder if we should open a bug on vnu-jar? Or are our examples actually not valid html?

Also this was fixed recently here: #1354


spectranaut:
@nschonni can you review :) ?

.vnurc
```diff
@@ -4,7 +4,7 @@
 # Proposed, tracking in gh-429
 Bad value “” for attribute “aria-activedescendant” on element “ul”:.*
 # Ignoring for Slider examples because it is forcing rule with role=img
-Element “img” is missing required attribute “alt”.
+.*“img”.*“alt”.*
```
nschonni, line 7: Might be good to just list out the new filter patterns so that any new ones are explicitly suppressed

.vnurc
```diff
@@ -4,7 +4,7 @@
 # Proposed, tracking in gh-429
 Bad value “” for attribute “aria-activedescendant” on element “ul”:.*
```
mcking65, line 5: issue 429 is closed, and the related validator issue is closed too. Can we remove this?

mcking65:
Do I have this right ... when updating vnu.jar from 20.3.16 to 20.6.30, the new version of vnu.jar finds errors related to the alt attribute on elements with role img that the prior version did not find?


Exactly what are the errors?

Do we have an open issue related to these errors to either fix them in the APG example or to track an open issue against the validator?

nschonni:
@mcking65 there is an issue from @spectranaut here #1467 that has the new failure messages

spectranaut:
I can't actually replicated this bug anymore after re-running `npm install`

nschonni:
I was seeing it with the package bump here https://github.com/nschonni/aria-practices/pull/8

spectranaut:
Wow I'm really not seeing clearly today. I thought I was at the latest version of vnu-jar locally but I wasn't

.vnurc
```diff
@@ -5,6 +5,9 @@
...
 # Ignoring for Slider examples because it is forcing rule with role=img
 Element “img” is missing required attribute “alt”.
+An “img” element with no “alt” attribute must not have a “role” attribute.
+An “img” element with no “alt” attribute must not have any “aria-*” attributes other than “aria-hidden”.
```
nschonni, line 9: It may be getting tripped up on the `*`. I'll see if escaping it works locally

nschonni:
> Wow I'm really not seeing clearly today. I thought I was at the latest version of vnu-jar locally but I wasn't

No worries, I know @sideshowbarker sometimes turns around patches pretty quick, so I checked to see that wasn't the case too :)

.vnurc
```diff
@@ -5,6 +5,9 @@
 Bad value “” for attribute “aria-activedescendant” on element “ul”:.*
 # Ignoring for Slider examples because it is forcing rule with role=img
 Element “img” is missing required attribute “alt”.
```
nschonni, line 7: Looks like this one isn't needed anymore
```suggestion
```

sideshowbarker:
> No worries, I know @sideshowbarker sometimes turns around patches pretty quick, so I checked to see that wasn't the case too :)

Is there some pending patch I haven’t merged? And that’s causing problems?

mcking65:
In the multi-thumb slider, we have img elements with aria-label. If that is not valid, we need to change those to use alt. I didn't know that it is invalid to use aria-label instead of alt on img elements.

@sideshowbarker is that a recent change? In ARIA, aria-label is global. And, in ARIA, aria-label is supported on elements with the img role. So, whouldn't it be supported on elements with an implied img role?

sideshowbarker:
> @sideshowbarker is that a recent change?

I don’t know — I’d need to see the exact error or warning messages the HTML checker is emitting. I don’t see any actual error messages cited in this issue. 

> In ARIA, aria-label is global. And, in ARIA, aria-label is supported on elements with the img role. So, whouldn't it be supported on elements with an implied img role?

I don’t know. I’m happy to test with any actual markup for which the HTML checker doesn’t seem to be doing the right thing.

I do know that a while back (last year, I guess), I added a discretionary warning for aria-label on elements that the ARIA working group has been considering to disallow it for. Maybe that warning is what’s being run into in this case?

mcking65:
@sideshowbarker 

The errors are listed in #1467. We get the following errors on the [multithumb slider example](https://github.com/w3c/aria-practices/blob/master/examples/slider/multithumb-slider.html).

* `An “img” element with no “alt” attribute must not have a “role” attribute.`
* `An “img” element with no “alt” attribute must not have any “aria-*” attributes other than “aria-hidden”.`
* `An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.`

spectranaut:
fixed those new comments, @nschonni ! :)

nschonni:
@mcking65 this is the PR that will fix the vnu-jar version, and includes the config fixes

nschonni:
@mcking65 this can land before my other PR to switch to Actions

mcking65:
Unfortunately, as a screen reader user, It's really difficult to use the github actions output on the checks tab when trying to find what failed. In the navigation where it lists all the checks, it's really hard to find the one that failed. I think I figured out that the button for expanding has a badge with the number of actions with failures. It's extremely to hear that badge; the label doesn't say it is the numbering of failing actions. Even harder is reading the log. It seems to be in a scrolling div of some kind and the scrolling doesn't work with a screen reader. Since the output is TAP, I used the search box to search for "not OK". That helped get the right content on the screen, but it was difficult to read.

All I figured out from the log is there is some kind of webdriver error.

nschonni:
@mcking65 yeah, it looks like an unrelated regression test timeout issue. I think you might have an option to "Re-run check" for any failing tests, but I think this is OK to land either way.

There is a separate issue for seeing if we can add those inline failures like the JS/CSS/Spelling checks over here https://github.com/w3c/aria-practices/issues/1521

nschonni:
Horray! All green again

spectranaut:
@mcking65 it was an webdriver error... I think the answer really is to rerun the checks when that happens, but it's frustrating that it is so hard for you to get through to the content of the error from the github checks. I just wrote back to the internal Github Support tracker linking to your comment, for what it's worth!

PR #1470, opened 2020-07-27
utils.js: Recognize file inputs as focusable elements

mcking65:
Made from #1418  by @MartijnCuppens 

Could not merge #1418 directly because it is from a patch. Replacing with this PR.


mcking65:
Thank you @MartijnCuppens for the contribution!

PR #1479, opened 2020-07-30
Date Picker Calendar Grids: Fix color contrast for selected date

### LINKED ISSUE #1478: The calendar selected date (when it doesn't have focus) is indicated by light blue which fails 1.4.11
image

There is a graphical representation of the selected date. It is represented by light blue and it doesn't have a 3:1 contrast to indicate that it is selected to who have low vision or color disabilities.
https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

### SOLUTION
Consider turning the selected date to inverse (dark background with light text).

https://w3c.github.io/aria-practices/examples/dialog-modal/datepicker-dialog.html

carmacleod:
Closes #1478 

This just implements the simplest possible fix for issue #1478 
@jongund, please let me know if you like it, or if you would prefer to fix the issue some other way.

#### Preview Link

[View example in compare branch using RawGitHack](https://raw.githack.)


carmacleod:
Thanks, @jon! I'll do the same thing for DatePicker Combobox, then.

carmacleod:
Ok, I fixed up combobox datepicker selected date color as well.

Note that for some reason, I also had to delete the following line (in addition to changing the colors for [tabindex="0"] ):
```css
.combobox-datepicker .dates td:focus,
```

I don't know why I didn't have to delete the :focus line in the datepicker dialog.

Can you make sure I didn't mess something up? I've tested both, and they seem ok to me, but maybe I missed something.

jongund:
@carmacleod 

I think we still need the `:focus` selector, but I will test.

jongund:
@carmacleod 

Couple things I found in testing:
* I think the background color makes the date with tabindex=0 stand out alot like it has focus.  I think the background should be a little more subtle.
* In high contrast mode on Windows we do not have a visual indicator of the current date with tabindex=0 when focus is on another control in the dialog, this is a bug in the current picker dialog, too.   Is this something you could work on?

mcking65:
@carmacleod , based on Jon's last comment, it seems there's a little more to do here?

mcking65:
Per discussion in today's meeting, this is ready to merge.

Issue #1482, opened 2020-08-01
Missing caption in table example


JAWS-test:
https://w3c.github.io/aria-practices/examples/table/table.html:

As first child element the table has a div with the table description. I suspect that tables must not have any content outside the table logic (row > cell, rowgroup > row > cell, caption). At least this is the case in HTML and probably the same is true for ARIA. So the div element should be outside the table or marked with role=caption

PR #1484, opened 2020-08-02
added caption role to table example

jongund:
Added caption role to the table example for issue #1482 and also updated list of roles, properties and states and the table regression test to include caption.


carmacleod:
Hi @jongund.

Unfortunately the build is failing because the Nu checker doesn't recognize the caption role.
Not sure what we usually do about this - do we just ignore it?
Or do we open a ["caption is valid in ARIA 1.2"](https://www.w3.org/TR/wai-aria-1.2/#caption) issue against the [Nu checker](https://github.com/validator/validator/issues) and wait for them to implement so the build can pass?

The build also seems to have a problem with the deleted `'use strict';`.
The failure is "Use the global form of 'use strict'  strict".
I don't know what that means - I hope you do?

jongund:
@carmacleod 
This may have been why the table example did not include the `caption` role when it was created.

carmacleod:
@jongund @jnurthen Do we need to wait until ARIA 1.2 is Rec before we open Nu checker issues for the new roles? i.e. `caption` is in for sure, right? I can open a Nu checker issue if we're absolutely certain it won't get pulled out?

examples/table/table.html
```diff
@@ -42,7 +42,7 @@ <h2 id="ex_label">Example</h2>
     <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
     <div id="ex1">
       <div role="table" aria-label="Students" aria-describedby="students_table_desc">
```
mcking65, line 44: According to our [table naming guidance](https://w3c.github.io/aria-practices/#naming_with_captions), the aria-describedby is superfluous on table elements if a caption is present. I assume this would be true if the the element is an ARIA caption as well.

examples/table/table.html
```diff
@@ -121,6 +121,12 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
           <td><code>div</code></td>
           <td>Refers to the element that serves as the caption for the table.</td>
```
mcking65, line 122: This table row would need to be removed if aria-describedby is not needed.

examples/table/table.html
```diff
@@ -121,6 +121,12 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
...
         </tr>
+        <tr data-test-id="caption-role">
+          <th scope="row"><code>caption</code></th>
+          <td></td>
+          <td><code>div</code></td>
+          <td>The element containing a description of the table contents.</td>
```
mcking65, line 128: Suggest wording consistent with other roles

```suggestion
          <td>Identifies the element as a caption for the table.</td>
```

carmacleod:
@mcking65 
> According to our [table naming guidance](https://w3c.github.io/aria-practices/#naming_with_captions), the aria-describedby is superfluous on table elements if a caption is present. I assume this would be true if the element is an ARIA caption as well.

Actually, implicit labelling/describing with ARIA caption role has not been specified.

The [ARIA spec for role="caption"](https://w3c.github.io/aria/#caption) says authors SHOULD use `aria-labelledby` to point to the ARIA caption, or they MAY use `aria-describedby` if there's already an `aria-label[ledby]`.

The table naming guidance that you linked to is specifically about an HTML table element with an HTML caption element child. HTML-AAM specifies this relationship in its [mapping for HTML caption element](https://w3c.github.io/html-aam/#el-caption) and the section on [Accessible Name and Description Computation for HTML table element](https://w3c.github.io/html-aam/#table-element).

[CORE-AAM maps the new caption role to platform roles](https://w3c.github.io/core-aam/#details-id-9), but it defers to ACCNAME for naming and describing.

So I think that if we were to specify implicit labelling/describing for the new ARIA caption role, it would need to go somewhere in ACCNAME. Maybe it could somehow be crammed into [step 2D](https://w3c.github.io/accname/#step2D), or perhaps added as a new step after step 2D (step 2Dii anyone?), with words something like:
> Otherwise, if the current node's role (e.g. table) provides a role (e.g. caption) that defines a text alternative, return that alternative in the form of a flat string.

We might also need to reword other parts of ACCNAME a bit here and there, for example [name from author](https://w3c.github.io/accname/#mapping_additional_nd) currently says:
> name is generated from values provided by the author in explicit markup features such as the aria-label and aria-labelledby attribute, or a host language labeling mechanism, such as the alt or title attribute in HTML, or the desc element in SVG.

The [ARIA spec for table](https://w3c.github.io/aria/#table) would also need to mention the caption role somewhere (e.g. might need a new "Allowed Owned Elements:" characteristic that lists "caption" role). Other roles that can be named by a caption role, i.e. figure, grid, treegrid, would also need to mention caption - they currently don't.

Seems to me that this behavior is on the borderline between what user agents are willing and not willing to do. On the one hand, it only affects assistive technologies because it just provides another way to link up an accessible name or description. On the other hand, it further complicates the already complex ACCNAME algorithm.

I have a vague memory of the group discussing this in several ARIA WG calls, but I don't recall the conclusion. Might be all rolled into the "Name from legend" discussion, which I believe we still need to finalize before ARIA 1.3?

jongund:
The value of the `caption` role is dubious at best, since most developers will think it provides some type of table naming or describing function without the need of for using `aria-labelledby` or `aria-describedby`.   Just another reason we need authoring practices to help authors understand the differences (e.g. gotcha's) between named elements in HTML and ARIA roles with the same name in ARIA.

Issue #1403, opened 2020-05-14
Add `queryElement` (singular) API in addition to `queryElements`


zcorpan:
In #1343 we added `queryElements` as a wrapper for WebDriver `findElements`, which would fail the test if no element was found.

In #1017 we tried to adopt this in the test for that PR, but there were multiple cases where it used `findElement` to get the first element. Using `queryElements` makes the test slightly slower (in theory) and the code more convoluted since it needs to get the first element of the resulting array.

```js
  await (await t.context.queryElements(t, ex.buttonSelector))[0].sendKeys(Key.ENTER);
```

vs (proposed)

```js
  await (await t.context.queryElement(t, ex.buttonSelector)).sendKeys(Key.ENTER);
```

cc @smhigley @jongund @nschonni

smhigley:
Is this primarily for author convenience, so both functions have the same naming convention? I think we could just alias `queryElement` to `findElement`.

PR #1486, opened 2020-08-06
Add queryElement test helper function

spectranaut:
This PR fulfills the request from issue https://github.com/w3c/aria-practices/issues/1403 to have a `queryElement` helper function that is analogous to `queryElements`. It also replaces all the calls of `queryElements` when only the first element is needed.


jongund:
@spectranaut 
How will this change effect current regression tests and the regression tests in pull requests?
Will there be linting to make using `findElement` failing?

spectranaut:
@smhigley yup! just for parity.

@jongund thanks for asking but I didn't add any new lint errors, and you can still use `findElement`. The goal was just to provide an API like `queryElements` called `queryElement`, because authors might expect it.

mcking65:
OK, thank you @spectranaut  and @smhigley, this seems ready to go. @jongund , it appears that it won't cause any disruption of work for you. I will merge.

Issue #1415, opened 2020-05-28
test/tests/dialog-modal_dialog.js is failing locally but passes in TravisCI


zcorpan:
Test: `test/tests/dialog-modal_dialog.js`
Example tested by this test: `examples/dialog-modal/dialog.html`

3 tests are failing locally for me in `openDialog3` in master branch. They seem to not be failing in Travis CI.

<details>
<summary>error log</summary>

```
3 tests failed

  dialog-modal_dialog › dialog-modal/dialog.html [data-test-id="key-tab"]: tab changes focus within dialog

  test/tests/dialog-modal_dialog.js:85

   84:   const dialog = (await t.context.queryElements(t, '#dialog3'))[0];
   85:   assert(                                                          
   86:     await dialog.isDisplayed(),                                    

  Rejected promise returned by test. Reason:

  AssertionError {
    actual: false,
    code: 'ERR_ASSERTION',
    expected: true,
    generatedMessage: false,
    operator: '==',
    message: 'dialog3 should have successfullly openned',
  }

  openDialog3 (test/tests/dialog-modal_dialog.js:85:3)
  test/tests/dialog-modal_dialog.js:249:3



  dialog-modal_dialog › dialog-modal/dialog.html [data-test-id="key-shift-tab"]: shift tab changes focus within dialog

  test/tests/dialog-modal_dialog.js:85

   84:   const dialog = (await t.context.queryElements(t, '#dialog3'))[0];
   85:   assert(                                                          
   86:     await dialog.isDisplayed(),                                    

  Rejected promise returned by test. Reason:

  AssertionError {
    actual: false,
    code: 'ERR_ASSERTION',
    expected: true,
    generatedMessage: false,
    operator: '==',
    message: 'dialog3 should have successfullly openned',
  }

  openDialog3 (test/tests/dialog-modal_dialog.js:85:3)
  test/tests/dialog-modal_dialog.js:358:3



  dialog-modal_dialog › dialog-modal/dialog.html [data-test-id="key-escape"]: escape closes dialog

  test/tests/dialog-modal_dialog.js:85

   84:   const dialog = (await t.context.queryElements(t, '#dialog3'))[0];
   85:   assert(                                                          
   86:     await dialog.isDisplayed(),                                    

  Rejected promise returned by test. Reason:

  AssertionError {
    actual: false,
    code: 'ERR_ASSERTION',
    expected: true,
    generatedMessage: false,
    operator: '==',
[...]

PR #1488, opened 2020-08-07
Temporarily comment out flaky tests

spectranaut:
This is a bit unsatisfying of a PR after several hours of investigation, but there are a few flaky tests that I don't want to trip people up until they are fixed.

This PR comments out the tests that I saw failing most frequently in the CI. I commented the tests out so that we would see them reported as "missing" in the regression coverage report. This issue is being used to keep track of the tests: https://github.com/w3c/aria-practices/issues/1415


jongund:
@spectranaut 
Valerie thank you for checking into this and verifying my experience.  Hopefully can get to the bottom of the issue.

Issue #1446, opened 2020-07-13
Date Pickers: Document why space and enter behave differently in accessibility features section


mcking65:
In the [07 Jul 2020 meeting](https://www.w3.org/2020/07/07-aria-apg-minutes.html), we discussed the Enter and Space behavior in the combobox date picker dialog.

* Enter selects a date and closes the combobox popup dialog.
* Space selects the date without closing the combobox popup dialog. The user can tab to the OK button to close the dialog.

It was mentioned that it might be helpful to readers if this differentiation were explained in the accessibility features section.

The rationale for the current implementation is that by having Enter and Space differentiated, both the single date picker and the date range picker could have the same behaviors for Enter and Space. 
It is also worth noting that this is not a necessary distinction, and is one that may be revisited based on community feedback.

dotjay:
Related to this, note that in the [current datepicker example](https://w3c.github.io/aria-practices/examples/dialog-modal/datepicker-dialog.html), pressing Space does not cause the `aria-selected` attribute to update and therefore style the selected date accordingly. Space only appears to update the value of the field.

jongund:
@dotjay 

Thank you for reporting this bug, [pull request 1494](https://github.com/w3c/aria-practices/pull/1494) to fix issue, could you review it to see if it addresses the issue you raised?
The pull request has a preview link to test the fix.

dotjay:
@jongund Only nearly 4 years later, but I tested this and found that it now works as expected, but never replied to indicate this.

I notice that the documentation does not yet differentiate between the Space and Enter behaviours. Would a pull request that separates the Enter and Space behaviours under the heading [Date Picker Dialog: Date Grid](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/#kbd_label_5) be useful?

PR #1494, opened 2020-08-12
Date Picker Dialog: fix aria-select  bug when using space key

jongund:
Fixed bug reported in issue #1446 related to updating `aria-selected` attribute when the space key is pressed, since pressing space does not close the dialog.

[Preview Link](https://raw.githack.com/w3c/aria-practices/datepicker-dialog-space-select/examples/dialog-modal/datepicker-dialog.html)


mcking65:
I removed link to #1446 because this issue doesn't add the documentation that issue is about. We need that to stay open after this PR closes. Instead, I will label this a bug and put it in the datepicker project so we can track it without a related issue.

charmarkk:
@mcking65, I think I'm following. I don't usually do things this in depth!

Win/Firefox: I do see in the accessibility tree that the name is updating to the currently selected date, but NVDA doesn't announce anything to me, and will say "not selected" even on the selected date.

Win/Chrome: I'm think I'm seeing the same as you, Matt.

I must confess, I don't know Macs well enough to that inspection. 😅 Maybe I could tap in @curtbellew or @carmacleod for that?

carmacleod:
@mcking65 I agree with your assessment.

Win/Chrome: NVDA announces "selected" right away. JAWS doesn't announce it at all, but if you arrow away from the selected cell and then back again, JAWS will say "selected".

Win/Firefox: Confirmed that "selected" never shows up in the accessibility states for the cell. Definitely seems like a Firefox bug to me. In addition, Firefox has accessibility errors and warnings showing up in the accessibility tree viewer, bogusly complaining that "Clickable elements must be focusable and should have interactive semantics" for every date cell, even though they all have the focusable state because they all have a tabindex. So after this is merged, I'll open a couple of Firefox issues.

Mac/Chrome: Confirmed that selected is never reported by VoiceOver. I can see in the [MacOSX Accessibility Inspector that comes with XCode](https://developer.apple.com/library/archive/documentation/Accessibility/Conceptual/AccessibilityMacOSX/OSXAXTestingApps.html) that Selected is in fact true, so my best guess is that VO is not looking for the selected state on a cell.

Note that I can't for the life of me figure out how to browse the contents of the Accessibility Inspector with a keyboard. Is it possible that this Inspector doesn't support keyboard?? I can use Control-Option-Command-F5 to move the mouse cursor to the focused item (for example, the date cell), and then Option-Space to freeze/unfreeze the display of the item under the mouse in the Inspector (don't use Command-F7 - the doc clearly needs updating). However that's as far as the keyboard will take me.

Mac/Safari: Not only is the selected cell announced as a group, but there also seems to be a caching problem, where even if you move the selection to another cell, the old selected cell is still announced as group. This is very likely a webkit bug. I'll open a bug for webkit after we merge.

carmacleod:
Looking at Core-AAM, it seems that Mac doesn't have a way to distinguish between [cell](https://www.w3.org/TR/core-aam/#role-map-cell) and [gridcell](https://www.w3.org/TR/core-aam/#role-map-gridcell)?

Both roles map to:

AXRole: AXCell
AXSubrole: `<nil>`
AXRoleDescription: 'cell'

Also, their parent roles [table](https://www.w3.org/TR/core-aam/#role-map-table) and [grid](https://www.w3.org/TR/core-aam/#role-map-grid) both map to:

AXRole: AXTable
AXSubrole: `<nil>`
AXRoleDescription: 'table'

So I'm not sure how OSX or VoiceOver is supposed to tell the difference?
It does make a difference in this case, because `cell` role supports `aria-selected`, but `gridcell` does not.

carmacleod:
I found a few keyboard commands in the Accessibility Inspector menubar's Inspect menu. You can use Control-Command-right/left arrow to inspect the next/previous sibling element, or Control-Command-up/down arrow to inspect a parent/child element. Those used to be buttons in the Inspector's UI. I wondered where they went! However, VoiceOver still won't read the content that the Inspector is displaying.

mcking65:
I raised #1507 to track progress on raising related browser and ARIA issues, and took the liberty of assigning @carmacleod  to it. I will go ahead and merge.

PR #1497, opened 2020-08-19
chore: Allow spellchecking to fail the build

nschonni:
This landed awhile ago with it being allowed to fail. I think it's been around awhile and the instances where is started to fail PRs were valid rather than false positives, so I think it's OK to enable this now to fail PRs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,6 @@ jobs:
     - stage: Lint
       name: Spellcheck
       script: npm run lint:spelling
-      env: ALLOW_FAILURE=true
     - stage: Lint
       name: Regression Tests Coverage Report
       script: node test/util/report.js


mcking65:
Not sure we want this ... will add one more complexity to merging, and we already struggle with the weight of current complexities. If spellcheck is informative, then we won't have multiple PRs modifying the exceptions. We can instead occasionally review/modify exceptions and make changes in batches.

One thing we should do as part of editorial review is look at the result of spellcheck test and see if any of the failures are due to mistakes in the PR being reviewed.  If the PR contains content that requires a spelling exception, I'd rather not be forced to make that change to speellcheck in that PR.

nschonni:
I'm going to close this one in favour of https://github.com/w3c/aria-practices/pull/1520

PR #1499, opened 2020-08-19
Navigation links for site navigation using menu buttons

jongund:
Initial implementation of a set of navigational links using menu-buttons.

[Preview Link](https://raw.githack.com/w3c/aria-practices/menu-button-nav/examples/menu-button/menu-button-navigation.html)

### DIFF
--- a/test/tests/menu-button_navigation.js
+++ b/test/tests/menu-button_navigation.js
@@ -1,0 +1,394 @@
+
+
+const { ariaTest } = require('..');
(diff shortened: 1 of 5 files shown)


test/tests/menu-button_navigation.js
```diff
@@ -0,0 +1,394 @@
+
```
spectranaut, line 1: There should be a `'use strict;'` line here!

jongund:
no interest in including as an example right now

PR #1501, opened 2020-08-22
chore: Switch ESLint to use GitHub Actions

nschonni:


mcking65:
@nschonni commented:

> so you can see how the inline comments on the files tab works https://github.com/w3c/aria-practices/pull/1501/files#file-examples-accordion-js-accordion-js-L19

Totally love this! I wish they would put a heading on each "Check Failure" though. If I can remember the string, I can search for it in the meantime. But, a heading sure would be nice.

I wish this kind of integration were possible with AVA so we could have something like that for regression test failures.

I think we should go forward with this. Thoughts @spectranaut?

nschonni:
> Totally love this! I wish they would put a heading on each "Check Failure" though. If I can remember the string, I can search for it in the meantime. But, a heading sure would be nice.

I'm not sure where the best place to provide feedback is for that feature. The "Unchanged files with check annotations" is still beta, but I think the regular inline annotations are general release now.

> I wish this kind of integration were possible with AVA so we could have something like that for regression test failures.

It's probably not too bad to add, as it the feature is really built on a JSON file with a regex for matching like https://github.com/xt0rted/stylelint-problem-matcher/blob/master/.github/problem-matcher.json. Any of those matchers can be registered manually, but the `setup-node` action just has the ESLint and TypeScript Compiler ones built-in.

> I think we should go forward with this. Thoughts @spectranaut?

Just let me know when you want to rebase out the the last "wip" commit that is showing some failures on purpose, and I'll turn off the draft flag.

.github/workflows/lint-js.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: ESLint
+        run: npm run lint:es
```
spectranaut, line 34: wow this file is so easy to understand.

spectranaut (approved):
This is great @nschonni ! And what a cool idea to do this for the AVA tests. I'm going to make a backlog issue for that.

nschonni:
OK, I've removed the intentionally failing ESLint config, and swapped this from the draft state. I also added a comment in the config on why `setup-node` is being used, even if we're not specifying a version.

nschonni:
@mcking65 rebased the ESLint PR since there were conflicts with the `.travis.yml` file

nschonni:
@mcking65 I think this one makes sense to land next of the batch of CI changes

PR #1502, opened 2020-08-22
chore: Remove extra files in gh-pages

nschonni:


mcking65:
correct ... just protected to restrict access.

PR #1515, opened 2020-09-07
chore: Switch regression test to GitHub Actions

nschonni:
I played around with keeping the parallel job splitting by using letter ranges in a separate branch, but found that this was the most straightforward.
This uses @spectranaut script for limiting the runs to changed files, and adds the GitHub Action path filtering so the job is also only triggered if those files are touched.

@spectranaut should the full suite run on merges to the main branch?

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -1,0 +1,56 @@
+name: Run regression tests
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/regression.yml"
+      - "package*.json"
+      - "test/**"
+      - "examples/**"
+      - "!examples/landmarks/**"
+  push:
+    branches-ignore:
+      - master
+    paths:
+      - ".github/workflows/regression.yml"
+      - "package*.json"
+      - "test/**"
+      - "examples/**"
+      - "!examples/landmarks/**"
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        CI_NODE_INDEX: [0, 1, 2, 3, 4]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run Ava on changed files in PR
+        run: ./scripts/regression-tests.sh
+        if: github.event_name == 'pull_request'
+        env:
+          COMMIT_RANGE: "origin/${{ github.base_ref }}...${{ github.sha }}"
+          CI_NODE_TOTAL: 5
+          CI_NODE_INDEX: ${{ matrix.CI_NODE_INDEX }}
+          TEST_WAIT_TIME: 10000
+
+      - name: Run Ava for changed files from origin branch
+        run: ./scripts/regression-tests.sh
+        if: github.event_name == 'push'
+        env:
+          COMMIT_RANGE: "origin/master...${{ github.ref }}"
+          CI_NODE_TOTAL: 5
+          CI_NODE_INDEX: ${{ matrix.CI_NODE_INDEX }}
+          TEST_WAIT_TIME: 10000
--- a/scripts/regression-tests.sh
+++ b/scripts/regression-tests.sh
@@ -4,18 +4,6 @@ if [[ "$CI" != "true" ]]
 then
   # When running this script locally, compare the current branch to master
   COMMIT_RANGE="..master"
-
-elif [[ "$TRAVIS_PULL_REQUEST" != "false" ]]
-then
-  # If we are on a PR build, we can use TRAVIS_COMMIT_RANGE
-  COMMIT_RANGE=$TRAVIS_COMMIT_RANGE
-
-else
-  # If we are on a branch build, and it has been force pushed, then TRAVIS_PULL_REQUEST will
-  # not contain useful information for the branch build.
-  COMMIT_RANGE="origin/master...$TRAVIS_BRANCH"
-  git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
-  git fetch origin master
 fi
 
 AVACMD="npm run regression -- -t"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ git:
 
 stages:
 - Lint
-- Test
 
 jobs:
   fast_finish: true
@@ -30,31 +29,3 @@ jobs:
       name: Regression Tests Coverage Report
       script: node test/util/report.js
       env: ALLOW_FAILURE=true
-
-    - stage: Test
-      name: AVA Regression Tests
-      addons:
-        firefox: latest
-      script: scripts/regression-tests.sh
-      env:
-        - TEST_WAIT_TIME=1000
-        - CI_NODE_TOTAL=3
-        - CI_NODE_INDEX=0
-    - stage: Test
-      name: AVA Regression Tests
-      addons:
-        firefox: latest
-      script: scripts/regression-tests.sh
-      env:
-        - TEST_WAIT_TIME=1000
-        - CI_NODE_TOTAL=3
-        - CI_NODE_INDEX=1
-    - stage: Test
-      name: AVA Regression Tests
-      addons:
-        firefox: latest
-      script: scripts/regression-tests.sh
-      env:
-        - TEST_WAIT_TIME=1000
-        - CI_NODE_TOTAL=3
-        - CI_NODE_INDEX=2


mcking65:
@nschonni asked:

> @spectranaut should the full suite run on merges to the main branch?

No, we are limiting for performance reasons. It takes too long to run the full suite for every merge to the main branch.

This is probably unrelated, but I wonder if we need to test both the branch and the PR when merging from a branch to main. Isn't it sufficient to test the PR?

nschonni:
I can turn off the branch testing for pushes to the default branch. I had left the `push` part on, because I thought it might still be useful for people working on their forks, or for testing on feature branches.

From https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags it would mean adding the following to the existing `push` sectiong
```yaml
    branches-ignore:
      - master
```

nschonni:
PS: I found it only took around 9 minutes to run the full set of regressions https://github.com/nschonni/aria-practices/runs/1076320104?check_suite_focus=true when I just called the `npm run regression` in my first take on this

scripts/regression-tests.sh
```diff
@@ -4,18 +4,6 @@ if [[ "$CI" != "true" ]]
...
-
-else
-  # If we are on a branch build, and it has been force pushed, then TRAVIS_PULL_REQUEST will
-  # not contain useful information for the branch build.
-  COMMIT_RANGE="origin/master...$TRAVIS_BRANCH"
-  git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
-  git fetch origin master
 fi
 
 AVACMD="npm run regression -- -t"
```
spectranaut, line 9: When ava was running tests in parallel process in travis, the tests became pretty flaky, so we had to locked down concurrency to 1 with `-c 1` (by default ava uses the number of "cores"). I removed that command line argument when we switched to parallelizing across VMs. I'm not sure how flaky the tests will be in this new Github action environment? Have you tested triggering the tests a few times in a row?

spectranaut (commented):
cool that the tests ran so fast! Ava can parallelize in two ways -- by VM or by process. You took out the by VM parallelization we were using but we can revisit if the tests become slow in the future -- 9 mins is similar to what we had after the VM parallelization.

However, see my comment about flaky tests -- Ava is probably using process parallelization right now.

scripts/regression-tests.sh
nschonni: I didn't notice the flakey test in the other approach https://github.com/w3c/aria-practices/compare/master...nschonni:ava-github-action, although I don't think the library Ava uses for parallelization supports GitHub Actions right now https://github.com/avajs/ava#parallel-runs-in-ci so it might not be triggering.
Could add those environmental variables back though if you think it is better

spectranaut: Let's see how it goes with this in, if I see flakey tests we know a quick fix (to locked down the concurrency). It could be the throttled travis environment that lead to the flakey tests.

spectranaut (approved):

mcking65:
@spectranaut you're comfortable merging without further testing for flakiness?

nschonni:
Do you also want me to filter it out from running on merges to `master`?

spectranaut:
Actually @nschonni can you trigger a bunch of runs on your fork to test for flakiness? Will pushing an empty commit trigger the github action?

Also we might as well filter out running merges to master, as you suggested!

nschonni:
@spectranaut I've added you to my fork. You can rerun any of the jobs again by hitting the "Re-run jobs" button on the top right of any Action run https://github.com/nschonni/aria-practices/runs/1077220667?check_suite_focus=true

spectranaut:
@nschonni awesome but unfortunately all of the tests are skipped in the PR (by the bash script) do re-running it won't show anything. We need to edit the bash script for a test/index.js or test/util/* file to trigger all the test to run. I can push something (called "WIP" we can remove like your other tests) and then rerun

nschonni:
Ah, here is an older workflow that just called the `npm run regression` as well https://github.com/nschonni/aria-practices/runs/1076320104?check_suite_focus=true

spectranaut:
Hmm I just ran all tests twice and both times the same test failed ([here](https://pipelines.actions.githubusercontent.com/RNdM1g9anitrXHMO0JdvfNuq6TXr8dEd94CqDlR1kyNTXNPcgV/_apis/pipelines/1/runs/19/signedlogcontent/3?urlExpires=2020-09-08T22%3A28%3A05.1364402Z&urlSigningMethod=HMACV1&urlSignature=iaQDwMhrbohmBZTBw3TS%2FHYm6wG4u%2F%2B305hO%2F%2BvWTsY%3D) and [here](https://pipelines.actions.githubusercontent.com/RNdM1g9anitrXHMO0JdvfNuq6TXr8dEd94CqDlR1kyNTXNPcgV/_apis/pipelines/1/runs/20/signedlogcontent/3?urlExpires=2020-09-08T22%3A40%3A30.6068787Z&urlSigningMethod=HMACV1&urlSignature=bp5FkkTg1Qvvv2r4eugvZxVLYZRUC3zuhuHIliEdsmw%3D)) -- it's a test that opens a link, it could be a time out. We could try extending the time out or switching to concurrency = 1?

.github/workflows/regression.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: actions/setup-node@v1
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run Ava on changed files in PR
+        run: ./scripts/regression-tests.sh
+        if: github.event_name == 'pull_request'
+        env:
+          COMMIT_RANGE: "origin/${{ github.base_ref }}...${{ github.sha }}"
```
spectranaut, line 38: We need to add a env variable here to extend the wait time for the CI -- `TEST_WAIT_TIME=1000` was the wait time we set in travis.

spectranaut (changes requested):

.github/workflows/regression.yml
nschonni: Odd, looks like the default is supposed to be higher https://github.com/avajs/ava/blob/master/docs/07-test-timeouts.md#test-timeouts but I'll take a look

nschonni:
Arg, green here, but red over here https://github.com/nschonni/aria-practices/runs/1088609115?check_suite_focus=true

nschonni:
@spectranaut I added back your parallelization of the job (and increased it to 5, because why not :)), but I'm still getting consistent failures on `not ok 113 - treeview_treeview-2a › treeview/treeview-2/treeview-2a.html [data-test-id="key-enter-or-space"]: Key enter opens folder and activates link`

.github/workflows/regression.yml
```diff
@@ -0,0 +1,53 @@
...
+      - "test/**"
+      - "examples/**"
+      - "!examples/landmarks/**"
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        CI_NODE_INDEX: [0, 1, 2, 3, 4]
```
spectranaut, line 26: oh cool! :)

.github/workflows/regression.yml
spectranaut: Oh you know I missed this, this is a different time out than the default ava time out, it's a timeout used for a different purpose in the tests. The timeout is related to how long we might want to wait for the DOM to change. I'm going to push that variable and see if it works!

package.json
```diff
@@ -63,7 +63,8 @@
   "ava": {
     "files": [
       "test/tests/*"
-    ]
+    ],
+    "timeout": "1m"
```
spectranaut, line 67: Yeah this also doesn't effect the waiting for DOM change timeout!

nschonni:
@spectranaut I'm still getting random failures in some treeview tests, but I also hit the same thing with Travis on another PR, so I don't think this is introducing anything new.
I bumped the parallel up to 10 jobs just to see if it helped, but I can reduce it again if you want. Also turned off "fast-fail" so all the jobs will finish even if one fails, and that one job can just be restarted (if you have permission),

Should I rebase off the the last commit so this can land?

spectranaut:
@nschonni that all sounds good to me. I think we can fix the treeview tests (I have some ideas!) but I just haven't had a moment to look into it.

nschonni:
Thanks @spectranaut, I forgot to mention I also increased the timeout to 10000 to see if it helped, but I can put it back to 1000

spectranaut:
@nschonni did increasing the jobs to 10 cause a noticeable speedup? because I think we might as well decrease it to 5 simply because it might make looking through failures a bit annoying (if you have to scroll through/click through some many different jobs to see them all).

Also are you saying that increasing the timeout from 1000 to 10000 didn't actually help prevent the treeview failures? that's pretty crazy. I looked at your fork but didn't see where I could find them. Can you link me a failure with the 10000 or is it too hard to find..? I could alternatively run some tests in my own branch.

nschonni:
> I looked at your fork but didn't see where I could find them

Ah , sorry, I found out I could delete old Action results so I cleaned up the various names I had used for the job name that were cluttering up left side bar. I think it's more likely a "flaky" test where Ava might be missing some sort of ready even rather than a real timeout. Seems most of the failures I was seeing were around the Treeview tests, which can be finicky I guess :)

I'll change the parallelization back to 5

nschonni:
@spectranaut back to 5 in parallel now

spectranaut:
@mcking65 Ok! I think this is good to merge :)

nschonni:
Thanks @spectranaut think you might need to change your review from "Request Changes"

spectranaut (approved):
Definitely approving this PR! I tested to see if I could see the flaky treeview tests.. and I can't any more: https://github.com/bocoup/aria-practices/pull/32 (note: on that branch there are travis build failings you have to click the checks to see the regression pass every time)

The tests all rely on reaching wikipedia, so maybe there was some github checks infrastructure internet problem when we first tested this? Not sure. Anyway stoked for this change!

nschonni:
Ah, interesting. Maybe there is some rate-limiting going on. I know GitHub itself can sometimes return 429 status' when limits are hit

nschonni:
@mcking65 I've rebased this one now that the Stylelint PR landed

mcking65:
@nschonni thank you for all your work on this!!

nschonni:
@spectranaut was checking out link checkers and it looks like wikipedia is returning 429 responses on a bunch of those links. Might be work trying to find a different site for the links (if that makes sense)

spectranaut:
> @spectranaut was checking out link checkers and it looks like wikipedia is returning 429 responses on a bunch of those links. Might be work trying to find a different site for the links (if that makes sense)

Hmm I see... Maybe we should have a policy of only redirecting to internal websites for examples to make sure the examples remain easy to test. Or we could just add some kind artificial slow down when testing tests that refer to external websites. For now lets just keep an eye on it and address it if it causes failures, and address it then?

PR #1516, opened 2020-09-07
chore: Trigger regression for lock file changes

nschonni:
I thing usually when dependencies are updated it includes both package and the lock file, but I've seen some security Dependabot updates that only update the lockfile to fix security issues. This will then trigger the regression tests to make sure lockfile only changes don't regress anything.


spectranaut (approved):
thanks for this catch! :)

PR #1517, opened 2020-09-07
Infrastructure: Move Regression coverage to PR comment

nschonni:
Currently this is a job that is allowed to fail on Travis CI. This moves it to an action that will leave a comment on the PR if any of the regression related files are changed. Additional pushes will update the same comment instead of continuing to create new ones on the PR.
~~I'm set it so it also runs when the CI config file is touched so you'll see it on this PR~~, but wouldn't on a PR that doesn't touch anything in the `examples` or `test` folder.

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -1,0 +1,60 @@
+name: Regression Tests Coverage Report
+
+on:
+  pull_request_target:
+    paths:
+    - "examples/**"
+    - "test/**"
+    - "!examples/landmarks/**"
+
+jobs:
+  report:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: refs/pull/${{ github.event.pull_request.number }}/head
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run coverage report
+        run: |
+          node test/util/report.js >> coverage.log || true
+
+      - name: Comment on PR
+        uses: actions/github-script@0.9.0
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const fs = require('fs');
+            const commentBody = '```' + fs.readFileSync('coverage.log', 'utf8') + '```';
+
+            if (commentBody.length === 0) {
+              return
+            }
+            // Get the existing comments.
+            const {data: comments} = await github.issues.listComments({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.number,
+            })
+            // Find any comment already made by the bot.
+            const botComment = comments.find(comment => comment.user.id === 41898282)
+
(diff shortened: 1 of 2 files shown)


nschonni:
Ah, looks like the security restrictions on PRs isn't letting this preview from something coming from a fork. I don't think this will be an issue once it's in the main branch. You can see result over here though https://github.com/nschonni/aria-practices/pull/19#issuecomment-688025952

nschonni:
Using the `pull_request_target` gives access to a read/write token in order to actually make the comment https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks

mcking65:
@nschonni which PRs will this add a comment to? All? Or, just PRs that change directories where some of the examples have missing tests?

Something like this would be cool for spell check.

nschonni:
From this part of the config
```yaml
  pull_request_target:
    paths:
    - "examples/**"
    - "test/**"
```
This job will only trigger if the PR touches files under the `examples` or `tests` folders. For other PRs that don't touch those files, the job won't even start or leave a comment

.github/workflows/coverage-report.yml
```diff
@@ -0,0 +1,59 @@
+name: Regression Tests Coverage Report
+
+on:
+  pull_request_target:
+    paths:
+    - "examples/**"
+    - "test/**"
```
spectranaut, line 7: We should filter out examples/landmarks -- `"- !examples/landmarks/**"`

.github/workflows/coverage-report.yml
```diff
@@ -0,0 +1,59 @@
...
+            // Find any comment already made by the bot.
+            const botComment = comments.find(comment => comment.user.id === 41898282)
+
+            if (botComment) {
+              await github.issues.updateComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                comment_id: botComment.id,
+                body: commentBody
+              })
```
spectranaut, line 51: Does this replace the comment, not append? Because it we don't replace it will get really long! Seems like it from the documentation but just wanted to be sure.

spectranaut (changes requested):
One small change to filter out landmark example changes -- but this is awesome, thanks so much! :)

.github/workflows/coverage-report.yml
nschonni: It replaces it if it exists. You can see it in the "edit" dropdown on my test PR in my fork https://github.com/nschonni/aria-practices/pull/19#issuecomment-688025952

.github/workflows/coverage-report.yml
nschonni: Good call! Done

spectranaut (approved):
Super great! Can't wait for this :)

mcking65:
@nschonni  thank you again ... looking forward to how this will bring the report to the surface.

PR #1518, opened 2020-09-07
chore: Move Stylelint to GitHub action and fix broken config

nschonni:
- Swaps Travis for Stylelint step
- Adds the "problem matcher" so issues are commented directly on the lines in PRs
- Fix the styleline config which was in 2 files, and wasn't valid in one, so hasn't been running

### DIFF
--- a/.github/workflows/lint-css.yml
+++ b/.github/workflows/lint-css.yml
@@ -1,0 +1,36 @@
+name: CSS Linting
+on:
+  push:
+    paths:
+      - "package*.json"
+      - ".stylelint*"
+      - "**/*.css"
+      - ".github/workflows/lint-css.yml"
+      - "!common/**"
+
+  pull_request:
+    paths:
+      - "package*.json"
+      - ".stylelint*"
+      - "**/*.css"
+      - ".github/workflows/lint-css.yml"
+      - "!common/**"
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+
+      - uses: xt0rted/stylelint-problem-matcher@v1
+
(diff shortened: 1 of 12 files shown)


nschonni:
@mcking65 it looks like Stylelint has been broken since the config file changes in #1255 
I can submit those changes separately from the CI changeover if you want, but this was useful in also highlighting the things the problem matcher is flagging

mcking65:
> @mcking65 it looks like Stylelint has been broken since the config file changes in #1255

Wow ... interesting. Wonder how that snuck past us.

> I can submit those changes separately from the CI changeover if you want, but this was useful in also highlighting the things the problem matcher is flagging


I think a PR to fix the config and change the CI is fine. Thank you for this!

I'll need a few people to look at this. @jongund, this could impact other work in progress, but I don't think in any significant way.

nschonni:
The remaining issues that I didn't fix was for this rule https://stylelint.io/user-guide/rules/no-descending-specificity
It could either be disabled to get back to green, or it needs to be addressed in the files that are flagged

spectranaut (changes requested):
Thanks so much for this! If fixing all the no-descending-specificity is too much for you, then can you push a commit to ignore the rule, and add a comment and an issue?

nschonni:
@spectranaut good call. I opened https://github.com/w3c/aria-practices/issues/1522, but couldn't add a comment in the RC file, because then it stopped reading it again 😄 . The rule is null'd again for now

spectranaut (approved):
At least we have the issue :)

nschonni:
@spectranaut the regression tests doh't even like me here

spectranaut:
@nschonni the regression tests had a strange timeout error I haven't seen before (something like "no new tests in 10 minutes") so I just restarted the job and it passed. This error looks unrelated to the PR and isn't a pattern I've seen anywhere else.

@mcking65 I think this can be merged!

PR #1520, opened 2020-09-08
Infrastructure: Convert spellcheck to GitHub Action

nschonni:
Runs the spellchecking only if the `aria-practices.html` or HTML files in the `example` directory. Add a problem matcher so that comments are left on the lines with spelling issues in PRs and pushes.
The main change that this swap brings, is that GitHub Actions don't have the concept of "allowed failures", so these will be blocking failures.

Closes #1497

### DIFF
--- a/.github/workflows/cspell-problem-matcher.json
+++ b/.github/workflows/cspell-problem-matcher.json
@@ -1,0 +1,16 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "cspell",
+      "pattern": [
+        {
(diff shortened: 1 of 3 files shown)


mcking65:
OK, per discussion in telecon yesterday, we are taking the plunge and making spelling a required check ... we'll see how it goes!!

@nschonni thank you again for all your work to help us ensure quality before ship.

nschonni:
@mcking65 it won't actually be "required" unless you see branch policy, but I'll spin up a separate discussion on that

PR #1366, opened 2020-04-03
DO NOT MERGE: demonstrate Build Linter failure for #1365

zcorpan:
See #1365. Do not merge this PR.

The second commit here changes `examples/accordion/accordion.html` from `aria-expanded="true"` to `aria-busy="true"`, which would cause a change to the index. The expectation is that the new Build Linter check will fail in Travis CI.

PR #1524, opened 2020-09-10
Infrastructure: Add CI for examples/index.html syncing

### LINKED ISSUE #1079: Autogenerated and check in examples/index.html

nschonni:
- Adds CI to check if examples/index.html was regenerated correctly when files were changed
- Add local husky/lint-staged to generate the index as files are edited

Fixes #1079
Closes #1365
Closes #1366


nschonni:
@spectranaut alternately I could remove the file and add it to the deploy Github Action as an extra step

nschonni:
@mcking65 here is an example of the report working in the above comment

spectranaut:
@nschonni this is a cool trick (adding a check to make sure the examples/index.html file had been regenerated) but I do think it would be best to just add to the deploy. Is it ok @mcking65 to remove the file from the repo and just generate it on deploy?

mcking65:
I don't understand how the changes in this PR work. Does it check to see if the changes in a PR would make the examples index out of date and then, if yes, generate a new one to be included in the files changed by the PR?

If that is the approach, I'm concerned about conflicts in the examples index file. Say I create a PR that adds an example. This action runs against my PR and generates the index in my branch. Say, there is a conflict with the index currently in the main branch. I don't want to have to manually resolve a conflict in a generated file.

@spectranaut, I think you are suggesting making the examples index a generated file that appears only in the gh-pages branch? My concern there is that Michael Cooper does not use the gh-pages branch as the basis for publication to TR. I suppose we could manually generate the examples index in the publication branches. But, then we have only solved part of the problem; we still have to remember to manually run the script as part of the publication process. BRTW, the pub branches are apg-1.2, apg-1.3, etc.

nschonni:
> I don't understand how the changes in this PR work. Does it check to see if the changes in a PR would make the examples index out of date and then, if yes, generate a new one to be included in the files changed by the PR?

1. It uses the existing commit hook setup, so that if any of the example files are touched, it tries to re-generate that index file. Most cases, if people are just updated content in those files, it wouldn't actually cause anything to change.
1. If someone bypasses the pre-commit hook on purpose, or by making changes through the web UI, this adds a CI job to fail the build if it sees the examples index file is out of date

> Say I create a PR that adds an example. This action runs against my PR and generates the index in my branch. Say, there is a conflict with the index currently in the main branch. I don't want to have to manually resolve a conflict in a generated file.

The Action part of this today, is just a backstop to make sure they're in sync. To resolve the conflict you could re-run locally `node scripts/reference-tables.js` or `npm run node reference-tables`, since that is the script that today just fully generates that page.

> My concern there is that Michael Cooper does not use the gh-pages branch as the basis for publication to TR. I suppose we could manually generate the examples index in the publication branches. But, then we have only solved part of the problem; we still have to remember to manually run the script as part of the publication process. BRTW, the pub branches are apg-1.2, apg-1.3, etc.

If you want to switch to a generate only for specific branches, that could be handled later and parts of this disabled. It would probably involve the following:
1. Disable the pre-commit hook
1. Add the file to the gitignore
1. Delete the file from the default branch
1. Create a new Action or adjust the branch filter on this action so it only checks branches like `apg-*`
1. Add a step in the current deploy.yml file to call the script

package.json
```diff
@@ -63,6 +64,10 @@
     "*.js": [
       "eslint --fix",
       "git add"
+    ],
+    "examples/**/*.html": [
+      "npm run reference-tables",
+      "git add examples/index.html"
```
spectranaut, line 70: this hook is cool, but I was testing it and it doesn't run when you are deleting a file, which might make it a bit of an inconsistent experience. I also think it's a bit weird to have something be automatically added to your commit -- maybe we should just stick the the failure in github checks letting you know you haven't regenerated that particular file?

spectranaut (changes requested):
Yeah I agree lets not worry about autogeneration in the github actions for now. I'm torn about leaving the regeneration in the github hooks, I feel like it might be confusing for contributors when it runs, where as the github checkers failure is very clear and consistent.

spectranaut (approved):
You know what, I changed my mind. I think it's fine that the examples/index.html file is regenerated in a commit hook. I'm just going to add this to a wiki page.

nschonni:
@mcking65 did the explanations in https://github.com/w3c/aria-practices/pull/1524#issuecomment-707159502 help, or is it still fuzzy

package.json
nschonni: I'm good with dropping this if you want

spectranaut:
I was just looking at old PRs, and I found #1365 from @zcorpan which has the same ultimate functionality as this PR! It adds a local linter step to do the recreation and comparing of the index file instead of a git commit hook. The git commit hook is nice though, as it removes a step you would otherwise had to perform manually.

mcking65 (approved):
@nschonni 
After discussing with @spectranaut, I am fine with trying this approach. If we have any problems, we can revisit.

Ideally, I wish there was some kind of post-merge action that could have the permission to auto-commit to the main branch. Then, instead of running index generation before the merge and making it part of the merge, the action would generate the index.html in master after merge and then auto-commit the result to master if the file has changed. Then, we'd be guaranteed to never see merge conflicts in the generated index.html file.

PR #1528, opened 2020-09-21
Listbox Example with Grouped Options: Apply aria-hidden to option group label

WilcoFiers:
This ensures that the label of a group in a listbox is not counted / identified by AT as also being a member of the group. Putting role="presentation" on the li just means the li will be ignored by AT, but not its children. The text node instead ends up as a member of the group.

#### Preview link

[View proposed changes using RawGitHack](https://raw.githack.com/WilcoFiers/aria-practices/patch-1/examples/listbox/listbox-grouped.html)

### DIFF
--- a/examples/listbox/listbox-grouped.html
+++ b/examples/listbox/listbox-grouped.html
@@ -46,7 +46,7 @@
           <span id="ss_elem" class="listbox-label">Choose your animal sidekick</span>
           <div id="ss_elem_list" tabindex="0" role="listbox" aria-labelledby="ss_elem">
             <ul role="group" aria-labelledby="cat1">
-              <li role="presentation" id="cat1">Land</li>
+              <li aria-hidden="true" id="cat1">Land</li>
               <li id="ss_elem_1" role="option">Cat</li>
               <li id="ss_elem_2" role="option">Dog</li>
               <li id="ss_elem_3" role="option">Tiger</li>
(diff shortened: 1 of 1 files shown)


examples/listbox/listbox-grouped.html
```diff
@@ -46,21 +46,21 @@ <h2 id="ex_label">Example</h2>
           <span id="ss_elem" class="listbox-label">Choose your animal sidekick</span>
           <div id="ss_elem_list" tabindex="0" role="listbox" aria-labelledby="ss_elem">
             <ul role="group" aria-labelledby="cat1">
-              <li role="presentation" id="cat1">Land</li>
+              <li aria-hidden="true" id="cat1">Land</li>
```
mcking65, line 49: Using aria-hidden to manage screen reader verbosity related to elements referenced by aria-labelledby has been contraversial. The task force has not supported doing so. It is a technique that can have severe negative consequences. In this particular case, it might be innocuous, but is addressing a symptom of a larger problem rather than the root cause, doing so  with a risky approach.

mcking65 (commented):
I think the root problem is that screen readers, such as VO in Chrome, render the label text in a way that does not distinguish it from the selectable options. But, this raises the larger discussion surrounding screen reader rendering of labeling content.

I will add this to the October 13 agenda. I'm not sure what the next step is. Currently, this PR is not consistent with APG labeling guidance.

mcking65:
Added to draft of [October 13, 2020 Agenda · w3c/aria-practices Wiki](https://github.com/w3c/aria-practices/wiki/October-13%2C-2020-Agenda).

mcking65:
Closing per group discussion on October 13.

The approach in this PR conflicts with APG labeling guidance, which  recommends against hiding visible labeling text. Another approach to resolving the set numbering would be to explicitly declare setsize and posinset. The group also notes that some screen readers, including macOS VoiceOver, do not distinguish between a selectable option and static labeling text. This is something that should be addressed by aria-at tests.

@WilcoFiers 
If you disagree or would like to resolve the issue with another approach, please open an issue or propose a different PR.

PR #1533, opened 2020-09-25
Codepen button should show only when resources loaded

spectranaut:
It was a good point @smhigley to show the button after the resources have been loaded instead of hiding it if they don't! :)

I didn't implement the promise array because [XMLHttpRequest.send](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send) doesn't return a promise but that it's true that would have been much cleaner.


examples/js/examples.js
```diff
@@ -378,32 +380,28 @@ function addOpenInCodePenForm (exampleIndex, exampleHeaderId, exampleCodeId, exa
               totalFetchedFiles++;
           }
           else {
-              hideButton(buttonId, "Could not load resource: " + href);
+              console.log("Not showing 'Open in Codepen' button. Could not load resource: " + href);
```
nschonni, line 383: Does it make sense to warn/error for these?
```suggestion
              console.warn("Not showing 'Open in Codepen' button. Could not load resource: " + href);
```

spectranaut: makes sense! pushed commit for switch.

mcking65:
@smhigley friendly ping when you get time

smhigley (approved):
love it! :)

carmacleod (approved):
+1

Looks good!

mcking65:
Thank you @smhigley  and @carmacleod  ... will merge!

PR #1550, opened 2020-09-30
chore: Stop linting Deleted Section Archive

nschonni:
Removed the HTML validation suppression only related to that file


spectranaut (approved):
looks good!

PR #1523, opened 2020-09-09
Infrastructure: Switch HTML linting to GitHub Actions

nschonni:
Moves Travis-CI job to GitHub Actions, and adds a problem matcher so errors are identified inline on the files tab of PRs

PR #1551, opened 2020-10-01
Infrastructure: Disable push builds for Dependabot

nschonni:
They get run as part of the PRs submited and avoids a double build

### DIFF
--- a/.github/workflows/lint-js.yml
+++ b/.github/workflows/lint-js.yml
@@ -1,6 +1,8 @@
 name: JavaScript Linting
 on:
   push:
+    branches-ignore:
+      - "dependabot/**"
     paths:
       - "package*.json"
       - ".eslint*"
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-js.yml
```diff
@@ -1,6 +1,8 @@
 name: JavaScript Linting
 on:
   push:
+    branches-ignore:
+      - "dependabot/**"
```
spectranaut, line 5: Oh now I see why you are skipping this in PR #1523, approving there and here!

spectranaut (approved):

.github/workflows/lint-js.yml
nschonni: Yup, just avoids those times when an unstable test would pass on the PR and not the branch or vice-versa :)

PR #1553, opened 2020-10-01
Infrastructure: Add Prettier ESLint config

nschonni:
Related to https://github.com/w3c/aria-practices/issues/1180
Kept the the defaults, except for the quoting, that seems primarily single quotes


.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "singleQuote": true
```
nschonni, line 2: This seemed to be the more common pattern so I used this over the default double quote

spectranaut (approved):
This looks great too!

After this merges @jongund, `npm run lint:es` will fail on a lot more things related to coding style -- but you can fix all of these failures by running `npm run fix`. It will break long lines up into multiple lines and switch double quotes with single quotes and other things like this. You will probably have to run `npm run fix` on all your open PRs once this merges.

jongund (approved):

smhigley (approved):

PR #1554, opened 2020-10-01
Infrastructure: Add Prettier Stylelint config

nschonni:
Related to #1180
Was built off the same branch, so there is overlap with the config, and then suppressing the option in the Stylelint config

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "stylelint-config-standard",
+  "extends": ["stylelint-config-standard", "stylelint-prettier/recommended"],
   "rules": {
     "media-feature-name-no-unknown": [
       true,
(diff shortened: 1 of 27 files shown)


.stylelintrc
```diff
@@ -9,6 +9,7 @@
         ]
       }
     ],
-    "no-descending-specificity": null
+    "no-descending-specificity": null,
+    "prettier/prettier": [true, {"singleQuote": false}]
```
nschonni, line 13: Thus is undoing the what is set in the config, because I had the config from a larger branch looking at Prettier across the different file types

spectranaut (approved):
looks good to me! :)

jongund (approved):

smhigley (approved):
looks great to me!

PR #1558, opened 2020-10-07
Navigation Treeview Example: Convert to single page example with functionality equivalent to disclosure navigation example

### LINKED ISSUE #1526: Revise Navigation Tree Example to match Disclosure Navigation Example
As part of plan to resolve issue #89, revise the [Navigation Treeview Example](https://w3c.github.io/aria-practices/examples/treeview/treeview-2/treeview-2a.html) to represent mythical university, emulating behavior of [Example of Disclosure for Navigation Menus](https://w3c.github.io/aria-practices/examples/disclosure/disclosure-navigation.html).

jongund:
Resolves #1526  by revising the navigation tree view example.

Changes:
* Deleted previous treeview for links examples
* Created a new file treeview-navigation.html

To dos:
* Change images for collapsing and expanding treeitems to use SVG
* Regression tests still need to be completed.

#### Preview link

[View revised example in compare branch](https://raw.githack.com/w3c/aria-practices/update-tree-nav/examples/treeview/treeview-navigation.html)

### Review checklist

* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @a11ydoer 
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @spectranaut 

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1558.html" title="Last updated on Jan 19, 2021, 3:32 PM UTC (1f9ca3b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1558/cd797df...1f9ca3b.html" title="Last updated on Jan 19, 2021, 3:32 PM UTC (1f9ca3b)">Diff</a>


jongund:
@spectranaut 
What does Travis CI failing mean?
Is this something I should worry about?

nschonni:
@jongund it's flagging some HTML validator issues
```
"file:/home/travis/build/w3c/aria-practices/examples/treeview/treeview-navigation.html":223.13-223.85: error: The “main” element must not appear as a descendant of the “section” element.

"file:/home/travis/build/w3c/aria-practices/examples/treeview/treeview-navigation.html":223.13-223.85: error: The “main” element must not appear as a descendant of the “main” element.

"file:/home/travis/build/w3c/aria-practices/examples/treeview/treeview-navigation.html":223.13-223.85: error: A document must not include more than one visible “main” element.

"file:/home/travis/build/w3c/aria-practices/examples/treeview/treeview-navigation.html":508.62-508.69: error: Stray end tag “group”.

"file:/home/travis/build/w3c/aria-practices/examples/treeview/treeview-navigation.html":508.136-508.140: error: End tag “td” seen, but there were open elements.

"file:/home/travis/build/w3c/aria-practices/examples/treeview/treeview-navigation.html":508.51-508.56: error: Unclosed element “code”.
```

examples/treeview/treeview-navigation.html
```diff
@@ -531,15 +492,21 @@ <h2 id="rps_label">Role, Property, State, and Tabindex  Attributes</h2>
...
                   <li>Indicates the parent node is open, i.e., the descendant elements <strong>are</strong> visible.</li>
                   <li>The visual indication of the expanded state is synchronized by a CSS attribute selector.</li>
                 </ul>
               </td>
             </tr>
+            <tr data-test-id="treeitem-aria-owns">
+              <td></td>
+              <th scope="row"><code>aria-owns=&quot;IDREF&quot;</code></th>
+              <td><code>a</code></td>
+              <td>Refers to the element with role <code>group</group> that contains the next level down <code>treeitem</code> elements.</td>
```
nschonni, line 508:
```suggestion
              <td>Refers to the element with role <code>group</code> that contains the next level down <code>treeitem</code> elements.</td>
```

mcking65:
@jongund, thank you, this is wonderful progress!

Do we need that much filler content? It takes a long time to read through it. Can we have short, simple content that matches what we have in the 
[Example Disclosure for Navigation Menus](https://w3c.github.io/aria-practices/examples/disclosure/disclosure-navigation.html#mythical-page-content)

Instead of a complementary landmark region for the filler, can we use a region named "Mythical University sample page content"? That is what we have on the disclosure example page.

In the disclosure example, activating a link moves focus to the heading of the content. Shouldn't we do the same here?

mcking65:
When the page loads, the home content page is displayed, but it doesn't have aria-current set.

mcking65:
I expanded about and loaded overview. If I refresh the page, the about branch is collapsed so the user can't see which page is current.

Should the current page always receive focus when tabbing into the tree?

After expanding a parent tree item, the name seems to get duplicated. For example, JAWS reads about as "Open about about".

jongund:
@mcking65 
Made some updates to reduce filler content and set tabindex=0 on the current page.

mcking65:
@jongund 

Nice progress! I think this is getting close.

I don't understand what is happening with tabindex 0 for current page treeitem; it doesn't seem to be working. To repro:
1. Tab into tree and press enter on about to make it current page.
2. Tab back into tree and arrow up to home; do not press enter; keep about as current page.
3. tab out of the tree.
4. Tab back into tree.
   * Expected: About is focused
   * Actual: Home is focused

mcking65:
@jongund 

There is a bug with left arrow on descendant items. To repro:

1. Tab to tree.
2. Press down and right to expand about.
3. press down to focus overview.
4. Press left.
   * Expected: focus moves to about
   * Actual: no change

I'm surprised the regression tests didn't catch this. They should. So, that needs to be investigated as well.

mcking65:
When you press enter on a tree item, I wonder if we should focus the H1 instead of the region. That seems like it would be a better experience. Just a thought; curious what others think.

It also seems like the site name and tag line should be in a banner region.

This is nice and clean now; it feels great with a screen reader.

jongund:
@mcking65 

Changing the behavior of `tabindex=0` to not follow focus when navigating the tree, and having `tabindex=0` stay on the link with `aria-current=page` has the following issues:
1.  What happens if the link with `aria-current=page` is on a leaf that is hidden by the user?  When the user tabs out and back to the tree view, is the tree suddenly expanded to show the hidden link, or is the user never allowed to hide the link with `aria-current=page`?
2. If this change is made it is different from the previous treeview examples, we would need to describe this some place.

jongund:
@mcking65 
I am a little concerned about moving focus to the `h1` element.  While it is useful in our little example, it is also telling the authors ,who may just copy this code or behavior, that when a navigation link moves to a new page you should set focus on the `h1` element.   I don't think this is a general rule we want to promote.  I think it would be better to set focus on the container of the Mythical Uniuversity example to make the behavior similar to what usually happens when a user follows a link, you start at the beginning of the page.   Your thoughts?

sinabahram:
Sorry, but I have never understood the advice to focus a container. This is deeply problematic for screen reader users. Often times, you can't do anything from that point, focus is locked, you have to exit into a virtual cursor, etc. Can we please not only set focus to the h1 here, but I want to actually advocate for what @jongund  is recommending against. I would love it if developers focused the heading or first element and never the pane. Also, often times there is a ton of info in the pane and all of it gets read out confusingly/annoyingly.

jongund:
@sinabahram 
@mcking65 
I can focus the `h1` in this example, but I think it requires a more general discussion to determine if this is a good idea as a general practice.  Most people expect to start tabbing from the beginning of the page.  I think it will be unexpected for keyboard only users to start from the middle of the page, and what browsers do by default.  Asking authors to now actively set keyboard focus on a non-interactive element just doesn't sound right to me, since they will probably not do it correctly, and even if it done correctly it's not what most users will expect.

jongund:
@mcking65 
I fixed the arrow left bug, the regression tests are being updated to the new example.   The changes to the example required updating many of the tests.

mcking65:
@jongund wrote:

> I can focus the `h1` in this example, but I think it requires a more general discussion to determine if this is a good idea as a general practice.

Agree; we need to continue to improve our guidance on focus movement.

>Most people expect to start tabbing from the beginning of the page. I think it will be unexpected for keyboard only users to start from the middle of the page, and what browsers do by default.

True if loading a new page. Often these trees are in apps that don't load a new page. And, in this case we do not.

When we rebuilt facebook.com, for perf reasons, most links do not load a new page. So, we don't get default browser behavior on load. That forced us to decide what to do. In most cases, we are focusing an H1. So far, we have had very positive feedback in user research from both screen reader and keyboard users.

>Asking authors to now actively set keyboard focus on a non-interactive element just doesn't sound right to me, since they will probably not do it correctly, and even if it done correctly it's not what most users will expect.

Well, you have a point there. This is why skip links almost never work. I'm not sure that is an argument against the technique of focusing the H1. The best we can do is model what we expect and provide resources and tools that help authors get it right more often than not.

BTW, Sina and I are **NOT** suggesting that this is what you want to do when loading a new page. But, there are circumstances where it makes sense.

JAWS-test:
In the [treeview navigation example](https://raw.githack.com/w3c/aria-practices/update-tree-nav/examples/treeview/treeview-navigation.html) I can't see the [location](https://www.w3.org/WAI/WCAG21/Techniques/general/G128) in the page in high contrast mode, because the position is only shown in the tree with a background color. I suggest to mark the current page with another visual means (e.g. underline or icon or italics)

mcking65:
@jongund commented:

> Changing the behavior of `tabindex=0` to not follow focus when navigating the tree, and having `tabindex=0` stay on the link with `aria-current=page` has the following issues:
> 
> 1. What happens if the link with `aria-current=page` is on a leaf that is hidden by the user?  When the user tabs out and back to the tree view, is the tree suddenly expanded to show the hidden link, or is the user never allowed to hide the link with `aria-current=page`?

If the user can't collapse a branch containing the node that has aria-current=page, I think the tree would feel broken. this would make navigating around the tree to explore or choose a different page more difficult.

Another option is to auto-expand on blur, if necessary,  to ensure the node that represents the current page is visible. We might want to consider also collapsing other branches that have been expanded to help ensure the visibility and findability of the current page node. This might better serve all users if you could do it in a way that doesn't feel disruptive or jarring.

If we show the current page node when tabbing in, that doesn't really help users who have not tabbed in. Think about users, including screen reader users using a reading cursor, who may have difficulty scanning through the tree to find where the current page is in the hierarchy.

> 2. If this change is made it is different from the previous treeview examples, we would need to describe this some place.

Definitely an accessibility feature to document.

mcking65:
@jongund, we have some merge conflicts here that I think are from #1553. I know you are already aware of #1554, but you may not have seen #1566, which changes linting specifically for the tests. Hopefully, these are very simple conflicts to resolve.

jongund:
@mcking65 
@sinabahram 
Pushed updates treeview navigation example to remove roving tabindex and leave tabindex=0 on aria-current, and also updated the regression tests.

examples/treeview/treeview-navigation.html
```diff
@@ -0,0 +1,640 @@
...
+            Mythical University footer information
+          </footer>
+        </div>
+        <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
+      </section>
+
+      <section>
+        <h2>Accessibility Features</h2>
+        <ol>
+          <li>The <code>aria-current="page"</code> attribute which link in the tree is associated with the current page or view.</li>
```
carmacleod, line 271: Missing word.
```suggestion
          <li>The <code>aria-current="page"</code> attribute indicates which link in the tree is associated with the current page or view.</li>
```

smhigley:
Independent of what we do with focus in this example, @sinabahram would you be up for sharing some more details about your comment that "often times, you can't do anything from that point, focus is locked, you have to exit into a virtual cursor" when focus is set to a container? I haven't encountered that before, and it seems important to fix for both routing and non-routing container focus use cases. Is there a particular screen reader/browser/set of interactions needed to trigger it?

I've made a demo page with three navigation methods (same-page anchor with no js, heading focus, and section focus) to help the navigation focus discussion: https://cdpn.io/smhigley/debug/XWKoJjo. Also, should this be moved to its own issue? I don't want to hold up the tree PR :)

sinabahram:
@smhigley  sure thing, and thank you for asking about it. Your tests will not exercise what I am talking about because none of them involve being in focus/forms mode. The way to test/exercise/surface what I am talking about is to use a tabs example, as this is where it happens so very many times. Just slap a tabindex of 0 on the container after the tab and you'll see this happen e.g. you switch to the second tab, let's say, and then want to go to its content. You hit tab, and bam you're stuck.

mcking65:
@sinabahram wrote:

> @smhigley sure thing, and thank you for asking about it. Your tests will not exercise what I am talking about because none of them involve being in focus/forms mode. The way to test/exercise/surface what I am talking about is to use a tabs example, as this is where it happens so very many times. Just slap a tabindex of 0 on the container after the tab and you'll see this happen e.g. you switch to the second tab, let's say, and then want to go to its content. You hit tab, and bam you're stuck.


I don't see that happening on the tabs examples. Yes, @smhigley, let's please move this discussion to an issue to make sure we get it sorted.

carmacleod:
@jongund

The SVG for aria-current needs to be inline and use currentColor, because using CSS ::after fails in Windows High Contrast Black.
The quickest way to check this is to use a [Color Contrast Analyzer](https://developer.paciellogroup.com/resources/contrastanalyser/) and type in `#005a9c` for the foreground and `#000000` for the background, which gives a failing ratio of 2.9:1. (If you prefer to test in actual Windows High Contrast Black, the easiest way is to run the example in Edge, then go to Windows Settings, Ease of Access, High Contrast and turn on High Contrast with the High Contrast Black theme).

Jon, I seem to remember you asking for suggestions for an SVG to represent aria-current (to replace the circle). I created a left-pointing arrow that (I think?) gets the point across (it would go inline after the text content in the aria-current anchor so that it gets the correct link text color in WHCM) - please give it a try and let me know what you think:
```
<svg xmlns="http://www.w3.org/2000/svg" width="14" height="9" viewBox="0 0 14 9">
  <g stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
    <line x1="2" y1="4" x2="7" y2="1" />
    <line x1="2" y1="4" x2="13" y2="4" />
    <line x1="2" y1="4" x2="7" y2="7" />
  </g>
</svg>
```

a11ydoer:
@carmacleod @jongund 
As the person who is assigned for visual design review also thought about the circle svg. I don't think we don't need any svg either circle or arrow. The function of svg circle or arrow is for visual indication of aria-current, where the users are in the site menu navigation. The most of the design I have seen for those is just adding the thick border or darkening the background color for the selected. 

In my mock up, I did NOT consider any color contrast issue Carolyn mentioned above but I did add boder styling for aria-current. This will be enough to indicate users where they are currently. 
added border styling for aria-current

a11ydoer:
The attached example just changed the font color from white to blue for currently selected menu item. 
visual design example - font color change for current menu item

a11ydoer:
Accessibility review

1. Improvement on color contrast -attached the potential solution but it is not complete yet. 
2. Tab order looks good.
3. Keyboard support works as documented. 
3. Touch target size issue in mobile(partly responsive design issue) 
color contrast improvment

carmacleod:
> As the person who is assigned for visual design review also thought about the circle svg. I think we don't need any svg, either circle or arrow.

Oops - sorry, @a11ydoer. I agree with you that we don't need an SVG to indicate the current page. I'm not so sure about using an outline to indicate the current page, though, because outline is currently being used as the focus indicator. For example, in this screen snapshot, Home is the current page and Overview is the focused item:

image

I like your suggestion about using a blue vertical bar on the left. That's what the (draft) Carbon TreeView uses to indicate the current page. In the following screenshot, Containers is current, and DevOps has focus:

image

a11ydoer:
@jongund 
We had a lengthy discussion about the behavior - focus automatically moved to right hand section when the menu was selected on the left. 

Here is the [detailed info](https://www.w3.org/2020/11/17-aria-apg-minutes.html#item04
) in the minute.

spectranaut (changes requested):
This is an approving review for the test review -- the tests look great!

And the code looks great as well except the css is not showing up in the "Open In Codepen" button, and this should be fixed before merging. I can spend a bit more time trying to figure out why it's not working.

jongund:
@mcking65 
Updated aria-current styling, but have not updated moving focus when a link is selected.  Seems like it needs more discussion in the group.  Whatever decision is made about moving focus, may also affect how tabindex=0 is handled, right now it stays on the item with `aria-current='page'` .

carmacleod:
@jongund
I created pull request https://github.com/w3c/aria-practices/pull/1643 on your update-tree-nav branch which adds an orange box around the cautionary note.

Turns out that the w3c base stylesheet already has a class called "advisement" that does almost exactly what we want, except that for some reason, the w3c "advisement" style has borders on both the left and right. I decided that the right border looked inconsistent with the green note box (which is right above and only has a left border), so I removed the right border from .advisement class in our examples/css/core.css file. Hope you agree.

carmacleod (approved):
This is looking really good @jongund!
Only 2 more things:
- merge the style from pr #1643
- @mcking65's upcoming wording in the Accessibility Features section

examples/treeview/css/treeview-navigation.css
```diff
@@ -0,0 +1,156 @@
+#ex1 header {
```
spectranaut, line 1: This is why the CSS isn't applying to the example when it is opened in codepen! Can you use something else of the CSS specifiers? Same reason as here: https://github.com/w3c/aria-practices/pull/1612/files#r535718223

spectranaut (changes requested):

examples/treeview/css/treeview-navigation.css
jongund: fixed

jongund:
@spectranaut @mcking65 
Fixed code pen issue and linting issues

spectranaut (approved):
Awesome, this is great! Thanks Jon. This is an approving code review.

cspell.json
```diff
@@ -34,7 +34,9 @@
         "Cook'n",
         "Copernicium",
         "Coughlin",
+        "Cpolygon",
```
carmacleod: should take this out as soon as #1656 is merged

cspell.json
```diff
@@ -34,7 +34,9 @@
...
         "crossorigin",
+        "Csvg",
```
carmacleod: should take this out as soon as #1656 is merged

cspell.json
mcking65: Merged in new file and it is gone.

cspell.json
mcking65: Merged in new cspell file and csvg is now gone.

examples/treeview/treeview-navigation.html
```diff
@@ -0,0 +1,741 @@
...
+              </td>
+            </tr>
+            <tr data-test-id="treeitem-aria-current">
+              <td></td>
+              <th scope="row"><code>aria-current="page"</code></th>
+              <td><code>a</code></td>
+              <td>
+                <ul>
+                  <li>Applied to the <code>treeitem</code> that is the link to the currently displayed page.</li>
+                  <li>The visual indication of the <code>treeitem</code> with <code>aria-current</code> is an SVG circle with a transparent fill after the label text.</li>
```
mcking65, line 645: We changed this look, right?

mcking65 (approved):
Editorial review is complete. We are just about ready to merge.
I have one open question about line 645.

jongund:
@mcking65 @carmacleod 
I removed the section about default borders around treeitems, we only need default borders around buttons.
I also added a note in the "Focus Movement After Content Load" section when keyboard focus is moved:
Note: Keyboard only users will need to navigate back to the navigation treeview to activate other items to view changes in content.  To make navigation back to the treeview efficient, the treeview should be one tab stop away from the element with focus.

examples/treeview/treeview-navigation.html
a11ydoer: @mcking65 
yes. You are correct. We removed the svg circel from indicating "aria-current"

mcking65:
@jongund commented:

> I removed the section about default borders around treeitems, we only need default borders around buttons.

Thank you!

> I also added a note in the "Focus Movement After Content Load" section when keyboard focus is moved:
> Note: Keyboard only users will need to navigate back to the navigation treeview to activate other items to view changes in content. To make navigation back to the treeview efficient, the treeview should be one tab stop away from the element with focus.

Excellent addition. Thank you.

I revised a bit to remove the word "should", reduce words,  and to emphasize the need for the design to accommodate the tab sequence. We wouldn't want people to make a visually unpredictable tab sequence to make it 1 tab stop away.

examples/treeview/treeview-navigation.html
mcking65: @a11ydoer 
But we have something else in its place, right? How would you describe the appearance of the new current page indicator? And, is it made with SVG?

a11ydoer:
Communicated with @jongund regarding color contrast issue. - tree item links in hover status. 

Foreground color, #005a9c and Background color #ccc has color contrast ratio of 4.44:1. This fails  Nomal text WCAG AA,  WCAG AAA, Large text WCAG AA.

a11ydoer (changes requested):
Tree item link hover color contrast issue can be addressed before the merge.

examples/treeview/treeview-navigation.html
a11ydoer: I see that documentation was already updated for aria-current page with below, which is correct. 

"The visual indication of the treeitem with aria-current is a vertical bar to the left of the treeitem label."

jongund:
@a11ydoer 
Updated hover color

a11ydoer (approved):
Changed background color, #ADDDFF now has color contrast ratio of 4.94:1 and it pass Normal text WCAG AA, Large text WCAG AA and WCAG AAA.

a11ydoer:
@mcking65 and @jnurthen 
Would you two mind reviewing this PR so that we can merge this? Once you are done, please approve PR or check your name in review check list.

examples/treeview/treeview-navigation.html
```diff
@@ -0,0 +1,742 @@
...
+        </div>
+        <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
+      </section>
+
+      <section>
+        <h2>Accessibility Features</h2>
+        <h3>Focus Movement After Content Load</h3>
+        <p>
+          An important aspect of designing a navigation tree experience is the behavior of keyboard focus when an item in the tree is activated.
+          If activating a tree item changes content on the page without triggering a browser page load, i.e., works like typical single-page apps, the focus position after the content load significantly effects efficiency for keyboard and assistive technology users.
```
smhigley, line 345: Basic spelling fix

```suggestion
          If activating a tree item changes content on the page without triggering a browser page load, i.e., works like typical single-page apps, the focus position after the content load significantly affects efficiency for keyboard and assistive technology users.
```

jongund:
@mcking65 
Fixed bug with aria-current link being visible when treeview loses focus and added test case to regression tests.  So this should be ready to merge.

test/util/assertAriaOwns.js
```diff
@@ -0,0 +1,48 @@
+const assert = require('assert');
+
+/**
+ * Confirm the aria-owns element.
+ *
+ * @param {obj} t                  - ava execution object
+ * @param {String} elementSelector - the element with aria-owns set
```
nschonni, line 7:
```suggestion
 * @param {string} elementSelector - the element with aria-owns set
```

examples/treeview/js/treeview-navigation.js
```diff
@@ -0,0 +1,505 @@
...
+      event.stopPropagation();
+      event.preventDefault();
+    }
+  }
+}
+
+/**
+ * ARIA Treeview example
+ * @function onload
+ * @desc  after page has loaded initialize all treeitems based on the role=treeitem
```
nschonni, line 496:
```suggestion
 * @description  after page has loaded initialize all treeitems based on the role=treeitem
```

examples/treeview/js/treeview-navigation.js
```diff
@@ -0,0 +1,505 @@
...
+class NavigationContentGenerator {
+  constructor(siteURL, siteName) {
+    this.siteName = siteName;
+    this.siteURL = siteURL;
+    this.fillerTextSentences = [];
+
+    this.fillerTextSentences.push(
+      'The content on this page is associated with the <a href="$linkURL">$linkName</a> link for <a href="$siteURL">$siteName</a>.'
+    );
+    //  this.fillerTextSentences.push('The text content in this paragraph is filler text providing a detectable change of content when the <a href="$linkURL">$linkName</a> link is selected from the menu.  ');
```
nschonni, line 21: Delete commented out code

examples/treeview/js/treeview-navigation.js
```diff
@@ -0,0 +1,505 @@
...
+
+    if (flag) {
+      event.stopPropagation();
+      event.preventDefault();
+    }
+  }
+}
+
+/**
+ * ARIA Treeview example
```
nschonni, line 494:
```suggestion
 * ARIA Treeview example
 *
```

mcking65:
@jongund 
Thank you for all your work on this. This is a huge step forward!
We have all required reviews complete, all tests are passing, and this is now ready to ship!

carmacleod:
@nschonni 
It seems that your suggestions and Matt's review crossed paths within a minute of each other somewhere on the internet, so your suggestions weren't merged.
Maybe create a new little PR for your suggestions?

nschonni:
Thanks @carmacleod I spun up https://github.com/w3c/aria-practices/pull/1728

PR #1572, opened 2020-10-15
Accessible Name Guidance by Role Section: Change marquee name from required to discretionary

smhigley:
Related to https://github.com/w3c/aria/issues/1339 and https://github.com/w3c/aria/pull/1342 on the ARIA repo.

I chose "discretionary" to match the guidance for log.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1572.html" title="Last updated on Oct 23, 2020, 4:35 AM UTC (ee56d0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1572/c2ee25e...ee56d0f.html" title="Last updated on Oct 23, 2020, 4:35 AM UTC (ee56d0f)">Diff</a>

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -4586,7 +4586,7 @@ So, As advised by <a href="#naming_rule_avoid_fallback">Rule 4: Avoid Browser Fa
             </tr>
             <tr>
               <td><a href="#marquee" class="role-reference"><code>marquee</code></a></td>
-              <td>Required</td>
+              <td>Discretionary</td>
               <td>Use <code>aria-labelledby</code> if a visible label is present, otherwise use <code>aria-label</code>.</td>
             </tr>
             <tr>


aria-practices.html
```diff
@@ -4586,7 +4586,7 @@ <h4><span id="naming_role_guidance_heading">Accessible Name Guidance by Role</sp
             </tr>
             <tr>
               <td><a href="#marquee" class="role-reference"><code>marquee</code></a></td>
-              <td>Required</td>
+              <td>Discretionary</td>
               <td>Use <code>aria-labelledby</code> if a visible label is present, otherwise use <code>aria-label</code>.</td>
```
jnurthen, line 4590: I think this is ok but we could consider changing this to match the other discretionary roles which describe why you might want to give it an accessible name. However - I'm not sure I can really come up with a valid use case for that.

jnurthen (approved):

aria-practices.html
carmacleod: It's valid to give a marquee a name, just didn't make sense to require it.

I very quickly tested [4 marquees](https://carmacleod.github.io/playground/marquee-test.html) using jaws/nvda in chrome/ff.

Notes:
- The first 2 marquees use (obsolete) HTML `<marquee>` elements. One has `aria-labelledby` and the other has `aria-label`.
- The next 2 are fake marquees that use a span (I didn't add any animation). As above, one has `aria-labelledby` and the other has `aria-label`.
- Both browsers expose the role correctly, but I had to use `role="marquee"` ("belts and braces" style) on the 2 HTML `<marquee>` elements (otherwise, they just had a generic text container role).
- Both browsers expose all 4 accessible names correctly, and they also expose the `aria-live="off"`.
- Both screen readers read the `aria-labelledby` span, because it's visible text content. :)
- Neither screen reader read the `aria-label`.
- Both screen readers read all of the marquee's content, i.e. they ignored the animated scrolling.

So we could say something along the lines of, "At this time, screen readers do not support marquee role well. When naming a marquee, `aria-labelledby` with a visible label is recommended, because some screen readers do not read an `aria-label` on a marquee."

I'll try Safari/VO tomorrow.

carmacleod (approved):
Thanks, @smhigley !
Ok to go in as-is, but consider adding more specific info in the guidance column.

mcking65 (approved):
Thank you @smhigley!

Issue #1548, opened 2020-09-30
Update Branch Policies for default branch


nschonni:
On the https://github.com/w3c/aria-practices/settings/branches I think the Travis Required check and add the new "lint" one that seems to be covering all the new jobs.
@mcking65 I could potentially rename the cspell key so you could keep that one as optional.

Think this will block https://github.com/w3c/aria-practices/pull/1523 unless you want me to leave the Travis file with a no-op job till this is updated.

/cc @michael-n-cooper

nschonni:
https://github.com/w3c/aria-practices/pull/1549 should help picking the job names in the UI

michael-n-cooper:
@nschonni Matt asked me to look into this, but I can't tell from the issue description what exactly you need done. Can you provide more context?

nschonni:
I think the minimal "fix" is to go onto https://github.com/w3c/aria-practices/settings/branches and edit the branch policy for master and uncheck Travis as a required check.
I'm not sure if the Actions equivalents actually need to be marked as required, but that would be something for the working group to decide. That could be done separately though

If you do the first part an mark Travis as no longer a required check, then https://github.com/w3c/aria-practices/pull/1576 can land, and you could disable the Travis integration (optional)

mcking65:
@michael-n-cooper 

Unfortunately, this has become urgent.

We aren't using travis for any real checks; it is just a no-op. But, it is [rejecting all our jobs now](https://travis-ci.com/github/w3c/aria-practices/requests) with the message:

>Job rejected
>Owner w3c is not on a new pricing.

This seems to flipped right around 12:30 AM pacific on December 21.

So, we can't merge anything that expects a travis check because travis never returns an answer even for our no op job.

mcking65:
Closing as complete.

We have decided to leave the checks run by the github workflows as optional instead of required for merge. @a11ydoer  and I will only merge if all checks pass unless we are absolutely positive a failure is not due to a problem with the changes in the PR. We want to keep the flexibility of being able to merge in such situations, e.g., some kind of infra problem.

nschonni:
Administrators can always merge things with an override if there is a failing required check. That still adds and extra checkbox of "I really want to override the failing job" rather than an accidental merge with a failure.

PR #1576, opened 2020-10-21
chore: Remove Travis-CI config

nschonni:
Coverted over to GitHub Actions.
Depends on #1548 to be completed


mcking65:
Unfortunately, this PR and #1548 have become urgent.

Travis is [rejecting all our jobs now](https://travis-ci.com/github/w3c/aria-practices/requests) with the message:

>Job rejected
>Owner w3c is not on a new pricing.

This seems to flipped right around 12:30 AM pacific on December 21.

So, we can't merge anything that expects a travis check because travis never returns an answer even for our no op job.

What horrible timing for a change like this ... right when tons of people are out.

nschonni:
@mcking65 looks like this can land now that Travis is no longer a required check

PR #1584, opened 2020-10-25
Position date picker dialog below help text

### LINKED ISSUE #1531: Dialog Datapicker Calendar button in incorrect place
Hi it looks like the calendar button on the [dialog datapicker example](http://localhost:8000/dialog-modal/datepicker-dialog.html) is in the incorrect place. It is over the text below the input box which explains the expected format of the date. I'm not handy enough with css to know how to debug/fix it!

danieldafoe:
Originally intended to address issue #1531. As it turns out, that issue might only be reproduceable on Linux and is slightly different; it is related to the position of the button, not the dialog.

[Preview](https://raw.githack.com/danieldafoe/aria-practices/fix/date-picker-dialog-overlaps-help-text/examples/dialog-modal/datepicker-dialog.html)


carmacleod:
Thanks, @danieldafoe!
We discussed this and decided that the dialog could go right under the button/text field, i.e. margin-top: 0; (or don't set at all).
This would have the effect of completely obscuring the help text, but when the dialog is open the help text isn't really necessary.
Do you agree?

danieldafoe:
@carmacleod Ah, that makes sense indeed! I can make that change instead.

danieldafoe:
@carmacleod I hope you don't mind, but my design eye found that `margin-top: 0` was a bit too close to the icon, so I chose a value that gave a bit more breathing room, yet still covers the help text.

Let me know what you think!

carmacleod (approved):
Looks great!
Thanks @danieldafoe !

spectranaut (approved):
I'm perfectly fine with this change but it doesn't actually solve the issue of the calendar butotn obscuring the help text, in chrome or firefox, as seen in my comment here: https://github.com/w3c/aria-practices/issues/1531

a11ydoer:
Screen capture is in the is[sue 1531](https://github.com/w3c/aria-practices/issues/1531#issuecomment-745474635)

mcking65:
Per comment from @spectranaut, I'm removing the association between this PR and #1531. Nonetheless, we have consensus that this is an improvement in the dialog positioning. So, going forward with the merge.

Thank you @danieldafoe

Issue #1594, opened 2020-10-29
Occasional regression test timeout when waiting for link to load


spectranaut:
We have a few times seen failures related to timeouts when trying to test functionality in an APG example that results in following a link to an external page.

Here is an example: https://github.com/bocoup/aria-practices/pull/43 You can see if fail on the first commit and pass on the next commit. (The commits just include debugging statements in them). I originally saw this failure in https://github.com/w3c/aria-practices/pull/1424.

When switching from travis to github checks, we saw some failures to follow through on links and discussed options: https://github.com/w3c/aria-practices/pull/1515

One option is to have a policy of not having external links in our example widgets.

mcking65:
@spectranaut 
Regression 2 is consistently failing. It failed in #1532 and #1533. It appeared to be time outs in both cases. I am assuming because of this external link issue.

I guess I was able to merge both of those PRs even with the failures because we haven't yet changed the branch policies (#1548). Is that right @nschonni?

I have no problem with a no external link policy in examples. That does seem like it would be a bit of work to clean up. There are several examples with external links.

How to best clean up the external links is a good question. In some cases, like the disclosure nav example and new tree nav example that is in progress, we use the link to load some content. In other places, like the grid of search results, the links are just placeholders and don't have to do anything. Could we just have them all popup an alert that says that if this were a real link it would go some place? Or, should they just do nothing at all? @jongund, what do you think?

Would a stopgap be to let those tests fail until we get the external links removed?

nschonni:
> I guess I was able to merge both of those PRs even with the failures because we haven't yet changed the branch policies (#1548). Is that right @nschonni?

Correct, none of the jobs are marked as required currently except for Travis. I left a note over there for Michael to at least make Travis no longer required. I'm not sure if you wanted to make the new GitHub jobs required, but you can leave a note there if you do.

spectranaut:
@mcking65 a stopgap could be to mark those tests as failing or comment them out, it's only a handful of tests. I don't have a problem doing that as long as it's a priority to get this fixed in the next month or so, which I think we definitely can. When commenting out I can create a list on this issue to "audit" of all external links in examples here and we can decide something reasonable to do.

[...]

PR #1600, opened 2020-11-06
Changes tests to not follow external links

spectranaut:
Fixes issue #1594

Following external links in tests to test key board interactions with links cause many time out related failures in github checks. I replace the four tests to replace the `href` on the dom element with a url fragment to test the keyboard behavior.

### DIFF
--- a/test/tests/link_link.js
+++ b/test/tests/link_link.js
@@ -103,18 +103,24 @@ ariaTest(
       let linkLocator = By.css(ex.linkSelector);
       let linkElement = await t.context.session.findElement(linkLocator);
 
-      await linkElement.sendKeys(Key.ENTER);
-      await t.context.session
-        .wait(() => {
-          return t.context.session.getCurrentUrl().then((url) => {
-            return url != t.context.url;
-          });
-        }, t.context.waitTime)
-        .catch(() => {});
+      // Update url to remove external reference for dependable testing
+      const newUrl = t.context.url + '#test-url-change';
+      await t.context.session.executeScript(
+        function () {
+          let [selector, newUrl] = arguments;
+          document.querySelector(selector).onkeydown = function () {
+            goToLink(event, newUrl);
+          };
+        },
+        ex.linkSelector,
+        newUrl
+      );
 
-      t.not(
+      await linkElement.sendKeys(Key.ENTER);
+
+      t.is(
         await t.context.session.getCurrentUrl(),
-        t.context.url,
+        newUrl,
         'ENTER key on element with selector "' +
           ex.linkSelector +
           '" should activate link.'
--- a/test/tests/link_link.js
+++ b/test/tests/link_link.js
@@ -108,7 +108,7 @@ ariaTest(
       await t.context.session.executeScript(
         function () {
           let [selector, newUrl] = arguments;
-          document.querySelector(selector).onkeydown = function () {
+          document.querySelector(selector).onkeydown = function (event) {
             goToLink(event, newUrl);
           };
         },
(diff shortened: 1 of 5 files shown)


test/tests/link_link.js
```diff
@@ -103,18 +103,24 @@ ariaTest(
       let linkLocator = By.css(ex.linkSelector);
       let linkElement = await t.context.session.findElement(linkLocator);
 
+      // Update url to remove external reference for dependable testing
+      const newUrl = t.context.url + '#test-url-change';
+      await t.context.session.executeScript(
+        function () {
+          let [selector, newUrl] = arguments;
+          document.querySelector(selector).onkeydown = function () {
+            goToLink(event, newUrl);
```
rwaldron, line 112: There's no `event` here

rwaldron: I think you can make this whole thing re-usable: 

```js
function onKeydownGoToLink(t, linkSelector, newUrl) {
  await t.context.session.executeScript(
    function () {
      let [selector, newUrl] = arguments;
      document.querySelector(selector).onkeydown = function (event) {
        goToLink(event, newUrl);
      };
    },
    linkSelector,
    newUrl
  )  
}
```

And called like: 

```js
const newUrl = t.context.url + '#test-url-change';
onKeydownGoToLink(t, ex.linkSelector, newUrl);
```

I've looked through the other cases and I think this meets them all.

test/tests/menu-button_links.js
```diff
@@ -270,13 +258,24 @@ ariaTest('"enter" on role="menuitem"', exampleFile, 'menu-enter', async (t) => {
...
+    const newUrl = t.context.url + '#test-url-change';
+    await t.context.session.executeScript(
+      function () {
+        let [selector, index, newUrl] = arguments;
+        document.querySelectorAll(selector)[index].href = newUrl;
+      },
+      ex.menuitemSelector,
+      index,
+      newUrl
+    );
```
rwaldron, line 271: My suggestion above could be used to replace this by providing a different selector: 

```js
onKeydownGoToLink(t, `${ex.menuitemSelector}:nth-of-type(${index}n)`, newUrl)
```

For a quick example: 
```js
document.querySelector('body:nth-of-type(1n)') === document.querySelectorAll('body')[0]
```

test/tests/treeview_treeview-2a.js
```diff
@@ -272,21 +272,24 @@ ariaTest(
...
+    // Update url to remove external reference for dependable testing
+    const newUrl = t.context.url + '#test-url-change';
+    await t.context.session.executeScript(
+      function () {
+        let [selector, newUrl] = arguments;
+        document.querySelectorAll(selector)[0].href = newUrl;
+      },
+      ex.linkSelector,
+      newUrl
+    );
```
rwaldron, line 284: Same replacement suggestion as my previous comment

rwaldron (changes requested):

test/tests/link_link.js
spectranaut: I appreciate the cleaner code but we can't actually apply this one to the others -- the [link example](https://w3c.github.io/aria-practices/examples/link/link.html) shows how to use aria to communicate something is a link that has onclick and onkeydown handlers, so there is a "goToLink" global function defined in the example page. The others are just regular links so we just need to replace the "href".

spectranaut:
Thanks for the feedback @rwaldron ! I added a helper function for all cases but the link case. It def cleans up the code :)

mcking65:
@spectranaut We should have some documentation about this technique and helper function on our wiki page about writing regression tests.

PR #1605, opened 2020-11-09
make filename in link text match href

carmacleod:
Code was changed to use `js/Menubutton.js` but link text in the "Javascript and CSS Source Code" section still said `Menubutton2.js`.
This PR just deletes the `2` in the link text's filename.


mcking65:
Ah, good catch ... I just merged all the other menubutton changes; I should have made that change there. Mind rebasing?

carmacleod:
@mcking65 
> Mind rebasing?

Actually, it turns out that that whole section has been rewritten, so this change is no longer necessary.
Closing.

PR #1607, opened 2020-11-09
updated indexes to include high contrast support information

jongund:
Updates the reference tables and coverage report to include information about high contrast support.
Many examples with high contrast support do not document the support.

In the reference table and the coverage report, examples with high contrast support have "(HC)" after the example title and a title attribute is added to the example link with the content "High Contrast Support".

In the coverage report there are two additional lists, one with examples with high contrast support and the other of examples without high contrast support.

[Preview link to reference table](https://raw.githack.com/w3c/aria-practices/index-with-hc/examples/index.html).

[Preview link to coverage report](https://raw.githack.com/w3c/aria-practices/index-with-hc/coverage/index.html).


mcking65:
@jongund, thank you, adding this to agenda.

One concern is that The meaning of(HC) is not clear to readers of the reference tables who don't see the title attribute. Is there room for a column titled "Supports High Contrast"? Wonder what ideas/thoughts others have on that.

jongund:
@mcking65 
The problem with adding a column is that the example references are currently in a list, so it would make the  table much more complex to generate if each example had to be in it's own role.  But maybe a new "table of examples", could have columns for roles, properties/states and high contrast.

jongund:
@mcking65 
I added a note before the table on the meaning of the "HC" abbreviation.  There maybe better solutions, but this is at least a step forward.

jongund:
@mcking65 
Looks like the conflict has been fixed.

spectranaut (approved):
The code changes all look sound to me! :)

examples/index.html
```diff
@@ -55,17 +56,17 @@ <h2 id="examples_by_role_label">Examples by Role</h2>
             <td><code>button</code></td>
             <td>
               <ul>
-                <li><a href="button/button_idl.html">Button  (IDL Version)</a></li>
-                <li><a href="button/button.html">Button</a></li>
+                <li><a href="button/button_idl.html" title>Button  (IDL Version)</a></li>
```
smhigley, line 59: We probably shouldn't append the `title` attribute if there is no title string

scripts/reference-tables.js
```diff
@@ -321,8 +324,14 @@ addLandmarkRole(['region'], true, 'Region Landmark', 'landmarks/region.html');
...
 function exampleListItem(item) {
+  let highContrast = '';
+  let title = '';
+  if (item.highContrast) {
+    highContrast = ' (<abbr title="High Contrast Support">HC</abbr>)';
+    title = 'High Contrast Support';
+  }
   return `
-                <li><a href="${item.ref}">${item.title}</a></li>`;
+                <li><a href="${item.ref}" title="${title}">${item.title}</a>${highContrast}</li>`;
```
smhigley, line 334: The empty `title` issue could be fixed if the `title="` string were part of the variable, e.g.:

```
let title='';
(...)
title='title="High Contrast Support"`;
(...)
<a href="etc"${title}>
```

Though I still think the link tooltip is better left off entirely :D

examples/index.html
```diff
@@ -55,17 +56,17 @@ <h2 id="examples_by_role_label">Examples by Role</h2>
...
               </ul>
             </td>
           </tr>
           <tr>
             <td><code>checkbox</code></td>
             <td>
               <ul>
-                <li><a href="checkbox/checkbox-1/checkbox-1.html">Checkbox  (Two State)</a></li>
-                <li><a href="checkbox/checkbox-2/checkbox-2.html">Checkbox  (Mixed-State)</a></li>
+                <li><a href="checkbox/checkbox-1/checkbox-1.html" title="High Contrast Support">Checkbox  (Two State)</a> (<abbr title="High Contrast Support">HC</abbr>)</li>
```
smhigley, line 68: I'm not a big fan of having a tooltip that says "High Contrast Support" on the link. It's a rather weird experience for a sighted user, tooltips have a bunch of known accessibility and usability issues, and we already have one on the abbreviation.

If we want a programmatic association on the link itself for screen reader users, we could use `aria-describedby` pointing at a single node.

mcking65:
Per conversation in [March 30 meeting](https://www.w3.org/2021/03/30-aria-apg-minutes.html):
1. Complete work on #1709 first.
2. Remove title attribute; no need for aria-describedby.

jongund:
@mcking65 
Since it was only a few lines of code I added the high contrast information on examples to pull #1709.

PR #1613, opened 2020-11-12
typo: Double word 'the'

nschonni:
Removed the extra "the" inside the link text rather than outside


carmacleod (approved):
+1

Heh, easy review. :)

PR #1639, opened 2020-11-30
Spin Button Date Picker: Fix unreliable date math test

spectranaut:
When you get towards the end of the month and try to add a month, Javascript Date objects start to do unexpected things. This is causing failures in the CI for all regression tests run today. This is the issue:

```javascript
let date = new Date('10/31/2020'); 
date.setMonth(date.getMonth + 1);
```
The second line increases the date by one month, resulting in date being December 1st, 2020, because there is no November 31st.

The fix is to put the day of the month to the 1st using `date.setDate(1)` before doing any math, because all months have the 1st of the month.

### DIFF
--- a/test/tests/spinbutton_datepicker.js
+++ b/test/tests/spinbutton_datepicker.js
@@ -408,6 +408,7 @@ ariaTest(
 
 ariaTest('up arrow on month', exampleFile, 'spinbutton-up-arrow', async (t) => {
   let date = new Date();
+  date.setDate(1); // This is necessary to do the correct date math for months.
 
   let monthSpinner = await t.context.session.findElement(
     By.css(ex.monthSelector)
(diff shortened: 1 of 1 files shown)


carmacleod (approved):
+1
Makes sense.
Thanks @spectranaut !

mcking65:
@spectranaut 
Seems like we now have a problem with arrow key testing on day spinner

test/tests/spinbutton_datepicker.js
```diff
@@ -408,6 +414,7 @@ ariaTest(
 
 ariaTest('up arrow on month', exampleFile, 'spinbutton-up-arrow', async (t) => {
   let date = new Date();
+  date.setDate(1); // This is necessary to do the correct date math for months.
```
nschonni, line 417: Is this part still needed?

Issue #1426, opened 2020-06-15
Page up/down on day and month for spin button is underspecified and has bug


spectranaut:
To reproduce:
- Go to the [date picker spin button example](https://w3c.github.io/aria-practices/examples/spinbutton/datepicker-spinbuttons.html)
- Focus on the day spinner
- Change the spinner to have the value of the last day of the month
- Press "page up"
- BUG: Notice that the day becomes the first of the month, not the 4th day of the month

If you change the date spinner to be the first of the month, and send "page down", you get a value that is 4 days before the end of the month. So you can see, page up and page down act differently at the boundaries, and the correct behavior is not explained in the Keyboard Support table.

I think we should always see a 5 unit increase and decrease, rather than turning it into the first or last day available!

Page up and page down for month has the same buggy behavior.

nschonni:
Was playing around with the tests today to see if I could fix it. One thing I was trying was to use END/HOME at the start of the test to get consistent results, but then I ran into the PAGEUP/PAGEDOWN issue where wrapping around the year seems off. EX:
- December + PAGEDOWN = July
- July + PAGEDOWN = February
- February + PAGEDOWN = December

Not sure if the expected behaviour is to still go -5 over the date boundry or if the current "Max Month" is correct

PR #1640, opened 2020-12-01
fix: Spin Button Date tests and sync spec to current behaviour

nschonni:
https://github.com/w3c/aria-practices/issues/1426 suggests that the behaviour should be updated in the future.
This enables the tests reliably and adds the current undocumented behaviour in the example around some keystrokes.

### DIFF
--- a/examples/spinbutton/datepicker-spinbuttons.html
+++ b/examples/spinbutton/datepicker-spinbuttons.html
@@ -198,6 +198,7 @@
             <td>
               <ul>
                 <li>Increases the value 5 steps.</li>
+                <li>When focus is on the <q>Day</q> spin button and the value is the last day of the month, changes value to the first day of the month.</li>
                 <li>When focus is on the <q>Month</q> spin button and the value is <q>December</q>, changes value to <q>January</q>.</li>
               </ul>
             </td>
--- a/examples/spinbutton/datepicker-spinbuttons.html
+++ b/examples/spinbutton/datepicker-spinbuttons.html
@@ -199,7 +199,7 @@
               <ul>
                 <li>Increases the value 5 steps.</li>
                 <li>When focus is on the <q>Day</q> spin button and the value is the last day of the month, changes value to the first day of the month.</li>
-                <li>When focus is on the <q>Month</q> spin button and the value is <q>December</q>, changes value to <q>January</q>.</li>
+                <li>When focus is on the <q>Month</q> spin button and the value is <q>December</q>, <q>November</q>, <q>October</q>, <q>September</q>, or <q>August</q>, changes value to <q>January</q>.</li>
               </ul>
             </td>
           </tr>
--- a/examples/spinbutton/datepicker-spinbuttons.html
+++ b/examples/spinbutton/datepicker-spinbuttons.html
@@ -189,7 +189,7 @@
               <ul>
                 <li>Decreases the value 5 steps.</li>
                 <li>When focus is on the <q>Day</q> spin button and the value is the first day of the month, changes value to the last day of the month.</li>
-                <li>When focus is on the <q>Month</q> spin button and the value is <q>January</q>, changes value to <q>December</q>.</li>
+                <li>When focus is on the <q>Month</q> spin button and the value is <q>January</q>, <q>February</q>, <q>March</q>, <q>April</q>, or <q>May</q>, changes value to <q>December</q>.</li>
               </ul>
             </td>
           </tr>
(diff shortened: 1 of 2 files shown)


mcking65:
Thank you @nschonni 
I'd like to have @spectranaut take a look

spectranaut (changes requested):
thanks so much for the fixes to downarrow! But I don't think we should fix the failing tests related to page down and page up -- those tests are failing because the example code is incorrect, as is tracked in issue #1426. We don't want to have tests that pass because they don't run into the bug that currently exists in the example.

Can you revert the changes to page up/page down in this PR?

nschonni:
@spectranaut I've updated the spec and test to reflect the current behaviour, then #1426 can be used to update it to the new behaviour

spectranaut (approved):
As long as Matt is ok with the change in the description of the `page up`/`page down` on the spin button, then I approve this PR.

nschonni:
@spectranaut feel free to cherry pick part of the fix into #1639 if you think this might be slower to land and want to get the test green quicker 😄

spectranaut:
I cherry picked the first commit (`Fix: Down arrow test`) into this PR: https://github.com/w3c/aria-practices/pull/1639

So if you could drop it from this PR we can consider the rest of the changes :)

nschonni:
👍 when that one lands I'll rebase and it'll drop out. If i remove it now, the tests will fail 😄 . Although because of the GitHub Actions issues earlier today, there are a bunch of failures right now

mcking65:
Based on above comments and commits in #1639, I merged that one. Do we still need something from this PR?

nschonni:
@mcking65 the other PR had just the first commit from this PR, now gone because I rebased. The rest of the commits took a similar approach to enable the other tests marked failing and updated the example description to match the current behaviour. The flip side to these changes is that @spectranaut opened the issue #1426 because the current behaviour might not be what is wanted in the future.

spectranaut:
It would be nice to get some eyes on this change -- what is the preferred way for the spin button date picker to behave at month and year boundaries with "page up" and "page down"? Should it always move the spinner 5 up or 5 down? Or, once you get to within 5 of the last day or first day of the month, should it just jump the last or first day of the month?

What do you think, @carmacleod, @jongund ?

nschonni:
Opened #1653 as an alternate approach to change the rollover behaviour

jongund:
@spectranaut @carmacleod 
I think the buttons should not roll over, e.g. when increasing it should stop at the maximum value, when decreasing it should stop at the minimum value.

mcking65:
@jongund commented:

> I think the buttons should not roll over, e.g. when increasing it should stop at the maximum value, when decreasing it should stop at the minimum value.


If we are going to change the behavior, let's make that a different issue and PR. Personally, in date pickers, I kind of like the ones where increasing the value just rolls time forward across boundaries, e.g., increasing day on Jan 31 changes date to Feb 1. However, stopping at boundaries is just as valid.

Jon, if you think we should have a discussion of the behavior, please raise an issue.

I'm going to go ahead with this PR so we at least have documentation and testing that matches the current behavior.

examples/spinbutton/datepicker-spinbuttons.html
```diff
@@ -185,11 +185,23 @@ <h2 id="kbd_label">Keyboard Support</h2>
...
+              </ul>
+            </td>
           </tr>
           <tr data-test-id="spinbutton-page-up">
             <th><kbd>Page Up</kbd></th>
-            <td>Increases the value 5 steps.</td>
+            <td>
+              <ul>
+                <li>Increases the value 5 steps.</li>
+                <li>When focus is on the <q>Day</q> spin button and the value is the last day of the month or any of the last four days, changes value to the first day of the month.</li>
```
mcking65, line 201:
```suggestion
                <li>When focus is on the <q>Day</q> spin button and the value is any of the last five days of the month, changes value to the first day of the month.</li>
```

examples/spinbutton/datepicker-spinbuttons.html
```diff
@@ -185,11 +185,23 @@ <h2 id="kbd_label">Keyboard Support</h2>
...
+                <li>When focus is on the <q>Month</q> spin button and the value is <q>December</q>, <q>November</q>, <q>October</q>, <q>September</q>, or <q>August</q>, changes value to <q>January</q>.</li>
```
mcking65, line 202:
```suggestion
                <li>When focus is on the <q>Month</q> spin button and the value is any of the last five months of the year, changes value to <q>January</q>.</li>
```

mcking65 (changes requested):
Thank you @nschonni!

I'd like to merge, but have some editorial suggestions to simplify the language.

mcking65 (approved):
Thank you, ready to ship now!

PR #1642, opened 2020-12-01
Button Examples: Add Open in Codepen and inline SVG in HTML using SVG "def"s

spectranaut:
#### Inlining in HTML

In https://github.com/w3c/aria-practices/pull/1534, @carmacleod suggested we use SVG defs to inline the SVGs into the HTML so they will load in codepen. See the code for the button example for an implementation of this suggestion.

[Preview Button Example](https://raw.githack.com/w3c/aria-practices/SVG-define/examples/button/button.html)


carmacleod:
Very cool, @spectranaut !

Nice work-around for :before and :after!
What tool did you use to get the svgs encoded with the special URL characters?

Not sure why the codepens are untitled? Did you create this branch from slightly old code? :)

Noticed that the menubar example needs some margin between it and the codepen button. Not sure why other examples have enough space? Changing the margin to have 5px on top (and bottom) fixes it, but I'm not sure if that's the best fix.
```css
.menubar-editor {
    margin: 5px 0;
    ...
}
```

spectranaut:
You are right @carmacleod I was on old code, I just rebased! Titles should be in now.

I used this URL encorder: https://yoksel.github.io/url-encoder/

Also, do you think this is a ok direction to move forward with the APG? Essentially, we will have to require that a "image" folder is no longer used, and instead the options are to inline the SVG in the CSS or inline them in the HTML.

carmacleod:
@spectranaut 
> Also, do you think this is a ok direction to move forward with the APG? Essentially, we will have to require that a "image" folder is no longer used, and instead the options are to inline the SVG in the CSS or inline them in the HTML.

Good question. I don't know enough about SVGs to know what people "usually" do. We need to be sure we don't break Windows HCM, which is why we converted some of the other examples over to inline SVGs in the HTML and made the SVGs use currentColor. I don't know if currentColor works for inline CSS content SVGs? It might.

Unfortunately, the [editor menubar](https://w3c.github.io/aria-practices/examples/menubar/menubar-editor.html) icons don't yet work properly in high contrast mode (not because of this PR - I guess we didn't convert the SVGs over to be inline and use currentColor, and I see that the grey triangles and blue dots and checkmarks fail in High Contrast Black). We did fix the [navigation menubar example](https://w3c.github.io/aria-practices/examples/menubar/menubar-navigation.html) though (inlined triangles that use currentColor), so maybe we can do something similar for editor menubar. (And then we can test whether currentColor works in CSS content url()'s or not).

I think the button example is good to go! The SVG defs work nicely for that one. Thanks for rebasing - title working now. Did you want to split this into 2 PRs so button example can go in right away? It seems we have some discussion and/or further svg coding/testing still to do for the editor menubar example.

spectranaut:
Thanks for the suggestions @carmacleod ! I'll break this up into two PRs and use the navigational menubar as an example for the inline-css option instead.

examples/js/examples.js
```diff
@@ -159,7 +159,10 @@ aria.widget.SourceCode.prototype.createCode = function (
 
     switch (childNode.nodeType) {
       case Node.ELEMENT_NODE:
-        this.createCode(sourceCodeNode, childNode, indentLevel, false);
+        // To remove the SVG defitions from the example's source code, add the id `svg_definitions`
```
nschonni, line 162:
```suggestion
        // To remove the SVG definitions from the example's source code, add the id `svg_definitions`
```

carmacleod (approved):
+1
Very nice, @spectranaut !
The mute off/on icons look great in both CodePen and High Contrast Mode (Black and White - tested in Edge).

mcking65:
@spectranaut 
Now we have this new feature in examples.js that uses the id`svg_definitions` to effect display of the HTML source. How can we enforce consistency in its use? Would it be possible to have a lint rule that can identify when it should be used?

PR #1643, opened 2020-12-01
Add orange box style for the cautionary note

carmacleod:
Wrap caution paragraph/list in a div with w3c "advisement" class, which gives the div a light orange background with a darker orange border on the left.

### DIFF
--- a/examples/css/core.css
+++ b/examples/css/core.css
@@ -123,6 +123,3 @@ table.data.attributes tbody td {
   box-shadow: inset 0 3px 5px 1px hsl(216, 82%, 30%);
 }
 
-.advisement {
-    border-style: none none none solid;
-}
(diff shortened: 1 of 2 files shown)


carmacleod:
Urgh. Apparently we're not allowed to override w3c styles. :(

The regression test failure is for "down arrow on day" which is being fixed in https://github.com/w3c/aria-practices/pull/1640 and/or https://github.com/w3c/aria-practices/pull/1639.

carmacleod:
Ok, I took out my css override and opened issue https://github.com/w3c/respec/issues/3233.
So for now we'll have to live with the inconsistent-looking right border in the orange box.

Everything is passing now except for the "down arrow on day" regression test mentioned above, so I think this is good to go.

mcking65:
interesting, I was unaware of the advisement class ... is that in the tr 2016 base css?

carmacleod:
> interesting, I was unaware of the advisement class ... is that in the tr 2016 base css?

I think so. The respec people asked me to open my [right border in advisement style is inconsistent](https://github.com/w3c/tr-design/issues/231) issue over in the tr-design repo.

Chiramate1993 (commented):
Review

PR #1644, opened 2020-12-01
chore: Explicit htmlparser2 devDependency

nschonni:
Was playing around with https://github.com/mysticatea/eslint-plugin-node, but didn't find enough here to PR it right now.

It did flag that `report.js` requires this library, but there isn't a formal dependency on it. It only works today because it's currently a dependency of `cherio`. I left the version the same as the `cherio` dependency to minimize the diff, but there is a newer version of the library released


spectranaut (approved):
What a random catch, good to have this little fix in!

PR #1648, opened 2020-12-03
disclosure examples: Support high contrast for graphics in Codepen by using current color in SVTG in CSS content

spectranaut:
This involves moving some PNG to SVGs in order to have the SVG inline with the CSS. We need to do this because CSS files cannot have relative paths that work correctly in Codepen (as codepen files do not have a file structure).

To review this PR, we need to review:
1. visual change of having an SVG instead of a PNG for the down and right arrows which indicate whether a node is expanded or collapsed.
2. That the "open in Codepen button" works correctly

[Preview Disclosure FAQ Example](https://raw.githack.com/w3c/aria-practices/SVG-move-png/examples/disclosure/disclosure-faq.html)

[Preview Disclosure Image Example](https://raw.githack.com/w3c/aria-practices/SVG-move-png/examples/disclosure/disclosure-img-long-description.html)


carmacleod (approved):
+1
Looks good - CodePen buttons work, including titles! :)

If you could change the fill and stroke for the SVGs to be 'currentColor', then they would pass in Windows High Contrast Mode.
I tested that currentColor works by changing the ::before content url's fill/stroke from #034575 to currentColor _**in the CodePen**_!! 🎉.
Let me know if you want me to open a separate issue for this.

spectranaut:
@carmacleod I added `currentColor` to get it in :) thx for alerting me to it!

cspell.json
```diff
@@ -34,7 +34,9 @@
         "Cook'n",
         "Copernicium",
         "Coughlin",
+        "Cpolygon",
```
carmacleod, line 37: don't need this after #1656 is merged

cspell.json
```diff
@@ -34,7 +34,9 @@
...
         "crossorigin",
+        "Csvg",
```
carmacleod, line 39: don't need this after #1656 is merged

mcking65:
Something is odd; looks like this PR is only changing to currentcolor for fill and stroke. Master already has the PNGs converted to SVG in CSS. I'll go ahead with it. I don't understand how the changes are already in master.

PR #1614, opened 2020-11-13
Add example of disclosure navigation with top-level links

smhigley:
Adds a new disclosure menu example page with top-level links in addition to disclosures, based on the 10/27 APG meeting. Related to #89. Also updates the disclosure nav js file to use a class pattern based on the updated code style guide.

[Preview Link](https://raw.githack.com/w3c/aria-practices/disclosure-nav2/examples/disclosure/disclosure-navigation-hybrid.html)

### Review checklist

* [x] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @mcking65 
* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @carmacleod 
* [x] [Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review) by @a11ydoer 
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @shirsha 
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @spectranaut 
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @spectranaut

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1614.html" title="Last updated on Mar 3, 2021, 7:18 PM UTC (00c1af9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1614/7f56b9e...00c1af9.html" title="Last updated on Mar 3, 2021, 7:18 PM UTC (00c1af9)">Diff</a>

PR #1651, opened 2020-12-03
disclosure navigation Example: Add "Open in CodePen" button

spectranaut:
Hi everyone! This "Open to codepen" is in it's own PR because I'm not totally sure how to move forward with it.

Here is example: [Preview Disclosure Navigation Example](https://raw.githack.com/w3c/aria-practices/codepen-disclosure-navigation/examples/disclosure/disclosure-navigation.html)

The problem is that when we open these examples in codepen, the links send you back to the APG page instead of just updating the main content within the copepen HTML rendering. The reason why is that we send a `base` url to Codepen so that all images and links in the HTML lead back to the APG website.

The codepen example works up until you click a link, so it is probably useful to most programmers. My suggestion is that we leave it in this state for now, and see if there is an elegant solution to keep the example links working in codepen when the example is updated to our APG coding guidelines (this example hasn't been updated yet).

### DIFF
--- a/examples/disclosure/disclosure-navigation.html
+++ b/examples/disclosure/disclosure-navigation.html
@@ -57,7 +57,9 @@
   </label>
 
   <section>
-    <h2 id="ex_label">Example</h2>
+    <div class="example-header">
+      <h2 id="ex_label">Example</h2>
+    </div>
     <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of" ></div>
     <div id="ex1">
       <nav aria-label="Mythical University">
@@ -349,7 +351,7 @@
   <section>
     <h2>Javascript and CSS Source Code</h2>
     <!--  After the js and css files are named with the name of this example, change the href and text of the following 2 links to refer to the appropriate js and css files.  -->
-    <ul>
+    <ul id="css_js_files">
       <li>
         CSS:
         <a href="css/disclosure-navigation.css" type="tex/css">disclosure-navigation.css</a>
@@ -367,7 +369,7 @@
     <pre><code id="sc1"></code></pre>
     <div role="separator" id="sc1_end_sep" aria-labelledby="sc1_end_sep sc1_label" aria-label="End of" ></div>
     <script>
-      sourceCode.add('sc1', 'ex1');
+      sourceCode.add('sc1', 'ex1', 'ex_label', 'css_js_files');
       sourceCode.make();
     </script>
   </section>


carmacleod:
Hmm, I see what you mean.

I guess the [Menubar navigation](https://w3c.github.io/aria-practices/examples/menubar/menubar-navigation.html) has the same problem - sorry I didn't notice earlier.

[Treeview navigation](https://raw.githack.com/w3c/aria-practices/update-tree-nav/examples/treeview/treeview-navigation.html) will have the problem, too - currently being reviewed in https://github.com/w3c/aria-practices/pull/1558.

I wonder if we should maybe leave the codepen button off of the navigation examples until this problem is fixed?

How might updating to the latest APG coding guidelines help?

Also, curious whether we would still need CodePen's `base` url to point back to the APG site if we inline SVG's? Not many of the examples have a non-inlined image, except for the map in the disclosure example, maybe? What links still need to go back to the APG site?

spectranaut:
> How might updating to the latest APG coding guidelines help?

We definitely need an item about updating the APG coding guidelines to include things you can and can't do in order to open the example in codepen! But I'm really not sure the technical solution for these navigational menus. I haven't had a moment to really put on my thinking cap about it, however.

> Also, curious whether we would still need CodePen's `base` url to point back to the APG site if we inline SVG's? Not many of the examples have a non-inlined image, except for the map in the disclosure example, maybe? What links still need to go back to the APG site?

Unfortunately, we really do need the `base` url, for every other image that is referenced by a relative path in the HTML. Take the carousel examples, or the the disclosure image -- these images are pulled into the codepen directly from the APG when the codepen renders. It's just relative URLS from the CSS that don't work (they don't know about the `base` url). This makes me think we should maybe just leave all reference to SVGs in the HTML.

carmacleod:
Right, forgot about the carousel images.

Are there any examples that need the `base` url to point to APG site for links?
Just trying to understand the scope of the problem. :)

The [Menu Button example](https://w3c.github.io/aria-practices/examples/menu-button/menu-button-links.html) uses absolute URLs, so it's ok (hmm, but the codepen button in that example doesn't have the right style).

I guess the [Breadcrumb Example](https://w3c.github.io/aria-practices/examples/breadcrumb/index.html) uses relative links into the APG, so it currently needs the `base` url to be APG, but the links can probably be changed to absolute URLs.

Can't think of any others, but I didn't look deeply.

spectranaut:
i just looked through and your right, I don't think there are any relative links that are essential to the examples, so it is just the images.

I also just tried to remove the "base" url override locally to test if the example would work in codepen, and it didn't -- but it turns out it didn't because the "onclick" handler is never replaced because of an error! I just pushed a fix for that. With that fix, the example **does work** if you remove the "base" reference. So one option for a fix is to consider changing the way we bring in images and removing base to get the examples that use fake links to work.

But I also thought of another option -- if we are replacing the default behavior for activating a link in this example anyway, we could just call `event.preventDefault()`. When you add `event.preventDefault()` right now, though, the url fragment does not update and the menu does not close, so we would have to add that behavior as well. What do you think of this option?

cspell.json
```diff
@@ -34,7 +34,9 @@
         "Cook'n",
         "Copernicium",
         "Coughlin",
+        "Cpolygon",
```
carmacleod, line 37: don't need this after #1656 is merged

cspell.json
```diff
@@ -34,7 +34,9 @@
...
         "crossorigin",
+        "Csvg",
```
carmacleod, line 39: don't need this after #1656 is merged

carmacleod:
> But I also thought of another option -- if we are replacing the default behavior for activating a link in this example anyway, we could just call event.preventDefault(). When you add event.preventDefault() right now, though, the url fragment does not update and the menu does not close, so we would have to add that behavior as well. What do you think of this option?

Interesting thought, @spectranaut. I will have to defer to @smhigley on that, because she wrote DisclosureNav.
Sarah, please read the previous comments to understand the back story for Valerie's question. :)

spectranaut:
Resolved the conflicts!

zcorpan:
> So one option for a fix is to consider changing the way we bring in images and removing base to get the examples that use fake links to work.

I think I prefer this option if it can work for all examples. `base` elements are a bit confusing in my opinion, and adding code to work around the effect of the `base` element seems like it would add further complexity that distracts from what the example is actually about.

zcorpan:
However, looking at this particular example, the scrolling that happens by navigating to the fragment `#mythical-page-content` is a bit annoying and could be disorienting, so maybe it should `preventDefault()` anyway to avoid the scrolling.

smhigley:
@zcorpan this was discussed and resolved in the other disclosure nav example, I think it would make sense to adopt the same approach for this one: https://github.com/w3c/aria-practices/pull/1614#issuecomment-784607544

zcorpan:
I noticed that #1614 also added the "fake link behavior" code that was almost identical to that in the second commit of this PR, so I removed that commit here with a git rebase.

mcking65:
Although minutes for [October 19 meeting](https://www.w3.org/2021/10/19-aria-apg-minutes.html) do not indicate a merge decision, my memory of the discussion is that the current behavior of the example when viewed in Codepen, while not ideal, is acceptable. So, I will merge.

mcking65 (approved):
Approving based on memory of Octt 19, 2021 meeting discussion

PR #1653, opened 2020-12-04
fix: Allow spinbutton to roll over min/max values

### LINKED ISSUE #1426: Page up/down on day and month for spin button is underspecified and has bug

nschonni:
Todo:
- [ ] Fix bad math in re-enabled test
- [ ] Add spec text in example HTML

Fixes #1426 
Closes #1640

### DIFF
--- a/examples/spinbutton/js/spinbutton-date.js
+++ b/examples/spinbutton/js/spinbutton-date.js
@@ -88,7 +88,11 @@ SpinButtonDate.prototype.setValue = function (value, flag) {
 
   if (value > this.valueMax) {
     if (this.wrap) {
-      value = this.valueMin;
+      // Handle zero based month values
+      if (this.valueMin === 0) {
+        value--;
+      }
+      value -= this.valueMax;
     } else {
       value = this.valueMax;
     }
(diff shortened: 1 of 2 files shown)


nschonni:
@spectranaut didn't fully clean this up, but thought I'd share to see if it was what you were thinking as the correct behaviour

nschonni:
@spectranaut @jongund @mcking65 I'm closing this one since #1640 landed, but it could be used as a base if you decide to change the behaviour

PR #1663, opened 2020-12-09
Infrastructure: chore: Ignore coding-template from report CI

nschonni:
These are ignored the same way in the regression job, and likely don't get touched often. This syncs the `paths` values from that other job.
This will start to trigger on some of Dependabot PRs, but will then be able to see if those bumps cause any issues in the job. Alternately a branch filter could be added to prevent that


spectranaut (approved):
looks good! :)

PR #1669, opened 2020-12-14
Infrastructure: Add newline at EOF when generating example index

nschonni:

PR #1670, opened 2020-12-14
Update examples index

mcking65:
Ran reference-tables.js to refresh the examples index page.


nschonni:
@mcking65 I updated the JS in #1669 because the newline will be added by anyone manually editing the files with EditorConfig installed

PR #1681, opened 2020-12-15
Open in codepen: Button IDL 

spectranaut:
Add codepen to the Button IDL example using the same pattern as Button example in the https://github.com/w3c/aria-practices/pull/1642. There is no styling for this example, and I checked out code from a month ago and didn't see the styling there either.

[Preview Button IDL Example](https://raw.githack.com/w3c/aria-practices/codepen-button-idl/examples/button/button_idl.html)

### DIFF
--- a/examples/button/button_idl.html
+++ b/examples/button/button_idl.html
@@ -38,7 +38,9 @@
     </p>
 
     <section>
-      <h2 id="ex_label">Example</h2>
+      <div class="example-header">
+        <h2 id="ex_label">Example</h2>
+      </div>
 
       <p>
         <strong>IMPORTANT:</strong> This example uses features of the draft ARIA 1.2 specification. As a draft specification, it is subject to change. Support provided by browsers or assistive technologies is experimental.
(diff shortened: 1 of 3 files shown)


carmacleod (approved):
+1

Looks good!
Probably should delete that old svg now, and the images directory, because they're no longer needed.

> There is no styling for this example

This example is supposed to be almost identical to the button example, except for the few changes that show how aria attributes can be set using reflection in the js instead of setAttribute() or hardcoding in the html. So the css is exactly the same as the button example. :)

spectranaut:
hi @carmacleod, thanks for the fast review as always! I actually did delete the image directory when converting the other button example, so the "mute"/"sound" SVG is missing from the button IDL example in master.

I'm confused, iin Firefox and Chrome both, -- the styling for button is not applied successfully to button IDL. All the styling is based on `[role="button"]` in the CSS file, and it seems like that is just not being applied to the button IDL? Did it used to apply? Is there a bug in the example that has been introduced?

carmacleod:
@spectranaut 
Interesting. It works ok for me (on Windows) in Chrome, but not Edge or Firefox.
I'll see if I can figure anything out.
(very weird that it works in Chrome but not Edge - @smhigley any suggestions for where to look?)

spectranaut:
I made an issue about this weird styling issue as it is unrelated: #1692  otherwise this PR is mergable, @mcking65 ! :)

mcking65 (approved):

a11ydoer:
As @spectranaut said two buttons need styling and codepen example is already done. I will assign this to myself @mcking65.

carmacleod:
Wait, @a11ydoer - please read https://github.com/w3c/aria-practices/issues/1692

PR #1682, opened 2020-12-15
chore: Remove run-node dependency

nschonni:
Doesn't seem like the dependency is used anymore


spectranaut (approved):
Good catch! Your right, this was used for greenkeeper to handle merge conflicts. thanks :)

mcking65 (approved):
Thank you for this clean up @nschonni!

PR #1694, opened 2020-12-18
Add "Open in Codepen" button to Treegrid example

spectranaut:
Adding the open in Codepen button! This involved moving some SVGs and moving some code.

[Preview](https://raw.githack.com/w3c/aria-practices/codepen-treegrid/examples/treegrid/treegrid-1.html)

### DIFF
--- a/examples/treegrid/css/treegrid-1.css
+++ b/examples/treegrid/css/treegrid-1.css
@@ -91,7 +91,8 @@
   cursor: pointer;
 
   /* Load both right away so there is no lag when we need the other */
-  background-image: url("expand-icon.svg"), url("expand-icon-highlighted.svg");
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='12'%3E%3Cpolygon fill='black' points='2,0 2,10 10,5'%3E%3C/polygon%3E%3C/svg%3E%0A"),
+    url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='12'%3E%3Cpolygon fill='hsl(216, 94%25, 50%25)' points='2,0 2,10 10,5'%3E%3C/polygon%3E%3C/svg%3E%0A");
   background-repeat: no-repeat;
 }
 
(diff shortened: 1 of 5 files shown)


examples/treegrid/css/treegrid-1.css
```diff
@@ -91,7 +91,8 @@
   cursor: pointer;
 
   /* Load both right away so there is no lag when we need the other */
-  background-image: url("expand-icon.svg"), url("expand-icon-highlighted.svg");
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='12'%3E%3Cpolygon fill='black' points='2,0 2,10 10,5'%3E%3C/polygon%3E%3C/svg%3E%0A"),
```
nschonni, line 94: Oops, I think my assumption about the trailing `;` in the cspell ignore `"url\\(\"data\\:image/svg\\+xml.*\"\\);"` is bad 😦

carmacleod: @nschonni 
Heh, yeah - apparently it can be a comma, too... :)
https://developer.mozilla.org/en-US/docs/Web/CSS/background-image

carmacleod (approved):
Looks good!

Approving now, even though it can't be merged until the cspell ignore regex is fixed.

@nschonni Is the following regex correct?
`"url\\(\"data\\:image/svg\\+xml.*\"\\)[,;]"`

nschonni:
@carmacleod makes sense to me, or the punctuation could possibly be dropped entirely. I was just trying to make sure the regex didn't get to greedy, but maybe that's not really an issue for this

carmacleod:
Thank-you, @nschonni - I created https://github.com/w3c/aria-practices/pull/1696 to add `[,;]` to the end of the regex. Good point about probably not needing to be that specific because it's pretty unlikely that anything else will ever match, however I decided to check  for the end punctuation anyhow. 😄 

I have no idea how to test whether merging 1696 will allow this PR to pass the spell-check. (but I think it should work...)

mcking65 (approved):
Looks great ... this would have been a perfect opportunity to improve how we set aria-current. I think it should be on the link inside the list item, not on the entire list item. Rendering aria-current on a long item is awkward for screen readers. I'll raise a separate issue for that though. I want to get this merged.

Issue #1619, opened 2020-11-17
iOS VoiceOver issue in Editable Combobox With List Autocomplete Example


useit-consulting:
I'm encountering a major focus-related issue with the [Editable Combobox With List Autocomplete Example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html) and iOS VoiceOver. It makes the component inaccessible for iOS VoiceOver users.

There's an [active discussion over at MaterialUI](https://github.com/mui-org/material-ui/issues/22956) where I've provided some videos of the effect. But since the effect also happens in this APG pattern, maybe it should be addressed here.

In short the problem is this:

- Swiping through the page or using explore-by-touch with VoiceOver triggers focus and blur events.

- The listbox is closed on textfield blur.

- As a result, the user cannot reach the contents of the listbox, since the listbox closes automatically once you start searching for it. 

- I've tried swiping backwards from the keyboard, closing the keyboard with the keyboard "Close" button or using explore-by-touch. In all instances, the listbox disappears before I can interact with it.

### Device details
Tested using iPhone SE and iOS 14.1 in Safari and Chrome

### Other info
This issue is **not** present in Android Talkback + Chrome. It does not appear that Talkback triggers focus/blur events the same way VoiceOver does.

The issue is also **not** present in the older APG 1.1 pattern [List Autocomplete with Manual Selection](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html) 

I believe the difference is that the 1.1 pattern keeps the listbox open on blur, while the 1.2 example closes automatically. So a possible solution might be to keep it open like the 1.1 version.

mcking65:
@useit-consulting 

Wow, I'm surprised we didn't catch this. Thank you for calling it to our attention.

This is perplexing. because leaving it open on blur doesn't seem like desirable behavior for desktop keyboard users. Maybe there is another way to address that.

I'm surprised that VoiceOver is creating blur event by default. That also seems problematic. It doesn't seem like navigating with VoiceOver should have that kind of side effect. It definitely limits the ability of VoiceOver users to explore what is happening on screen.

@sinabahram what do you think of this VoiceOver behavior?

StommePoes:
Definitely can't leave things positioned over the page open when users leave it. How did the 1.1 version prevent "hanging popups" for these lists?

[...]

PR #1699, opened 2020-12-22
Editable Combobox With List Autocomplete: Remove "onBlur" events.

spectranaut:
Adding the change request in this comment: https://github.com/w3c/aria-practices/issues/1619#issuecomment-748882630

In summary:
- Removed the "on blur" event which causes the listbox to close every time the textbox loses focus
- Close the listbox every time there is a click outside of the listbox or text box
- There was a request to close the listbox on `tab`, this is already supported 

[Preview](https://raw.githack.com/w3c/aria-practices/combobox-VOMac-fix/examples/combobox/combobox-autocomplete-list.html)

### DIFF
--- a/examples/combobox/js/combobox-autocomplete.js
+++ b/examples/combobox/js/combobox-autocomplete.js
@@ -515,7 +515,8 @@ class ComboboxAutocomplete {
   onBackgroundMouseUp(event) {
     if (
       !this.comboboxNode.contains(event.target) &&
-      !this.listboxNode.contains(event.target)
+      !this.listboxNode.contains(event.target) &&
+      !this.buttonNode.contains(event.target)
     ) {
       this.comboboxHasVisualFocus = false;
       this.setCurrentOptionStyle(null);
--- a/examples/combobox/js/combobox-autocomplete.js
+++ b/examples/combobox/js/combobox-autocomplete.js
@@ -57,7 +57,12 @@ class ComboboxAutocomplete {
       'focus',
       this.onComboboxFocus.bind(this)
     );
-    this.comboboxNode.addEventListener('blur', this.onComboboxBlur.bind(this));
+
+    document.body.addEventListener(
+      'mouseup',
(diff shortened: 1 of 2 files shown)


spectranaut:
fixed merge conflicts!

useit-consulting:
I've tested the preview with VoiceOver on an iPhone SE with iOS 14.2 without issues 👍

jongund (approved):
@mcking65 
I made a small change to also include testing the button in the new `mousedown` event. and pushed the change to the branch  Otherwise the coding changes to not use the `blur` event look good to me.

carmacleod (approved):
+1

Looks good, @spectranaut - thanks!

examples/combobox/js/combobox-autocomplete.js
```diff
@@ -57,7 +57,12 @@ class ComboboxAutocomplete {
       'focus',
       this.onComboboxFocus.bind(this)
     );
-    this.comboboxNode.addEventListener('blur', this.onComboboxBlur.bind(this));
+
+    document.body.addEventListener(
+      'mouseup',
```
oliviertassinari, line 62: One note, mouseup fire when dragging the scrollbar, blur doesn't.

mcking65: Is there another issue to resolve here? Does that mean that scrolling the page unexpectedly closes the dropdown? As a screen reader user, I can't test this, so I'm not quite clear on what the problem is.

oliviertassinari: I was raising a regression (not sure about the importance of it) on the UX for mouse users. I'm not aware of any impact on screen readers.

mcking65:
@jongund @smhigley @carmacleod 

Could one of you investigate to see if we have a regression here?

jongund:
@mcking65 
I can look at it, maybe we need to switch this example to use pointer events.

Jon

PR #1705, opened 2020-12-24
IDL Button example: Make browser support note more visible

### LINKED ISSUE #1692: Button IDL styling does not work on all browsers
The Button IDL styling ([css/button.css](https://github.com/w3c/aria-practices/blob/master/examples/button/css/button.css)) is not applied for:
- Firefox, linux
- Chrome, linux
- Edge, Windows
- Firefox, Windows

Where as it does appear to work in Chrome on Windows. See some discussion in #1681

carmacleod:
Fixes #1692 as follows:
- use advisement class to put note in an orange box
- update wording
- include note in codepen

@spectranaut, I've added you as a reviewer because you understand the issue (i.e. the warning needs to be really visible).
Also, please note that I moved the warning paragraph so that it's now part of the CodePen content. I hope you agree that the warning needs to go with the code, at least for now. :)  If the user deletes it from the CodePen, that's on them... ;)

Note also that I added [target="_blank"](https://html.spec.whatwg.org/multipage/browsers.html#valid-browsing-context-name-or-keyword) to the github issue link in the note so that it would open in a new window from the codepen.

[Preview](https://raw.githack.com/w3c/aria-practices/issue1692/examples/button/button_idl.html)

### DIFF
--- a/examples/button/button_idl.html
+++ b/examples/button/button_idl.html
@@ -32,7 +32,7 @@
     </p>
     <p>
       The JavaScript for the example buttons on this page uses the <a href="https://www.w3.org/TR/wai-aria-1.2/#idl-interface">IDL Interface defined in ARIA 1.2</a>.
-      The purpose of these examples is to illustrate how to use IDL for ARIA Attribute Reflection and provide a test case for browser and assistive technology interoperability.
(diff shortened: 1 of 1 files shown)


mcking65 (approved):
Thank you @carmacleod! looks good.
I only made a minor editorial revision to simplify wording, and it's ready to ship.

PR #1709, opened 2020-12-29
fixed bug with indexing aria-label property and added high contrast information

jongund:
The changes in the this pull request:
* Fixed bug in `aria-label` also getting references to `aria-labelledby`, by using a regex, and this was also applied to checking roles.
* Updated code to use `node-html-parser` for identify the roles, properties and states associated with an example, making the code much easier to read and understand.
* Added the identification of examples with specific high contrast support features, since it only adds a few lines of code to the example and coverage script files.
* Examples with high contrast support have links with `aria-label` to indicate that the example has high contrast support (e.g. removed the use of the `title` attribute from previous proposals).  The information will also then be available in screen reader "list of links" feature/

NOTE: The changes added `node-html-parser` to the `package.json` file.

[Preview of example index](https://raw.githack.com/w3c/aria-practices/fix-bug-index-generator/examples/index.html)
[Preview of coverage report](https://raw.githack.com/w3c/aria-practices/fix-bug-index-generator/coverage/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1709.html" title="Last updated on Apr 12, 2021, 4:08 PM UTC (88a7cdc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1709/0553bb5...88a7cdc.html" title="Last updated on Apr 12, 2021, 4:08 PM UTC (88a7cdc)">Diff</a>


carmacleod:
Do we want it to also match things like `setAttribute('aria-label', label);`?
If so, maybe use a regular expression instead of `indexOf`?
Something like the following should work, I think?
Note that `\b` means "end of word", and double-backslash is needed for a single backslash in javascript strings.
```
const regex = RegExp(ariaPropertiesAndStates[i] + '\\b');
regex.test(code);
```

jongund:
@carmacleod 
We are only matching aria properties and states inside CODE elements in the Properties and States tables.  I am not real great with regex statements, but I think your suggestion is a good one.  I will update the pull request to use the regex.

Jon

examples/index.html
```diff
@@ -539,48 +538,26 @@ <h2 id="examples_by_props_label">Examples By Properties and States</h2>
...
-                <li><a href="treeview/treeview-2/treeview-2a.html">Navigation Treeview  Using Computed Properties</a></li>
-                <li><a href="treeview/treeview-2/treeview-2b.html">Navigation Treeview  Using Declared Properties</a></li>
               </ul>
             </td>
           </tr>
           <tr>
             <td><code>aria-labelledby</code></td>
             <td>
               <ul>
-                <li><a href="checkbox/checkbox-1/checkbox-1.html">Checkbox  (Two State)</a></li>
```
mcking65, line 583: This appears to be a bug; aria-labelledby is used and documented in the checkbox-1 example.

examples/index.html
```diff
@@ -652,17 +628,12 @@ <h2 id="examples_by_props_label">Examples By Properties and States</h2>
             <td><code>aria-multiselectable</code></td>
             <td><a href="listbox/listbox-rearrangeable.html">Listboxes with Rearrangeable Options</a></td>
           </tr>
-          <tr>
-            <td><code>aria-orientation</code></td>
-            <td><a href="slider/slider-2.html">Slider  with aria-orientation and aria-valuetext</a></td>
```
mcking65, line 657: Same aria-orientation bug mentioned above.

mcking65 (changes requested):
Jon, This was a great catch, thank you!!

We were definitely overlisting aria-label, and this is much better in that respect.

I see some examples where aria-label is used but not documented, but that is a separate issue. They should not be captured in the index (one of the layout grids and the rearrangeable listbox). I don't think it is worth capturing that in a separate issue because we already have issues calling for rework on both.

I think this may introduce a new bug that effects indexing of aria-orientation. Strange that it is only that one property.

carmacleod:
@jongund In case it's helpful, here's a cheat sheet I use for regex: http://www.rexegg.com/regex-quickstart.html#ref
I can never remember what characters to use. ;)

examples/index.html
jongund: @mcking65 
@carmacleod 
I will update the indexing to use the regex suggestion and see if that fixed the issue with `aria-orientation`.

jongund:
@mcking65 
@carmacleod 
Use the regex suggested by Carolyn and it seems it fixed the problems with `aria-orientation` bug.

mcking65 (changes requested):
Jon, this is looking better still.

Wow, isn't this devilishly difficult to fix/review? I notice a couple of places where the previous version fixed errors that are no longer getting fixed.

1. Previous version fixed list of examples using aria-hidden; it removed Alert Dialog, which does not use aria-hidden.
2. Previous version fixed list of examples using aria-posinset and aria-setsize; it removed File Directory Treeview Using Computed Properties, which do not use those properties.

I also noticed some other places where attributes are being overlooked by the algorithm.

1. The aria-current index is missing Breadcrumb Example.
2. The aria-modal index is missing Modal Dialog Example.
3. The aria-selected index is missing:
  * Date Picker Combobox Example
  * Select-Only Combobox Example.
  * Collapsible Dropdown Listbox Example
  * Listbox Example with Grouped Options
  * Scrollable Listbox Example

jongund:
@mcking65 
The `aria-current` issues is because the breadcrumb example is in a file named `index.html`.  The current script ignores files name `index.html`.  I think the breadcrumb example should be renamed to `breadcrumb.html` or could make the script only ignore the `index.html` file in the `examples` directory..

carmacleod:
Renaming the breadcrumb example to `breadcrumb.html` makes sense for another reason as well: the open in codepen button uses the filename to set a title for the codepen, and currently this is "index" for the breadcrumb example (which isn't a very good title).

mcking65:
@jongund 
I'm fine with including the breadcrumb file rename in this PR.

jongund:
@mcking65 
I think I have fixed the problems with the reference index and the coverage report, `aria-current` is now referencing both treeview navigation and breadcrumb examples..
I also renamed `index.html` file in the breadcrumbs directory to `breadcrumbs.html`
Please review and let me know if there ae still problems.

mcking65:
Per discussion in [March 30 meeting](https://www.w3.org/2021/03/30-aria-apg-minutes.html), Sarah will assist with a code review and help ensure it is robustly supporting our desired indexing behaviors.

I thought we had those behaviors documented in the wiki; I can't find it. I believe they should be part of the [infrastructure documentation](https://github.com/w3c/aria-practices/wiki/APG-Infrastructure). We should have documentation of:
1. What we expect to get indexed, e.g., what authors need to do to ensure a section or page is listed in the index for a specific topic.
2.Where the index generation code is and when/how it is run.

coverage/index.html
```diff
@@ -875,14 +951,17 @@ <h2 id="props_with_more_than_one_label">Properties and States with More than One
             <td><a href="../aria-practices.html#naming_with_aria-labelledby">Naming with Referenced Content Via  aria-labelledby </a>
 </td>
             <td><ul>
+        <li><a href="../examples/accordion/accordion.html">Accordion</a></li>
         <li><a href="../examples/checkbox/checkbox-1/checkbox-1.html">Checkbox  (Two State)</a></li>
+        <li><a href="../examples/coding-template/Depricated-MultipleImplementationExample-Template.html">EXAMPLE_NAME</a></li>
```
smhigley, line 956: This file (and the coding-template directory) should be excluded

smhigley (commented):
There's one bug that should be updated, but otherwise it looks OK. Looking at the logic though, I wonder if we should consider moving to a node html parser in the future, so we're not trying to process html strings ourselves.

I don't want to hold up this PR, but it might be a good backlog item to look into converting this all to use something like this library: https://www.npmjs.com/package/node-html-parser

jongund:
@smhigley 
Thank you for finding the bug, I will look at the parser, it looks like the right way to go.

smhigley:
@jongund awesome! I'm also 100% in favor of merging this PR now, and letting that be a future item, though 😄

jongund:
@smhigley 
Can you re-review I update the reference and coverage scripts to use `node-html-parser`.  Makes the code alot cleaner and easier to read!

jongund:
@mcking65 @smhigley @carmacleod 
I updated the reference and coverage report scripts based on Sara's feedback on using `node-html-parser`, which made the code much cleaner.  Adding the high contrast was only a few lines of code, so to speed up merging I just added it to this pull request.  No longer using the `title` attribute on the link to provide high contrast information to screen readers, using `aria-label` instead.  So if people can re-review I think it is ready to merge.

@nschonni 
NOTE: Not sure why one of the regression tests is failing, since I did not change any examples.  I tried re-runing the job, but it still fails.  This seems to happen when pull requests languish for long periods of time.

jongund:
Pull #1859 superceeds this pull request.

PR #1720, opened 2021-01-10
Infrastructure: Update Dependabot commit message

nschonni:
This seems to be the pattern used currently


mcking65 (approved):
@nschonni, very nice; this should make things a bit quicker for me. Thank you!

PR #1721, opened 2021-01-13
Color Viewer Slider Example with Mobile Support

jongund:
This is an updated version of the slider 1 example for picking a color.  

New features include:
1. Javascript file that uses a single prototype object for the example.
2. Use SVG graphics for buttons, thumbs and rail.
3. Improve high contrast support through use of  inline SVG graphics.
4. Mobile support with buttons

### Preview Link

[Color Viewer Slider with mobile support example](https://raw.githack.com/w3c/aria-practices/slider-color-viewer-mobile/examples/slider/slider-color-viewer-mobile.html)

### Review checklist

* [ ] [Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review) by @mcking65
* [ ] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @a11ydoer 
* [x] [Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review) by @a11ydoer
* [ ] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @charmarkk
* [ ] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @spectranaut
* [ ] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @spectranaut

NOTE: The Javascript and CSS is the same as the updated color picker example, they share the same CSS and Javascript files

PR #1746, opened 2021-02-02
Color Viewer Slider: Improve high contrast support, touch AT support, and APG code quality guidelines support

jongund:
Updated color viewer slider based on the results [slider update focus group meeting](https://github.com/w3c/aria-practices/wiki/Determining-Features-of-APG-Slider-Updates).

Changes:
* Using APG coding practices and created as single object for color viewer.
* Using SVG graphic elements, instead of HTML elements for the visual rendering of the sliders.  Including using the `fill` of the rail `rect` to indicate the color the slider is associated with. The `fill` to the left of the thumb is the color associated with the slider.
* Using `currentColor` value for SVG `stroke` property to support high contrast modes in operating systems.
* Using `event.key`, instead of `event.keyCode` for responding to keyboard events.
* Updated focus styling to have a ring around the thumb.
* Using `pointer` instead of `mouse` events for moving the slider thumb to support touch devices.
* Added `group` role and accessible name for the color viewer widget.

[Preview link of color viewer slider](https://raw.githack.com/w3c/aria-practices/slider-color-viewer-update/examples/slider/slider-color-viewer.html)

* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @carmacleod (Windows, Mac, iOS, iPadOS)
* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by  @jongund (Windows touch)
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by  @mcking65 (Voiceover)
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by  @carmacleod (Jaws/NVDA)
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by  @a11ydoer (Talkback)
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @shirsha (High Contrast Mode testing)
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @smhigley


carmacleod:
@jongund 

Nice! Sighted users on touch screens can drag the slider now.

Tested in Safari on iPhone and iPadOS, and Chrome on iPhone. I don't have any other touch devices.

I did notice a couple of problems:
1. If the user taps on the value text above the thumb, the text gets selected. Then they can't drag the thumb. I'm not sure how to work around that - it seems to be fairly easy to do by accident on touch screens. Maybe CSS `user-select: none;` on the text element?
   image

2. The focus ring for the thumb goes too close to the focus ring around the whole slider. This is not a problem on desktop, although it does look a bit "squished" (see first snapshot below), but on mobile the thumb focus ring bites into the slider focus ring (see second snapshot below).
   image
   image

carmacleod:
@jongund There seem to be 2 extra files accidentally committed to this PR?
- examples/treeview/js/treeview-navigation.js
- test/util/assertAriaOwns.js

jongund:
@carmacleod 
Can you give it another try, I did some update this morning.
Thank you for your reviews.

carmacleod:
@jongund 

Wow, Jon! Not only does this work better on touch, it actually works _really well_! The sliders slide beautifully. 🎉
Tested in Safari and Chrome on iOS and iPadOS.

Pointer-events are awesome!

Also, as @patrickhlauke discovered, this _actually works with VoiceOver_  now, too!!

Not quite sure _how_ VoiceOver is controlling the slider, though. The [AOM user actions](https://wicg.github.io/aom/explainer.html#user-action-events-from-assistive-technology) table says that slider increment/decrement should synthesize arrow key events, but that doesn't explain what's going on here because:
- the example was already handling arrow keys before these updates, and it didn't work with VoiceOver before
- the example is now using pointerup/down/move instead of mouseup/down/move, and now VoiceOver works

So is the platform using synthesized pointer events to control the slider?
And if so, then as @patrickhlauke pointed out, where are the x, y change values coming from?

@cookiecrook Do you know if iOS/iPadOS/VO has implemented something to make ARIA sliders work?

patrickhlauke:
It does indeed seem that Safari/VO now fires faked keyboard events. The reason why the old/current slider example doesn't work is because VO sends left/right arrow fake key events for the horizontal slider, but the current slider example listens for up/down arrow.

TalkBack on Android doesn't seem to fire these fake events, but seems to have implemented some other sort of accommodation, allowing users to double-tap and hold, and then slide left/right while holding the finger down to change the value. However, it only announces the percentage of the slider, not the actual value text. Also, it seems for some reason - at least in my testing - that the entire slider doesn't get announced at all unless it's been double-tapped to activate it. Really not sure what is going on there...

carmacleod:
@patrickhlauke 
> The reason why the old/current slider example doesn't work is because VO sends left/right arrow fake key events for the horizontal slider, but the current slider example listens for up/down arrow.

Thanks for confirming that it's key events being faked! I see the difference between the old APG slider and the one in this PR.
The old slider _did_ handle left/down (for decrement) and right/up (for increment), but it was still using `event.keyCode` (which I assume VO is not setting). As part of this modernization update for the old slider example, @jongund switched to using `event.key`, so that is likely the important difference there.

However, there's still something mysterious, because the [preview slider](https://raw.githack.com/w3c/aria-practices/slider-color-viewer-mobile/examples/slider/slider-color-viewer-mobile.html) in PR https://github.com/w3c/aria-practices/pull/1721 was also updated to use `event.key` and it handles left/right (and up/down) arrows, but it doesn't work with VO touch.

The only potentially relevant difference I can see is that the working slider in this PR hooks keydown on one additional element - an svg text element that is displaying the valuenow - but I don't see how that could be the important difference.

> TalkBack on Android ... implemented some other sort of accommodation...

Good to know that TalkBack kinda-sorta works also.

> However, it only announces the percentage of the slider, not the actual value text.

Seems to be a TalkBack bug. Even happens with range input, according to this [blog post](https://jamescatt.ca/blog/2020/we-still-cant-use-the-html-range-input-either).

> Also, ... the entire slider doesn't get announced at all unless it's been double-tapped to activate it.

Agreed that this is weird and seems buggy. Is this still the only way to report TalkBack bugs?
https://support.google.com/accessibility/contact/feedback

carmacleod (approved):
This is an approving Functional review for Windows, Mac, iOS, and iPadOS,
and an approving Accessibility review for Jaws/NVDA in Chrome/Edge/Firefox.

Keyboard works well on Windows and Mac (used Fn+arrows for PageUp, PageDown, Home, End on Mac),
and left-right tap and drag gestures work well on iOS and iPadOS.

Nice work, Jon! This is vastly improved from what was there before. Thank-you so much for your perseverance!

test/tests/slider_slider-color-viewer.js
```diff
@@ -735,31 +774,6 @@ ariaTest(
     // Send 1 key to blue slider
     const blueSlider = sliders[2];
     await blueSlider.sendKeys(Key.PAGE_DOWN);
-
-    t.is(
```
smhigley, line 739: Did you intend to remove these tests? If so, the two lines above them (775-776) should also be removed, since there are no tests checking the result of sending `key.PAGE_DOWN` (though it seems like the tests should probably be re-added for symmetry with red and green).

smhigley (changes requested):
Looks good! Just one issue in the tests, where it looks like some lines were (likely accidentally) removed.

examples/slider/js/slider-color-viewer.js
```diff
@@ -0,0 +1,380 @@
...
+    var valueNow = this.getValueNow(slider);
+    var valueMax = this.getValueMax(slider);
+
+    if (value > valueMax) {
+      value = valueMax;
+    }
+
+    if (value < valueMin) {
+      value = valueMin;
+    }
```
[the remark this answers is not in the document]
jongund: I made the change, thank you for the suggestion

test/tests/slider_slider-color-viewer.js
jongund: @smhigley 
Thanks for catching this, I added the rest of the test in.

mcking65 (commented):
### Review with VoiceOver

#### iOS Safari and VoiceOver

* Slider and labels are read correctly
* Adjust value gesture works for single increment (single-finger flick up/down)
* Likely VO bug: Adjust value for multiple increments (3-finger flick up/down) does not work
* Decrement to min or increment to max: does not apply (no gesture with iOS VO).

I don't think there are any bugs in the example code that are effecting VO/Safari. It seems the VO support is partial, but it is much, much better than prior version of slider.

#### Mac Chrome and VoiceOver

* Slider and labels are read correctly when tabbing and reading with vo-left/right.
* Adjust value with VoiceOver commands does not work. Press shift-vo-down to interact and then vo-up/down or vo-left/right to adjust and nothing happens.
* Adjust value with using up/down arrows with quick nav off adjusts the value. however, VoiceOver focus jumps to the HEX value field when up or down is pressed. At this point, vo-F3 reports the hex field value and vo-F4 reports that the slider has focus. As a result, VoiceOver does not read the new value of the slider as up/down are pressed. We should try to understand the root cause of this behavior. It might be a VO or Chrome bug, but is worth digging into.
* Increment 10 steps with pageUp/PageDown works and VO correctly reads the new value.
* Decrement to min and increment to max with home/end: Similar behavior to up/down arrows. The value changes correctly, but the VO reading cursor jumps and reads useless information.
* When tabbing from hex field to RGB field, the selection events cause the labels to not be announced. This seems like a VO bug.

I'm pretty sure that the problems are VO or Chrome problems, but we should investigate to be sure.


### Mac Safari  and VoiceOver

* Slider and labels are read correctly when tabbing and reading with vo-left/right.
* Adjust value with VoiceOver commands  works correctly. Press shift-vo-down to interact and then vo-up/down or vo-left/right to adjust and the values change and are correctly anounced.
* Adjust value with using up/down arrows with quick nav off works correctly; it adjusts the value and reads the new value correctly.
* Increment 10 steps with pageUp/PageDown works and VO correctly reads the new value.
* Decrement to min and increment to max with home/end adjusts the value, but VoiceOver does not read the new value. Similar to performing the same command in Chrome, the VoiceOver reading cursor jumps off the slider. Indicates this is a VoiceOver bug.
* Tabbing from hex field to RGB field works correctly.

Before making this an approving review, it would be good to somehow confirm that the quirky VO behavior is not related to anything in our code.

shirsha:
It is working fine in High contrast mode Black and High Contrast mode White on Edge browser.

smhigley (approved):

a11ydoer:
@patrickhlauke 

I appreciate your comment and feedback since I am assigned to test the example with Android Talkback. We are the partners. ;-) 

>TalkBack on Android doesn't seem to fire these fake events, but seems to have implemented some other sort of accommodation, allowing users to double-tap and hold, and then slide left/right while holding the finger down to change the value. 

The gestures I used were 1)swipe right to get to individual slider, 2)then activate it with double tap, 3)use two fingers to adjust the value(Thumb value). 
Then it announces the value, which is the next topic below. 

> However, it only announces the percentage of the slider, not the actual value text. Also, it seems for some reason - at least in my testing -

I found the same issue. First, I thought that announced value was the percentage of the slider. However,  the percentage informaiton is not even consistent. For example, the slider was near the end of left side, announced percentage was like 60%, not less than 50%. It is a bizarre situation. 

> that the entire slider doesn't get announced at all unless it's been double-tapped to activate it. Really not sure what is going on there...

When you say "entire slider", does it  mean an individual slider like a red slider or three sliders altogehter as the group? I was able to go through each slider, slide right gesture and double tapped to activate individual slider. 

@jongund @mcking65 

As @patrickhlauke noted in his comment, I have the same feeling - "Really not sure what is going on there..." In particular, the annoucned value does not make sense to me at all.  No clue where the value comes from and how that is associated with shown Thumb value in the slider. I would need more time to look into this issue.

jongund:
@mcking65 
I have updated the example to resolve issues brought up at the last APG meeting:
* Removed outer focus ring.
* Updated documentation on the use of current color.
* Updated documentation on the use of pointer events.
* Tested on Windows 10 using a touch screen, it works great.

mcking65:
With the new synthesized kb events availabel in VoiceOver and TalkBack, this warning is now out of date:

>WARNING! Users of touch-based assistive technologies may not be able to fully operate widgets that implement the slider pattern because APIs for capturing the necessary gestures and commands from assistive technologies are not yet available. Authors should fully test slider widgets using assistive technologies on primarily touch devices before considering incorporation into production systems.

However, our testing reveals that support is still incomplete and sometimes wonky. The current level of support will be tested by the aria-at project. Slider tests are on the current development schedule and will be complete in the next couple of months. However, that data will not show up in APG till sometime in the 3rd quarter or so.

Given this, should we drop this warning or reword it to something like:

>WARNING! Users of touch-based assistive technologies may experience difficulty utilizing some features of widgets that implement the slider pattern because support for capturing the slider gestures and commands from those technologies is new, and in some cases, incomplete. Authors should fully test slider widgets using assistive technologies on devices where touch is a primary input mechanism before considering incorporation into production systems.

mcking65:
For this note:

>NOTE: The use of inline SVG graphics are needed to support the use of the currentColor  value.

Don't we want to emphasize that the SVG needs to be inline? Would it be accurate to say:

>Note: To use the currentColor  value, the SVG graphics must be inline.

mcking65:
Instead of:

>The use of pointer  events supports assistive technologies on touch devices to change the value of slider using gestures.

Is the following accurate:

>the example supports slider operation commands issued by assistive technologies, including from touch-based assistive technologies, by listening for pointer events.

patrickhlauke:
yeah, pointer events have nothing to do with touch+AT support.

mcking65:
@patrickhlauke 

Do you think I should just remove this point from accessibility features documentation?

>The use of pointer events supports assistive technologies on touch devices to change the value of slider using gestures.

There is some difference between this slider implementation and the currently published one that makes this one more compatible with iOS VO and Talkback, but I don't know what it is. The difference seems worth documenting.

Also, what are your thoughts regarding the wording of the warning I proposed:

>WARNING! Users of touch-based assistive technologies may experience difficulty utilizing some features of widgets that implement the slider pattern because support for capturing the slider gestures and commands from those technologies is new, and in some cases, incomplete. Authors should fully test slider widgets using assistive technologies on devices where touch is a primary input mechanism before considering incorporation into production systems.

Do we still need a warning? If so, is that wording appropriate?

mcking65:
@jongund 

Could you please edit the top comment and add a succinct list of things that are new/different in this slider compared to the one that is already published? The meeting summary page on the wiki doesn't really summarize the differences. I use that information in the commit when I merge, and that becomes part of the change history that we reference when we publish.

patrickhlauke:
> Do you think I should just remove this point from accessibility features documentation?

yes, that bullet point should be removed.

> There is some difference between this slider implementation and the currently published one that makes this one more compatible with iOS VO and Talkback, but I don't know what it is. The difference seems worth documenting

I believe the big change that makes this work with the (new) behaviour in VO that sends faked key events (it didn't seem to work yet for TalkBack when I checked, but maybe I was not testing it right) is that the new code checks `event.key` rather than `event.keyCode` (and the faked events that VO generates don't have the latter, but do have the former)

For the warning, I'd tweak it slightly to something like this (long story short: it's the behaviour of the AT to send these fake key events that's new/incomplete)

> WARNING! Users of touch-based assistive technologies may experience difficulty utilizing some features of widgets that implement the slider pattern because assistive technology support for generating the synthetic key events based on the increase/decrease gestures is new, and in some cases, incomplete. Authors should fully test slider widgets using assistive technologies on devices where touch is a primary input mechanism before considering incorporation into production systems.

jongund:
@mcking65 
Add a section in the initial post on the changes to this example.

a11ydoer:
@patrickhlauke @mcking65 @jongund 
I posted the question to a11y slack channel to get talkback testing help. In particular, why the read value was not the same as the shown thumb value.

a11ydoer:
@patrickhlauke @mcking65 @jongund 

I tested the example again. Thumb value seems to go with percentage value for now. 

Test enviorment was Android version 11, Talk back version 60107667

a11ydoer:
In addition to a lag to read updated value, even slight touch of starting value is announced before adjusted thumb value. It is a bit flaky depending on how to use touch. In my testing, I used the two thumb gestures to changed the value.

a11ydoer (approved):
Touch gesture in Android Talkback is a bit flaky(See above comments). However, it seems to be working. I was also able to use control setting in Android to add actual percentage for RGB value.

mcking65:
@jongund 
I changed the heading levels inside the example so they are 3 and 4 instead of 2 and 3. That way, they both fall under the level 2 example heading. I didn't look at the css to see if that needed any adjusting .... pls fix if that impacted any styling.

mcking65 (approved):
This is approving editorial review.

mcking65:
I've finished edits for editorial review. It'd be good for people to have a final look at the warning at the top and the accessibility features just to make sure I don't have any typos or the like.

Otherwise, this is, I believe, ready to ship!

jongund:
@jongund 
The statement in the note "To use the currentColor value, the SVG graphics must be inline. " seems a little awkward to me, for example you can use currentColor value on referenced SVG graphics, it just doesn't change the styling for our purposes.  Maybe something more like: "The use of currentColor value is only supported on inline SVG graphics at this time."  But its up to you, I think developers will get the gist of the note as you wrote it.

jnurthen:
@a11ydoer for the Android issues it is probably worth re-testing with chrome 90. There are a bunch of input type=range bug fixes in there which could help
https://issuetracker.google.com/issues?q=id%3A(175959402%7C175917356%7C175930703%7C175917355)

a11ydoer:
@jnurthen 
Thanks for the suggestion and resources. I was not 100% confident about my findings. I will look into Google input type=range issues.

a11ydoer:
@jnurthen 
I am impressed that those bugs were fixed by Google as of yesterday and today!

carmacleod:
@mcking65 

I agree with @jongund that the inline SVG statement seems a little awkward. How about (fewest words possible 😄 )...

> SVGs must be inline to use currentColor.

mcking65:
Thank you @jongund for leading the work to make this awesome set of improvements!!!

PR #1755, opened 2021-02-08
Slider thermostat update from focus group meeting results

jongund:
I have updated the thermostat slider example using the vertical slider and two text sliders:
* Updated example to use SVG graphics elements
* Added high contrast support 
* Updated code to use APG coding practices
* Using `event.key` instead of `event.keyCode`
* Use pointer events, instead of mouse events
* Use the `presentation` role on SVG element
* Setting the CSS property `forced-color-adjust` to `auto` on the SVG elements.
* Using `stroke-opacity` and `fill-opacity` instead of `transparent` values for setting stroke and fill colors for the SVG `rect` used for focus ring for high contrast mode.
 
[Preview slider thermostat example](https://raw.githack.com/w3c/aria-practices/slider-thermostat-update/examples/slider/slider-thermostat.html)

### Review checklist

* [x] [Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review) by @smhigley 
* [ ] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) on Windows by @charmarkk
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) on MacOS by @carmacleod 
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) on Android by @a11ydoer (Note Talkback problem in the comment)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1755.html" title="Last updated on Mar 25, 2021, 2:41 PM UTC (bd7967a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1755/0553bb5...bd7967a.html" title="Last updated on Mar 25, 2021, 2:41 PM UTC (bd7967a)">Diff</a>


jongund:
@patrickhlauke 
Can you checkout this slider example using pointer events when you get a chance.
I would be grateful to you for any feedback you have.

carmacleod:
Functional review for Windows Chrome/Edge/FF, Mac Safari/Chrome/FF: all work well with mouse and keyboard.
(Used Fn+arrows to test PageUp, PageDown, Home, End on Mac).

Functional review for iOS and iPadOS Safari/Chrome: works pretty well in Safari - it's a little difficult to start the drag on the Fan and Heat/Cool sliders unless I zoom in, but then it's ok. Chrome seems to have zoom problems, because it works ok until I zoom in to try to have better control, and then the dragging becomes erratic (i.e. it just stops randomly, even though I haven't lifted my finger). I suspect that this is a Chrome issue, and probably not anything to do with the slider code. We can open a chromium issue after this PR is merged.

----

I recommend using a div or span to display the temperature at the top of the slider. The problem with using an output element is that it acts as a live region, and so the temperature is spoken twice for every movement of the slider (once for the new slider value, and once again for the live region).

patrickhlauke:
General comment: I'd love to see these examples use

```
<meta name="viewport" content="width=device-width, initial-scale=1.0">
```

(though that will then also require some tweaking of the table styles to avoid them pushing out the content/leading to horizontal scrolling of the whole page) to make the experience a bit more mobile/small-screen friendly.

This example works very well on a Surface touchscreen, on iPhone/Safari, reasonably in iPhone/Chrome (seems the webview used for Chrome gets a bit flakey with respecting the `touch-action` once pinch-zoomed in ... this would become a non-issue once mobile-friendly viewport is added with the above meta). Also works well with iPhone/VoiceOver. 

Android/Chrome/TalkBack announces percentage only when focusing a thumb. Users can only change the value of the vertical slider properly, by double-tap-and-hold-ing the thumb, then moving their finger. Changing percentage value is announced when moving. The Fan and Heat/Cool sliders can't be changed using TalkBack (trying double-tap-and-hold and then moving the finger on the screen doesn't cause any change in value, compared to the HTML range example one).

The fact that it only announces percent values becomes confusing/an issue in the Fan and Heat/Cold examples. Hearing that "Fan" is at "25 percent" is even more meaningless to users in this case. Yes, it's a shortcoming in TalkBack of course, but at least in the case of a numeric slider that represents a range of values, it can "almost" be understandable (though without knowing min/max either, yeah, it's of low information value for these users).

Incidentally, I'm actually wondering if using sliders for these two cases is semantically appropriate. They're not really offering a sliding scale of values that the user jumps between, but conceptually they are mutually exclusive radio buttons?

The `touch-action` values are set on the outer containers. Particularly due to the size of the vertical one, this can actually cause problems for touch users in general (non-AT), as the whole `<div>` suppresses vertical scrolling. Once a user is sufficiently zoomed in, they may find themselves unable to actually scroll as their vertical scroll gesture is blocked by the `touch-action: pan-x`. I'd suggest adding the `touch-action` to the `.slider-group`, rather than the `.slider-thermostat-vertical`. Generally best to just add `touch-action` to the "most specific / minimal" element to avoid suppressing regular actions unnecessarily.

as with the HTML range example, the fact that the slider for "Fan" was visually shorter than the labels/buttons underneath it was confusing, visually

and again, for some reason, Chrome/TalkBack also didn't announce the actual displayed temperature in the <output>. It just announced the element as "Temperature, status" without announcing the value. Seems TB is quite confused/broken, sadly.

jongund:
@carmacleod 
@patrickhlauke 
I added the meta tag and removed the output element.
I am not sure I understand the labeling issues, the labels are centered on the position of the thumb when that value is selected.
Can you point me to an example of what you were expecting?

Also the fan and heat/cool could be a radio group, but we are trying to demonstrate use of `aria-valuetext` and it looks like a slider.  So this is the age old question, do ARIA roles and behaviors describe visually what a control looks like (e.g. a slider) or its functional properties (e.g. a radio group).

patrickhlauke:
> I added the meta tag

still not seeing it in the markup ... think that was missed out in this commit https://github.com/w3c/aria-practices/pull/1755/commits/96d161b1c65fbff770c2480bba0f153de2361f11

> I am not sure I understand the labeling issues, the labels are centered on the position of the thumb when that value is selected.
> Can you point me to an example of what you were expecting?

screenshot in Safari/iOS shows the last label *way* off track. I was looking at that when I mentioned it in my comment, hadn't even noticed that in other cases it's not as dramatic.

Safari/iOS screenshot

in Android/Chrome, it looks mostly ok. Android/Firefox, it's all looking a bit jumbled.

Android/Firefox screenshot

And even on Windows/Chrome and Windows/Firefox, it doesn't quite "feel" centered, at least not at the start. Feels more left-aligned.

chrome-windows

> So this is the age old question, do ARIA roles and behaviors describe visually what a control looks like (e.g. a slider) or its functional properties (e.g. a radio group)

as an AT user I may possibly be expecting a slider to slide between different numerical values. I don't think it makes quite semantic sense if not associated with some form of value or scale of a kind. Maybe just having the fan speed slider, and removing the auto value, so it essentially does represent fan speed (from no speed to high speed)?

carmacleod:
@jongund In case it's helpful, I noticed the labels-vs-rail-width problem in the range slider as well: https://github.com/w3c/aria-practices/pull/1757#issuecomment-777205637

Also noticed that it is NOT a problem in the currently-released thermostat slider example: https://w3c.github.io/aria-practices/examples/slider/slider-2.html

Maybe there's a clue in there somewhere. :)

jongund:
@patrickhlauke 
That is a good idea about removing the "Heat/Cool" slider and changing the values of the "Fan" slider.
I will also update the "Fan" slider to use SVG text elements for the labels, that should fix some of the alignment issues in mixing HTML and SVG between browsers.
Thanks so much for your feedback.

jongund:
@patrickhlauke 
@carmacleod 

I have completed the updates:
1. Removed the Heat/Cool slider, since it is not a range.
1. Updated Fan slider to use off, low, medium and high values.
1. Changed button elements to SVG `role=button` for more consistent positioning across browsers and OSes.
1. Added the `meta` element,
 
Would appreciate your feedback on the changes whenever you get a chance.

patrickhlauke:
@jongund that's looking really good. thanks for taking on board the feedback about the heat/cool and fan slider - i think this now makes very nice logical sense. works great with iOS/VO.

Android/TalkBack in both Chrome and Firefox is ... weirdly broken, but in roughly the same way as before (announcing slider values in percent only, and for some reason in Chrome the Fan slider just can't be operated with double-tap-and-hold-and-then-drag type gesture - it goes into text selection instead; Firefox lets you operate both with this gesture, but announces nothing when things change). Don't think it's anything related to your approach here, but just overall lack of polish/support for this sort of thing in TalkBack...

patrickhlauke:
quick note on this

> Added the meta element,

while that works nicely, the big tables at the end of the page do then mess things up. In Chrome and Firefox on Android, the page is still zoomed out enough to accommodate the table widths, while on iOS/webkit it does zoom right to the width of the first part of the page, but then leads to horizontal scrolling to accommodate the tables. But I'm guessing it'd be non-trivial to do anything with the tables, other than perhaps work on wrapping them in a container that is only as wide as the viewport / `width:100%` and adding `overflow:scroll` perhaps

carmacleod:
In case it's helpful, apparently a new version of TalkBack was released 2 days ago (for Google Pixel and Samsung Galaxy devices).
- more multi-finger gestures
- merged menus into one context-sensitive menu 
- customization of gestures and menu
- new tutorial
https://blog.google/products/android/all-new-talkback/
https://www.youtube.com/watch?v=t-v15owYuwk

I don't have an Android device, but I am curious whether or not sliders feel more polished in the new TalkBack. :)

smhigley (approved):
The visuals look perfectly good to me, a very not-designer-y person 😄

patrickhlauke:
if we're ok with the new rewording from https://github.com/w3c/aria-practices/pull/1746#issuecomment-787437377 suggest adding it here as well

jongund:
@mcking65 
I tested on iPadOS and iOS and it works with Voice Over on Safari and Chrome, although for the temperature going one step (e.g. 0.1 degree) at a time is tedious.

The interesting accessibility test is when using a blue tooth keyboard with the device the cursor key commands for changing the slider value do not work.

Jon

carmacleod (approved):
This is an approving Accessibility Review for VO + Safari/Chrome on macOS

<kbd>Fn+up/down arrows</kbd> work for PageUp/PageDown in VoiceOver the same as with VO off (i.e. temperature goes up/down by 2 degrees instead of 1/10 degree).

Unfortunately, <kbd>Fn+left/right arrows</kbd> for Home/End work oddly in VoiceOver: they do change the slider value to min/max as expected, but then they move the VO cursor to unexpected locations at the beginning and end of the Thermostat group. This seems more likely to be a VO bug and not anything in the example code. Almost feels like VoiceOver's code is missing a "preventDefault" or "stopPropagation" or similar. :)

So, other than the weirdness with Home/End, this example works well with VoiceOver on macOS.

I noticed an editorial nit in the Keyboard Support table:
> Increases temperature slider value multiple steps. In this slider, jumps ten steps.

I think this should read:
> Increases temperature slider value multiple steps. In this slider, jumps twenty steps.

Because "one step" is a tenth of a degree, and PageUp/PageDown move 2 degrees.

carmacleod:
@jongund 
Try going to Settings > General > Keyboard > Hardware Keyboard and maybe changing to US keyboard instead of automatic.
Or check if anything looks amiss on that settings page.

If that doesn't help, maybe something in Setting > Accessibility > Keyboards > Full Keyboard Access > Commands ?

If that still doesn't help, note that I got quite a few hits when I googled:
> ipad bluetooth keyboard arrow keys don't work

The specific fix may depend on the exact keyboard or iOS/iPadOS version, etc. You could try adding your keyboard model to the search and see what google comes up with.

jongund:
@carmacleod 
Thanks for the tips, I was able to get the keyboard cursor keys working if I used a key labelled  "alt opt" with cursor keys.  So I guess I need to get a more standard keyboard.   I have a Logitech K480.

jongund:
@carmacleod 
I will update the documentation about the twenty steps.

examples/slider/slider-thermostat.html
```diff
@@ -0,0 +1,521 @@
...
+    <h1>Slider Examples with aria-orientation and aria-valuetext</h1>
+    <div class="advisement">
+      <p><strong>WARNING!</strong> Some users of touch-based assistive technologies may experience difficulty utilizing widgets that implement this slider pattern because the gestures their assistive technology provides for operating sliders may not yet generate the necessary output.  To change the slider value, touch-based assistive technologies need to respond to user gestures for incrementing and decrementing the value by synthesizing key events.  This is a new convention that may not be fully implemented by some assistive technologies.  Authors should fully test slider widgets using assistive technologies on devices where touch is a primary input mechanism before considering incorporation into production systems.
+    </p>
+    </div>
+    <p>
+      The following example of a hypothetical online thermostat demonstrates the
+      <a href="../../#slider">slider design pattern.</a>
+      The desired temperature is set with the first slider, which is vertically oriented.
+      Two horizontal sliders set fan speed and mode.
```
a11ydoer, line 38: Since the fan design changed only to "speed", all "mode" info can be deleted. (There are a few more places to delete the word, "mode".) Btw, it was great idea to change the design to current one, @jongund and @patrickhlauke!

a11ydoer (changes requested):

a11ydoer:
I still see the same mismatched "percentage value" problem from Talkback testing like in the color view slider. It may be unique to Talkback. I can file a bug to the Talkback team if necessary. 

1. Vertical thermostat - it was a struggle to manuver the thumb with touch(two finger gesture after thumb was activated with one tap). The selected temperature value, celsius, was read as percentage value which is confusing.
2. Fan - it relatively works well due to button role. Only problem is that "changed" fan speed was not read. In other words, fan speed change is only confirmed visually.

examples/slider/slider-thermostat.html
jongund: @a11ydoer 
@mcking65 
The change to the reference links and descriptions you are requesting is  something that needs to be coordinated with Matt, since there are several pull requests related to slider examples pending.  When examples are merged to main the reference links in the other examples and the guidance document should be updated to reflect the current examples.

a11ydoer:
@jongund @mcking65 @carmacleod 

Now I am about to look into code itself to figure out where the percentage info comes from -  why not actual thermorstat value nor correct value of color viewer slider. 

Upon your testing with iOS and more, does above thermostat and color viewer slider read correct value of celsius, not the percentage?

carmacleod:
@a11ydoer Did you mean to close this?

jongund:
@mcking65 
I have updated the example based on the discussion in today's meeting:
* Removed button roles and associated properties
* Updated documentation to remove button references and description.
* Updated links to example page and example description to include only temperature and fan speed.

charmarkk (changes requested):
Approving review for NVDA latest + JAWS latest + Chrome latest, working as expected and intended.

Disapproving/requesting changes based on HCM Black w/ defaults not working in Firefox or Edgeium. On Firefox, only some text shows up. On Edgeium, _nothing_ shows up. Screenshots to follow. HCM White w/ defaults works as expected.

Firefox HCM-B:
image


Edgium HCM-B:
image

charmarkk:
Based on Jon's most recent commits in 5ecf7c2, Looks like things are working as intended now with HCM Black (and White) in Firefox 86.0.1! But still broken in Edge 89.0.774.54 - only showing up as black when in HCM White.

image

JAWS-test:
> Looks like things are working as intended now with HCM Black (and White) in Firefox 86.0.1! But still broken in Edge 89.0.774.54 

HCM Black does not work in Chrome either. In Edge and Chrome, the slider is only visible when it receives focus. Otherwise it is displayed in black and is therefore invisible. The cause seems to be that the elements of the slider inherit the color per currentColor from the ancestor element, but in the SVG element itself the color black is hardcoded

jongund:
I have updated this example to dynamically set the `color` attribute on the SVG images when the page is loaded.
The `color` attribute value is defined by the current HTML text color of the `h3` heading.
The `currentColor` values used in the SVG graphics seem to be using the 'color' attribute set value.
This seems to work in setting the colors to the colors used in the windows HCM modes for both Firefox and Chrome.
The main issue is that if you change the HCM setting when the page is loaded it does not update the colors until you do a reload.
Not aware of a browser event that is fired on changing the HCM settings.

The changes to Chrome support for HCM effects a lot of other APG examples using SVG.

I looked at [media queries for light an dark](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) display modes, but not sure how HCM settings effect these queries and at best there would only be two color choices.   

Interested in the groups thought on this approach and other ideas on improvements, especially dynamically updating the colors when HCM values are changed when the page is already loaded.

JAWS-test:
> especially dynamically updating the colors when HCM values are changed when the page is already loaded.

I don't think that has to be fulfilled. It should be clear to all that first HCM is turned on and then the page is loaded

a11ydoer:
Thanks so much for your review, @JAWS-test !

jscholes:
@jongund Testing the preview link in the first comment with NVDA plus Chrome, there are serious problems with screen reader output:

1. When I tab to the first slider, I hear: "To get missing image descriptions, open the context menu.  graphic".  Then, the actual slider itself, "Temperature  slider  25.0 degrees Celsius".
2. When I tab to the second slider, I don't hear the prompt about image descriptions, but I do hear the word "graphic" before "Fan  slider  off".

I'm guessing, at a first glance, that this is because the ARIA roles and properties are added to an element inside a `<svg>`, and the SVG itself is being recognised as an unlabelled image.  If I add `role="presentation"` to each SVG, the problem goes away, but I leave it to others to determine whether that is appropriate for an APG example.  I would say that resolving this is critical before the updates can be merged.

In addition to the above, IMHO it's a little odd having all of the textual values for the "Fan" slider appear in the DOM after it.  E.g., with NVDA in browse mode, when pressing Down Arrow I hear:

> Fan
> slider    off
> Off
> Low
> Med
> High

The last four of these are interactive controls, I can click on them to set the value of the slider.  But they have no accessible role, nor state to indicate which is the currently set value.  As such, the example would fail [4.1.2 Name, Role, Value (level A)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value) and possibly [1.3.1 Info and Relationships (level A)](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships) if the `<g>` element representing the currently set value is visually distinguished in some way.

CC @mcking65 and @sinabahram

jscholes:
Amendment to my last comment: the quoted NVDA speech output was with `role="presentation"` added to the "Fan" slider's SVG.  Without that fix, the output is even worse:

> Temperature
> graphic    To get missing image descriptions, open the context menu.
> graphic    25.0°C
> graphic    slider    25.0 degrees Celsius
> Fan
> graphic    slider    off
> graphic    Off
> graphic    Low
> graphic    Med
> graphic    High

jongund:
@jscholes 
I have updated the slider thermostat example to use `role=presentation` on the SVG elements and `aria-hidden` on the fan labels.  This seems to fix the problem with NVDA reading extra information.

.vnurc
```diff
@@ -20,3 +20,5 @@ The “row” role is unnecessary for element “tr”.
 Attribute “aria-activedescendant” value should either refer to a descendant element, or should be accompanied by attribute “aria-owns”.
 # https://github.com/w3c/aria-practices/issues/1678
 Section lacks heading. Consider using “h2”-“h6” elements to add identifying headings to all sections.
+# https://github.com/validator/validator/issues/1096
+An “svg” element can have a “role” attribute.
```
nschonni, line 24: The supressions work by string matching, so it's the exact error message (or a regex)
```suggestion
Bad value “presentation” for attribute “role” on element “svg”.
```

a11ydoer: Thanks @nschonni

mcking65:
@jongund  wrote:
> @jscholes
> I have updated the slider thermostat example to use `role=presentation` on the SVG elements and `aria-hidden` on the fan labels. This seems to fix the problem with NVDA reading extra information.

@jscholes 
While I understand the desire not to have the clutter of the fan labels, I don't think hiding them is good practice. And, while they are clickable, I don't think that presenting them as static text to screen reader users violates anything in WCAG. It is no different from a clickable label element on a form field.

Why is using aria-hidden not good?
1. The text is visible on screen. We do not want to encourage hiding visible text.
2. Sighted screen reader users should be able to tap the label and get the same effect as anyone else.

jscholes:
@mcking65 

> It is no different from a clickable label element on a form field.

Generally speaking, clicking the label of a form field doesn't change that field's value (with the possible exceptions of a checkbox or radio button).  Focus is also placed on the field in question, so in the worst case scenario the state of a checkbox or radio will change but the user will be made aware of that change immediately and in context.  There is also a one-to-one relationship between labels and form fields, not four-to-one.

Here, the value of the slider does change, but the only thing that seems to happen is that the screen reader (NVDA at least) announces the new value without really enforcing the fact that it has changed.  The textual values are positioned after the element with `role="slider"` in the DOM, so it's not even likely that a screen reader user will discover the change during the natural progression of continuing to read the page.

Happy to agree to disagree, given that the main concern of the graphic announcement has been addressed.  But I maintain that having four controls on a page that change the value of another element, with those four controls not having accessible roles and states or being hidden, is a WCAG violation.  I also think that the clutter can be played down in this case because the slider only has four possible positions.  But if somebody implemented this same pattern for, say, air conditioner modes, you'd quite possibly have eight or more of these redundant elements.

jongund:
This is ready for review, made the following changes:
1. Setting the CSS property `forced-color-adjust` to `auto` on the SVG elements.
2. Using `stroke-opacity` and `fill-opacity` instead of `transparent` values for setting stroke and fill colors for the SVG rect used for focus indication.

jesdaigle (approved):
This is ready to merge @mcking65

jscholes:
@jnurthen mentioned that the hit area really needs increasing regardless of whether this sample is kept in its current form, replaced, updated, etc.

jongund:
@mcking65 
The changes requested at the meeting on 6 April 2021 are in pull #1857, so this pull request is not longer needed.

PR #1782, opened 2021-02-27
Infrastructure: chore: Enable and fix additional HTMLHint rules

nschonni:
Realized that HTMLHint behaves a little differently, and all rules are opt-in

***
<a href="https://pr-preview.s3.amazonaws.com/nschonni/aria-practices/pull/1782.html" title="Last updated on Oct 18, 2021, 1:27 AM UTC (4f4ac21)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1782/45e9a6c...nschonni:4f4ac21.html" title="Last updated on Oct 18, 2021, 1:27 AM UTC (4f4ac21)">Diff</a>

### DIFF
--- a/.htmlhintrc
+++ b/.htmlhintrc
@@ -1,3 +1,25 @@
 {
-  "attr-lowercase": ["viewBox"]
+  "alt-require": false,
+  "attr-lowercase": ["viewBox"],
+  "attr-no-duplication": true,
+  "attr-unsafe-chars": true,
+  "attr-value-double-quotes": false,
+  "attr-value-not-empty": false,
+  "doctype-first": true,
+  "doctype-html5": false,
+  "head-script-disabled": false,
+  "href-abs-or-rel": false,
+  "id-class-ad-disabled": false,
+  "id-class-value": false,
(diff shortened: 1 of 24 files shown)


nschonni:
> When does HTMLHint run? Is it part of the HTML linting github action? What is the effect of turning on these on? Will it cause auto-fixing going forward?

Yes, the current call to `npm run lint:html` calls both HTML Hint and the HTML Validator tasks. I don't know if there is auto-fixing though.

> Some of these issues are in generated files:

Good catch. I've now included the `*.template` files in this linting so it flags stuff there. The change in the `examples/index.html` appears to be something that should have been caught by the autogeneration but isn't. It was because I touched that example, and it now is adding it under a few places

It also looks like the `coverage/index.html` file should probably be autogened/checked by CI like the `example/index.html` file, but I'll spin up a separate issue for that

zcorpan (approved):

mcking65 (approved):
OK, looks good. I'm sure Michael will appreciate these improvements as we head to publication. I'll merge now.

PR #1830, opened 2021-03-24
Update Accordion example code and behavior

### LINKED ISSUE #1819: Revisit accordion behaviors on demo page
Thanks to an overly long conversation last week, it seems that the current behavior in the accordion example might be encouraging some less-than-ideal assumptions about how accordions should behave in general.

The particular conversation I had was around using arrow keys to cycle through headers -- the misconception was that since the APG example was coded that way, arrow keys were inherently more beneficial and "accessible." I actually think the reverse is true: arrow keys are pretty important for scrolling the page particularly in content-heavy sections, which covers most accordions including ours. Taking that default functionality away seems generally more harmful than not, especially since the accordion pattern itself is already allowing users to skip past sections of content.

I can see why it might be useful in some very specific cases -- maybe if accordions were used to show/hide groups of links or checkboxes within a nav or filtering region, the benefit of not needing to tab back through a large number of items would be helpful enough to be worth taking away scrolling. The issue is that this doesn't match our example, and also we don't have any documentation on when arrow keys would beneficial.

This is also related to #1719 -- I like the suggestion there to clarify that not all accordions need to enforce one-at-a-time open state within a set. We could even just not model that behavior, and have all sections expandable and collapsible, then add documentation about the option to add one-at-a-time behavior if desired. Anecdotally, I've found that forcing one accordion to remain open is more likely to be confusing than helpful, unless there's a specific reason to include it. Since our specific example includes no context that would make one-at-a-time beneficial, I'd be in favor of removing it and modelling only the more simple expand/collapse behavior.

An overall accordion update would also be a great opportunity to update the component and resolve #1183 😄

TL;DR:
Let's take a look at making the following updates to Accordion:
- Remove the arrow key behavior in our working example
- Add documentation to the Keyboard Interaction section on when the optional arrow keys + home/end would actually make sense to implement
- Consider removing the one-at-a-time behavior in the working example
- Add documentation on having one-at-a-time behavior as an option (if we remove it), or put more emphasis on it not being necessary (if we keep it)
- Generally update the code

smhigley:
Resolves #1819 with the following changes:

[Preview link](https://raw.githack.com/w3c/aria-practices/accordion-defaults/examples/accordion/accordion.html)

General cleanup:
- Updates JS file to a class syntax
- Update classnames to be lowercase
- Fix some border-radius issues in focus styling
- Updated example page wrapper `id` attributes to match other example pages

Behavior changes:
- Removed arrow key and force-one-open functionality
- Updated tests to reflect behavior change
- Removed section that talked about focus styling for "enhanced keyboard interaction"
- Removed arrow key/home/end rows in example page keyboard table
- Removed aria-disabled row in example page attributes table

### Review checklist

* [X] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @mcking65 
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @jongund 
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @jongund 
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @jongund

### DIFF
--- a/examples/accordion/accordion.html
+++ b/examples/accordion/accordion.html
@@ -29,167 +29,132 @@
   <p>
     The below example section contains a simple personal information input form divided into 3 sections
     that demonstrates the
-    <a href="../../#accordion">design pattern for accordion.</a>
-    In this implementation, one panel of the accordion is always expanded, and only one panel may
-    be expanded at a time.
+    <a href="../../#accordion">design pattern for accordion</a>.
   </p>
   <section>
     <div class="example-header">
       <h2 id="ex_label">Example</h2>
     </div>
     <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
-    <div id="coding-arena">
-      <div class="demo-block">
-        <!-- Accordion Configuration Options
-
-          data-allow-toggle
-            Allow for each toggle to both open and close its section. Makes it possible for all sections to be closed. Assumes only one section may be open.
-
-          data-allow-multiple
-            Allow for multiple accordion sections to be expanded at the same time. Assumes data-allow-toggle otherwise the toggle on open sections would be disabled.
-          __________
-
-          Ex:
-            <div id="accordionGroup" class="Accordion" data-allow-multiple>
-
-            <div id="accordionGroup" class="Accordion" data-allow-toggle>
-        -->
-        <div id="accordionGroup" class="Accordion">
-          <h3>
-            <button type="button" aria-expanded="true" class="Accordion-trigger"
-              aria-controls="sect1" id="accordion1id">
-              <span class="Accordion-title">
-                Personal Information
-                <span class="Accordion-icon"></span>
-              </span>
-            </button>
-          </h3>
-          <div id="sect1" role="region" aria-labelledby="accordion1id" class="Accordion-panel">
-            <div>
-              <!-- Variable content within section, may include any type of markup or interactive widgets. -->
-              <fieldset>
-                <p>
-                  <label for="cufc1">Name<span aria-hidden="true">*</span>:</label>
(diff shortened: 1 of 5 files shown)


carmacleod:
10-second review: this is so much nicer than what was there before! Arrow keys are not needed when any section can be collapsed.
Thanks, @smhigley !

I guess we need the whole review checklist thing.

smhigley:
Thanks for adding the preview link @carmacleod! Somehow I always forget that 😅

shirsha:
Thanks @smhigley for the changes.
Following are few suggestion. These were existing from previous code

- Move the "+" and "-" before the accordion name so that users with screen magnifiers can see state easily.
- Place the accordions in an unorder list so that it improves the experience for screen reader users.
- Can we add note saying having the content under each accordion inside a labelled region is a best practice? Having too many landmarks will not help screen reader users. I recommend removing the markup related to region landmark.

jongund:
@smhigley 
Same problem as the other accordion examples for VO on macOS/Safari, when navigating using VO next/prev item, the `h3` is announced, but not the `button` element information.

jongund:
@smhigley 
Works fine for VO/safari for iOS 14.4.

jongund (approved):

a11ydoer:
@mcking65 @shirsha @jesdaigle 

Can we have update on review so that we can see whether it can be merged? If not, we can add this PR to [next week meeting agenda](https://github.com/w3c/aria-practices/wiki/September-28,-2021-Agenda) since we will review another accordion PR 1834

mcking65 (approved):
Editorial review done; looks good!

SergeyAmshaev:
[]()10001

Issue #974, opened 2019-01-19
Accordion Example disabled heading focus behavior


mrcsmcln:
Should tabbing/using arrow keys skip over disabled headings? The example uses `aria-disabled` instead of the `disabled` attribute for the button, though it appears to me the two accomplish the same goal. I imagine for illustrative purposes, the former was used. Nevertheless, using the `disabled` attribute disallows focusing, so I'm wondering if the example should be updated for consistency?

mcking65:
We will discuss this when finalizing pr #1834

PR #1834, opened 2021-03-25
Draft: Two accordion examples, one basic and one of a group with only one open accordion

smhigley:
Related to #1830 

This is an example for the purpose of discussion: it demonstrates having two separate accordion examples; one is a basic set of independent accordions, and the other is a grouped accordion example with the restriction of only having one accordion open at a time.

This isn't ready to be merged or formally reviewed -- it's missing documentation updates and tests.

[Basic accordion preview link](https://raw.githack.com/w3c/aria-practices/accordion-step-example/examples/accordion/accordion.html)
[Group accordion preview link](https://raw.githack.com/w3c/aria-practices/accordion-step-example/examples/accordion/accordion-group.html)

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -184,7 +184,8 @@ Some JavaScript and CSS may not function correctly in Internet Explorer.
       </p>
       <section class="notoc">
         <h4>Example</h4>
-        <p><a href="examples/accordion/accordion.html">Accordion Example</a>: demonstrates a form divided into three sections using an accordion to show one section at a time. </p>
+        <p><a href="examples/accordion/accordion.html">Accordion Example</a>: demonstrates three individual accordions, each containing static content.</p>
+        <p><a href="examples/accordion/accordion-group.html">Accordion Group Example</a>: demonstrates a form divided into three sections using an accordion to show one section at a time.</p>
       </section>
       <section class="notoc">
         <h4>Keyboard Interaction</h4>
--- a/examples/accordion/accordion-group.html
+++ b/examples/accordion/accordion-group.html
@@ -1,0 +1,285 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+<meta content="width=device-width, initial-scale=1.0" name="viewport" />
+<title>Accordion Group Example | WAI-ARIA Authoring Practices 1.2</title>
+<!--  Core js and css shared by all examples; do not modify when using this template. -->
+<link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/base.css">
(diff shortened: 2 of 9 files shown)


JAWS-test:
It seems to me that the Basic Example has no difference to [Disclosure](https://www.w3.org/TR/wai-aria-practices-1.2/examples/disclosure/disclosure-faq.html). I wonder if we need the same example under two different terms?

shirsha:
Thanks @smhigley for the changes.
Following are few suggestion

- Move the "+" and "-" before the accordion name so that users with screen magnifiers can see state easily.
- Place the accordions in an unorder list so that it improves the experience for screen reader users. 
- Can we add note saying having the content under each accordion inside a labelled region is a best practice?  Having too many landmarks will not help screen reader users.

jongund:
@smhigley 
On macOS the button seems to be hidden inside the h3 when you use next or previous commands.  Basically it identifies the heading, but on the button.  Works fine it you use the TAB key.

shirsha:
I don't see aria-hidden attribute on Edge, Chrome browser (Windows OS). Jemma can see it in Mac OS.

jongund:
@smhigley 
Works fine on iOS 14.4.2, using VO with Safari.  But like VO for macOS it does not provide information about the `aria-disabled=true` information.

jongund:
@smhigley 
At the APG meeting today is was suggested to remove the `aria-required` and `aria-hidden` from the standard form controls, since they are not needed for the accordion pattern and increase the amount of documentation and testing for the example.

I think we also need to add more of a description of the accordion behavior in each example, indicating whether one will always be open or not.

jongund (approved):

a11ydoer:
@smhigley will complete testing code by early next week. Thanks so much, @smhigley

jongund:
@a11ydoer  @mcking65 @smhigley There needs to be documentation of the use of `aria-disabled` and it should also be in the original source code for the group example.

Might want to change:
"The form can only have one step open at a time"
to
"The form has one step open at a time"

JAWS-test:
@a11ydoer 

> For example, the accordion was expanded visually but the screenreader says it is "unavailable expanded".

This has to be that way because of aria-disabled=true, since it can only be opened but not closed. I don't see any error in this, but the match of screenreader output and visual representation.

a11ydoer:
@JAWS-test thanks for great feedback! It makes sense. It seems that we need to work on aria-disabled more.

smhigley:
@mcking65 I updated the roles, tests, and related example links. Let me know what you think of it now :)

shirsha (approved):
I will raise a separate issue for my suggestions.

jesdaigle (approved):

a11ydoer:
[issue 974](https://github.com/w3c/aria-practices/issues/974)

PR #1852, opened 2021-04-09
Add deprecation warning to Collapsible Listbox example

### LINKED ISSUE #657: Listbox: How to make a collapsible listbox required?
The collapsible listbox https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html is based on button. How can this be set to be required as button does not support aria-required. 
The best I can think of is to set the child listbox to required but this doesn't really satisfy the desire to mark the button itself as required.

smhigley:
Resolves #657

Includes the following two changes:
- Updates the link on the main aria-practices.html page to point to the select-only combobox with a text note
- Adds a deprecation warning to the listbox example page

[Preview link for example page](https://raw.githack.com/w3c/aria-practices/remove-listbox/examples/listbox/listbox-collapsible.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/1852.html" title="Last updated on Aug 30, 2021, 7:07 AM UTC (b41a23c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/1852/c8b1c8c...b41a23c.html" title="Last updated on Aug 30, 2021, 7:07 AM UTC (b41a23c)">Diff</a>

### DIFF
--- a/examples/listbox/listbox-collapsible.html
+++ b/examples/listbox/listbox-collapsible.html
@@ -28,6 +28,9 @@
   </nav>
   <main>
   <h1>Collapsible Dropdown Listbox Example</h1>
+  <div class="advisement">
+    <p><strong>DEPRECATION WARNING:</strong> this pattern has been deprecated, and will be removed in a future version of the Authoring Practices. The <a href="../combobox/combobox-select-only.html">select-only combobox</a> should be used as an alternative to this pattern.</p>
+  </div>
   <p>
     The following example implementation of the
     <a href="../../#Listbox">design pattern for listbox</a>
@@ -38,6 +41,7 @@
   </p>
   <p>Similar examples include:</p>
   <ul>
+    <li><a href="../combobox/combobox-select-only.html">Select-Only Combobox</a>: A single-select combobox with no text input that is functionally similar to an HTML <code>select</code> element.</li>
     <li><a href="listbox-scrollable.html">Scrollable Listbox Example</a>: Single-select listbox that scrolls to reveal more options, similar to HTML <code>select</code> with <code>size</code> attribute greater than one.</li>
     <li><a href="listbox-rearrangeable.html">Example Listboxes with Rearrangeable Options</a>: Examples of both single-select and multi-select listboxes with accompanying toolbars where options can be added, moved, and removed.</li>
     <li><a href="listbox-grouped.html">Listbox Example with Grouped Options</a>: Single-select listbox with grouped options, similar to an HTML <code>select</code> with <code>optgroup</code> children.</li>
--- a/examples/listbox/listbox-collapsible.html
+++ b/examples/listbox/listbox-collapsible.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <meta charset="utf-8" />
-<title>Collapsible Dropdown Listbox Example | WAI-ARIA Authoring Practices 1.2</title>
+<title>(Deprecated) Collapsible Dropdown Listbox Example | WAI-ARIA Authoring Practices 1.2</title>
 
 <!--  Core js and css shared by all examples; do not modify when using this template. -->
 <link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/base.css">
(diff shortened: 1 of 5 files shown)


examples/listbox/listbox-collapsible.html
```diff
@@ -28,6 +28,9 @@
   </nav>
   <main>
   <h1>Collapsible Dropdown Listbox Example</h1>
+  <div class="advisement">
+    <p><strong>DEPRECATION WARNING:</strong> this pattern has been deprecated, and will be removed in a future version of the Authoring Practices. The <a href="../combobox/combobox-select-only.html">select-only combobox</a> should be used as an alternative to this pattern.</p>
```
carmacleod, line 32: I would add ARIA (or WAI-ARIA) as follows:

```suggestion
    <p><strong>DEPRECATION WARNING:</strong> this pattern has been deprecated, and will be removed in a future version of the ARIA Authoring Practices. The <a href="../combobox/combobox-select-only.html">select-only combobox</a> should be used as an alternative to this pattern.</p>
```

carmacleod (approved):
Looks good!
Thanks, @smhigley !
Approving whether or not the little suggested change is included.

mcking65:
@smhigley 

The deprecation warning looks good.

I think we should also remove the link to the collapsable listbox example from:
* Scrollable Listbox Example
* Example Listboxes with Rearrangeable Options
* Listbox Example with Grouped Options

I'd like to discuss whether we note deprecation in the main document or remove the link. I'm leaning toward removing the link. We might also want to put something in the metadata, perhaps a data attribute on the H1, that would tell the indexing code to ignore it.

Also, what about adding "(Deprecated)" to the beginning of the page title tag and the start of the H1? Hmm, maybe that could be the indicator to the indexing script.

a11ydoer:
APG groupd made a decision to remove deprecated examples on May 4, 2021

a11ydoer:
@jongund will update APG example index page after @mcking65 merge this issue.

mcking65:
Note to self:
1. Prepend `(Deprecated) ` to title tag and H1
2. 2. Remove link from main doc and other example pages

jongund:
@smhigley @mcking65 @a11ydoer 
The only review this pull request needs is editorial on the deprecation warning and testing if the updated link works.  This should only require two maybe three people.  The list of reviewers seems to accessibility testing on all of our platforms.   I suggest updated the list of reviewers on this pull request.

jongund:
@smhigley 
The name of the example file needs to change from `listbox-collapsible.html` to `listbox-select-only.html` to match the change in `aria-practices.html`.

jongund (changes requested):
@smhigley 
I think this pull request should just delete the `listbox-collapsible.html` reference in `aria-practices.html` and not add the new reference to  `listbox-select-only.html`.

The pull request that creates `listbox-select-only.html` should make it's own changes to `aria-practices.html` and `listbox-collapsible.html` so the links actually work, otherwise as a reviewer it is confusing.

smhigley:
@jongund the added link is for the already-existing `combobox-select-only` example, and the links should work (or at least, they work fine for me in the preview). If everyone things it'd be better not to have any link there at all, I can remove it, but I think it makes sense to use this PR to decide what example links to include in the Listbox section.

jongund:
@smhigley 
The link was not working yesterday when I tested, it works today.

jongund (approved):

examples/listbox/listbox-grouped.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <meta charset="utf-8" />
-<title>Listbox Example with Grouped Options | WAI-ARIA Authoring Practices 1.2</title>
+<title>(Deprecated) Listbox Example with Grouped Options | WAI-ARIA Authoring Practices 1.2</title>
```
mcking65, line 5: Oops, right change in wrong file :) I'll push a fix for this.

examples/listbox/listbox-grouped.html
```diff
@@ -26,7 +26,7 @@
     </ul>
   </nav>
   <main>
-  <h1>Listbox Example with Grouped Options</h1>
+  <h1>(Deprecated) Listbox Example with Grouped Options</h1>
```
mcking65, line 29: I'll fix this one too

mcking65 (approved):
OK, Looks good now. Thank you @smhigley!

Issue #1680, opened 2020-12-15
treeview examples exclude aria-selected, which is required by ARIA 1.1


stevenyxu:
The ARIA 1.1 spec for treeitem requires `aria-selected` https://www.w3.org/TR/wai-aria-1.1/#treeitem.

`aria-selected` is absent on the following authoring practices examples:

https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-1/treeview-1a.html
https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2a.html
https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2b.html

This is impacting us because we are attempting to use the ARIA 1.1 tree/treeitem pattern per the spec and are running into AT support issues (e.g., JAWS not announcing selected state, VoiceOver wrongly announcing as a listbox), but lacking an APG reference, it's harder to report bugs.

PR #1869, opened 2021-04-23
File Viewer Tree Examples: Add aria-selected

kdoberst:
Addresses #1680.

* [Preview updated File Directory Treeview Example Using Computed Properties](https://raw.githack.com/kdoberst/aria-practices/Issue-1680-add-aria-selected-to-treeview-items/examples/treeview/treeview-1/treeview-1a.html)
* [Preview updated File Directory Treeview Example Using Declared Properties](https://raw.githack.com/kdoberst/aria-practices/Issue-1680-add-aria-selected-to-treeview-items/examples/treeview/treeview-1/treeview-1b.html)


a11ydoer:
@kdoberst Thanks for your PR as well as your patiece. 
We will discuss this PR at ARIA APG meeting on June 22 and update the issue.

examples/treeview/treeview-1/js/tree.js
```diff
@@ -93,9 +93,11 @@ Tree.prototype.setFocusToItem = function (treeitem) {
 
     if (ti === treeitem) {
       ti.domNode.tabIndex = 0;
+      ti.domNode.ariaSelected = true;
```
smhigley, line 96: To set the `aria-selected` attribute, you'd need to use `el.setAttribute('aria-selected', 'true')`. AOM attribute reflection isn't broadly supported, or stable.

The selected state should also likely be set to the actual selected item, and not the currently focused item. We should also probably look into adding a visual selected state, though we could conceivably add the attribute even without the visual state.

smhigley (changes requested):
Thanks for starting this PR! Let me know if you have questions on my feedback.

mcking65:
Added to [November 2, 2021 Agenda · w3c/aria-practices Wiki](https://github.com/w3c/aria-practices/wiki/November-2%2C-2021-Agenda)

a11ydoer:
I don't see any aria-selected attribute from current [treview example.](https://w3c.github.io/aria-practices/examples/treeview/treeview-1/treeview-1a.html)

zcorpan:
@a11ydoer right, this PRs adds it. But I guess it should also include `aria-selected="false"` in markup on each item, as it's a required state.

jongund:
@zcorpan The roles, properties and states table needs to be updated and also the regression tests for the additional `aria-selected` attribute to be documented and tested in the example.

a11ydoer:
@jongund I agree with you.
Can you point where the roles, properties and states table is? It has been a while to see that table and I will make sure whether we have updated every attribute in the table.

zcorpan:
The implementation at first was setting `aria-selected="true"` to the currently *focused* item. Since the example doesn't change which file is "selected" (as shown in the "File or Folder Selected" box) when moving focus, only when clicking or pressing enter, I changed the implementation to set `aria-selected="true"` in `onClick`.

examples/treeview/treeview-1/js/tree.js
```diff
@@ -87,6 +88,14 @@ Tree.prototype.init = function () {
   this.firstTreeitem.domNode.tabIndex = 0;
 };
 
+Tree.prototype.setSelectedToItem = function (treeitem) {
+  if (this.selectedItem) {
```
mzgoddard, line 92: What do you think of mirroring how `setFocusToItem` is implemented with something like:

```js
  for (let i = 0; i < this.treeitems.length; i++) {
    const ti = this.treeitems[i];
    if (ti === treeitem) {
      ti.domNode.setAttribute('aria-selected', 'true');
    } else {
      ti.domNode.removeAttribute('aria-selected');
    }
  }
```

test/tests/treeview_treeview-1a.js
```diff
@@ -80,6 +80,60 @@ const isClosedFolderTreeitem = async function (el) {
...
+  async (t) => {
+    const items = await t.context.queryElements(t, ex.treeitemSelector);
+
+    for (let i = 0; i < items.length - 1; i++) {
+      await items[i].sendKeys(Key.ARROW_DOWN);
+      const nextItem = await t.context.queryElement(
+        t,
+        ex.treeitemSelector + '[tabindex="0"]'
+      );
+      const focusMoved = items[i] !== nextItem;
```
mzgoddard, line 113: As an aside this will always be true. As I understand it, `nextItem` may represent the same element in the dom but webdriver's js api doesn't guarentee that it will be the same object in this context. To test that they are the same I think you'd use `WebElement.equals(items[i], nextItem)` or `(await items[i].getId()) !== (await nextItem.getId())`.

test/tests/treeview_treeview-1a.js
```diff
@@ -80,6 +80,60 @@ const isClosedFolderTreeitem = async function (el) {
...
+        await treeitem.getAttribute('aria-selected'),
+        'false',
+        'treeitem should initially have aria-selected="false"'
+      );
+    }
+  }
+);
+
+ariaTest(
+  'aria-selected attribute on treeitem on down arrow and enter',
```
mzgoddard, line 101: I think this test tries to test too many behaviors in a single test and that one of the behaviors should also be tested with `Key.SPACE`.

So maybe this should be tested with three tests? One to test that selection does not follow focus in response to `Key.DOWN_ARROW`. One to test that `Key.ENTER` selects on each treeitem. One to test that `Key.SPACE` selects on each treeitem.

The 'does not follow focus' test could be something like:

```js
ariaTest(
  'aria-selected attribute does not follow focus',
  exampleFile,
  'treeitem-aria-selected-not-focus',
  async (t) => {
    t.plan((ex.treeitemCount - 1) * 2);
    const items = await t.context.queryElements(t, ex.treeitemSelector);
    // Test all but the last treeitem. The last treeitem does not lose focus when `Key.DOWN_ARROW` is pressed.
    const itemsExcludeLast = items.slice(0, items.length - 1);

    for (let i = 0; i < itemsExcludeLast.length; i++) {
      const treeitem = itemsExcludeLast[i];

      await treeitem.sendKeys(Key.ENTER);

      await treeitem.sendKeys(Key.ARROW_DOWN);
      const focusedItem = await t.context.queryElement(
        t,
        ex.treeitemSelector + '[tabindex="0"]'
      );

      t.is(
        await treeitem.getAttribute('aria-selected'),
        'true',
        'selected treeitem does not change after sending ARROW DOWN'
      );
      t.is(
        await focusedItem.getAttribute('aria-selected'),
        'false',
        'focused treeitem does not have aria-selected="true"'
      );
    }
  }
);
```

mzgoddard (commented):
Reading how `tree.js` and `treeitem.js` behave this example implements single selection and not multi selection. Is that right?

If that is right does the following apply to this example?

https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-23

> - When a single-select tree receives focus:
>   - If none of the nodes are selected before the tree receives focus, focus is set on the first node.
>   - If a node is selected before the tree receives focus, focus is set on the selected node.


https://www.w3.org/TR/wai-aria-practices-1.2/#tree_roles_states_props

> - If the tree does not support multiple selection, aria-selected is set to true for the selected node and it is not present on any other node in the tree.
> - if the tree supports multiple selection:
>   - All selected nodes have aria-selected set to true.
>   - All nodes that are selectable but not selected have aria-selected set to false.
>   - If the tree contains nodes that are not selectable, those nodes do not have the aria-selected state.

If that applies, I think a few changes need to be made:

- No treeitem in the modified example html should have the attribute `aria-selected="false"`
- Add event handling to select the first node when no node is selected
- Add event handling to focus the selected node when tree receives focus from outside the tree
- Regression test that clicking selects an item
- Regression test that Key.SPACE selects an item (ENTER is already covered in this PR)

mcking65:
@mzgoddard 

>If the tree does not support multiple selection, aria-selected is set to true for the selected node and it is not present on any other node in the tree.

With the changes made in ARIA 1.3 and already implemented in browsers, that statement in the pattern will need to change. I'll create a separate PR for that. If selection does not follow focus, like in this particular tree, then it is necessary to specify aria-selected on all selectable nodes. Otherwise, the browser might assume that selection follows focus and provide an implicit selected state to the focused treeitem.

mcking65:
@mzgoddard pointed out this aspect of the pattern:

>* When a single-select tree receives focus: 
>   * If none of the nodes are selected before the tree receives focus, focus is set on the first node. 
>   * If a node is selected before the tree receives focus, focus is set on the selected node. 

This example is not putting tabindex 0 on the item that is selected by pressing enter.

In the navigation tree, we not only ensure the current item is kept in the tab sequence but also ensure that it is visible after blur.

mcking65:
The example might be more clear if:

1. The label on the tree were "Select folder or file to view". Currently, the tree is labeled "File viewer", but the tree is not a viewer; it is a chooser.
2. The label on the read-only edit field were "Selected folder or file:" Putting the word "selected" first may improve clarity.

mcking65:
I noticed that enter both expands and selects a parent node. Shouldn't Enter only select?

mcking65:
I've started drafting updated selection guidance for the tree pattern in #2121.

I still have to add a note advising against using both selected and checked in the same tree and a note about when browsers imply selection. I'm hoping I can find time to finish it later tonight.

a11ydoer:
We will change example name will be "My file", not "File viewer" from APG meeting on Nov 9, 2021

mcking65:
@howard-e  The functional and label changes are perfect, thank you!

@jongund if there is anything else that should be added to the accessibility features, especially WRT HCM, could you please add a commit with proposed wording?

@smhigley  could you please re-review, and if your original feedback has been addressed, change your review status to approve?

jongund (changes requested):
Under keyboard support:

* For `enter` and `space` I suggest adding that for menu items that are folders default action also includes opening and closing folders.
* I suggest phrases "opens the node" should change to use "open the folder".

mcking65:
@jongund wrote:

>For enter  and space  I suggest adding that for menu items that are folders default action also includes opening and closing folders. 

It doesn't. We made the default action consistent across all nodes because selecting a parent item is possible in this tree.

My primary question is whether we need more accessibility features documentation related to HCM.

jongund:
@mcking65 Unfortunately when these examples where created we were not considering high contrast support, so there is no support right now.   I am interested in redoing these examples at some point to bring them up to our current coding, visual focus and HCM support.   Maybe also making it more realistic, for example a combobox or dialog.

mcking65:
@jongund wrote:

> @mcking65 Unfortunately when these examples where created we were not considering high contrast support, so there is no support right now. I am interested in redoing these examples at some point to bring them up to our current coding, visual focus and HCM support.

OK, thank you. This PR is at least addressing the selection issue and making it consistent with the spec and pattern.

I will finish fixing up the documentation and ask for re-review before merge.

Looks like we have CSS lint errors that will need to be ignored. Fixing them can wait until we fix up the rest of the code in a later version.

mcking65:
@howard-e 

FYI ... I was mistaken when I said the label change was good. I didn't actually look at this until I was editing other parts of the file:

```
<h3 id="tree1" aria-label="My Documents">File Viewer</h3>
```

Except in some very rare circumstances, none of which come to mind at the moment, it is not good practice to put aria-label on a heading. Headings get their name from content, and the aria-label overrides the name from content. Thus, in this case, screen reader users will think the heading is "My Documents", but the words on screen are "File Viewer".

I was imagining that we would keep the "File Viewer" heading as a heading on the entire example (tree + edit field). And, I thought @smhigley had suggested adding a label of "My Documents" for just the tree. Instead, I just changed the heading to "My Documents" and removed the aria-label. This may have been what @smhigley was suggesting anyway. I may have initially misinterpreted her suggestion.

BTW, for a more extensive description about name from content, which types of elements are named from content, and the dangers of aria-label, have a look at [Section 5.3.2.1: Naming with Child Content](https://w3c.github.io/aria-practices/#naming_with_child_content).

mcking65:
@jongund or @smhigley 
I think I have this in a good place to ship. Would be nice if either of you could take a look and change to approving review.

mcking65 (approved):
Finished editorial changes; this is approving editorial review.

jongund (changes requested):
1. The icons used to represent the open and closed folders and a document in the declared properties example are broken, the work in the computed example.
2. The documentation for the `space` key opening and closing a file folder is not documented in either example, and the APG guidance section only discusses the `space` key in the context of multi-selectable trees.  It is also curious to note the behavior of the `space` and `enter` keys is different when focus is on a folder node, `space` opens and closes, `enter` does not.

jongund:
@mcking65 @howard-e I added a review, a few more things I think need to be addressed.

mcking65:
@jongund  wrote:

>1. The icons used to represent the open and closed folders and a document in the declared properties example are broken, they work in the computed example.

@howard-e  can you fix this this morning?

>2. The documentation for the `space` key opening and closing a file folder is not documented in either example, and the APG guidance section only discusses the `space` key in the context of multi-selectable trees.  It is also curious to note the behavior of the `space` and `enter` keys is different when focus is on a folder node, `space` opens and closes, `enter` does not.

Behavior of space should be identical to Enter, which is that it should only select -- perform the default action.

@howard-e can you change space key behavior so it does not open and close parent nodes? It should perform selection only, just like Enter.

howard-e:
Sure, I'll take a look @mcking65

jongund (approved):

jongund:
@mcking65 @howard-e 
Approved this pull request

howard-e:
> Approved this pull request

Thanks @jongund.

> can you change space key behavior so it does not open and close parent nodes? It should perform selection only, just like Enter

Also just pushed the final change to account for @mcking65's request and updated the relevant regression tests.

test/tests/treeview_treeview-1a.js
```diff
@@ -80,6 +80,66 @@ const isClosedFolderTreeitem = async function (el) {
...
+        'treeitem should have aria-selected="false" after expanding by sending key ARROW_RIGHT'
+      );
+      await items[i].sendKeys(Key.ENTER);
+      t.is(
+        await items[i].getAttribute('aria-selected'),
+        'true',
+        'treeitem should have aria-selected="true" after selecting by sending key ENTER'
+      );
+      // move focus to the next item
+      await items[i].sendKeys(Key.ARROW_DOWN);
```
smhigley, line 136: Why this line?

smhigley (approved):

Issue #1163, opened 2019-09-03
Avoid language changes in alert example


JAWS-test:
https://w3c.github.io/aria-practices/examples/alert/alert.html

The language change within live regions is generally or at least partially ignored by assistive technology. Therefore it should be avoided. The language change in the alert example suggests to web developers that this is an acceptable practice. 

I suggest to use a pure English text without language change in the alert example.

See: https://github.com/FreedomScientific/VFO-standards-support/issues/8, https://github.com/w3c/aria-practices/issues/1162,

ZoeBijl:
Tested with VoiceOver and JAWS both of which handle this example just fine. If they’re not speaking languages it means that the speech synthesizers for those languages aren’t installed on your machine. If you find issues with other screen readers please re-open this issue with more details.

JAWS-test:
For JAWS it is not true: https://github.com/FreedomScientific/VFO-standards-support/issues/8#issuecomment-527444298.  The language is output correctly while reading, but never in the live region. This is a well-known bug in JAWS

JAWS-test:
To make matters worse, JAWS cannot output the Japanese and Korean characters at all. Instead, question marks are output with a comma between them. So this is really not a meaningful output. And JAWS doesn't know the language ko and only outputs "ko". Can you please reopen the ticket?

![alert](https://user-images.githubusercontent.com/52748763/65142917-08011300-da14-11e9-826f-962fab61d292.jpg)

patrickhlauke:
do you have the relevant language packs installed in windows?

JAWS-test:
@patrickhlauke: As I already wrote, the output is correct if the message is read with the arrow keys. The output is only wrong within the live region. And yes, this is a JAWS bug that has existed for years and is documented: https://github.com/FreedomScientific/VFO-standards-support/issues/8

patrickhlauke:
well then it's a jaws bug. not a fault of aria practices...

JAWS-test:
I never doubted it was a JAWS bug. I only think it would make sense if the examples in APG were designed to avoid as many mistakes as possible. I'm also wondering which screenreader simultaneously outputs the Latin, Japanese, and Korean characters correctly?

ZoeBijl:
It’s correct that neither VoiceOver nor JAWS read all the languages in the live region. However, as Patrick said, this is not an ARIA Authoring Practices error.

None of the code in this example is invalid.

JAWS-test:
[...]

PR #1940, opened 2021-06-09
Alert example: remove non-English text from alert content"

jscholes:
While preparing tests for this example within the ARIA AT project, we determined that testing the different languages used within the alert text would be impossible because it can't be guaranteed that users will have speech engines installed that can voice all of them.  During discussion of how to proceed, the community group couldn't think of a real-world case where having an alert fire in multiple languages at once (rather than a single language expected by the user) would be appropriate or helpful, and so it was proposed that the non-English text just be removed.  This PR accomplishes that task, and capitalises the "H" in "Hello" given its new position at the beginning of the alert text.

### DIFF
--- a/examples/alert/alert.html
+++ b/examples/alert/alert.html
@@ -50,7 +50,7 @@
 
         <!--  The following script element contains the content that will be inserted into the alert element. -->
         <script type="text/template" id="alert-template">
-          <p><span lang="da">Hej</span>, hello, <span lang="it">ciao</span>, <span lang="ja">こんにちは</span>, <span lang="ko">안녕</span></p>
+          <p>Hello</p>
         </script>
       </div>
       <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>


jscholes:
CC @mcking65

JAWS-test:
Related: https://github.com/w3c/aria-practices/issues/1163

mcking65 (approved):
Perfect, @jscholes  sorry for very tardy review!

ZoeBijl:
> we determined that testing the different languages used within the alert text would be impossible

@jscholes what exactly was impossible?

jscholes:
@ZoeBijl Before this change, the alert example included text which, depending on a user's configuration and/or screen reader, would not be spoken correctly or at all, because of the necessary speech engines not being installed/available.  This significantly hampered the creation of ARIA-AT tests, which are executed by people running screen readers with their default configurations, and rely on single points of behaviour that can be targeted in assertions.

In short, a user would have no way of knowing whether certain screen reader behaviour was a failure of the overall software, or just in their setup.  Similarly, it would be desirable, with multilingual content, to include assertions relating to the automatic switching of language by the speech engine.  However, this is difficult, if not impossible, to convey in a textual assertion.  How, for instance, can we ask a person to unambiguously state whether a particular word was spoken using French pronunciation rules?  Particularly if they are not a French speaker.

These can be seen as ARIA-AT project limitations, which indeed may be resolved one day (for example via automated test runs that can query the programmatic state of the screen reader under test).  For now, there was consensus among the ARIA-AT Community Group that testing the support for alert semantics was of primary importance, given how often alerts are used across the web, and hence this change was implemented to facilitate such testing.

And finally, while I recognise that this next point is subjective, it was also believed that multilingual alerts are legitimately an edge case on the real-world web.  A website/app should either provide an option to set the user's preferred language, and then fire alerts using that choice.  Or, if giving users information in multiple languages at once is a requirement for some reason, an alert is probably not the best idea, e.g. because the text could be quite long, leaving users at the mercy of a random developer who has decided which one(s) should be spoken first.

ZoeBijl:
> This significantly hampered the creation of ARIA-AT tests, which are executed by people running screen readers with their default configurations, and rely on single points of behaviour that can be targeted in assertions.

Ah that makes a lot of sense. Thank you so much for the elaboration, much appreciated!

PR #1976, opened 2021-07-15
Infrastructure: Restrict coverage report to issue-write

nschonni:
Use new syntax to restrict token scope so the job can be re-enabled https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -7,6 +7,10 @@ on:
     - "test/**"
     - "!examples/landmarks/**"
 
+permissions:
+  contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage-report.yml
```diff
@@ -7,6 +7,9 @@ on:
     - "test/**"
     - "!examples/landmarks/**"
 
+permissions:
+  issues: write
```
nschonni, line 11: The docs are a little light, so I'm not sure if it also needs `content: read` for the clone operation

PR #2032, opened 2021-08-28
Multiple examples: Fix comments to satisfy JSDoc lint warnings

nschonni:
Autofixed with `npm run lint:es -- --fix`

### DIFF
--- a/examples/combobox/js/grid-combo.js
+++ b/examples/combobox/js/grid-combo.js
@@ -9,11 +9,9 @@ var aria = aria || {};
 
 /**
  * @class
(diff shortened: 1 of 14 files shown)


mcking65:
Thank you @nschonni. 

@jongund @smhigley @jesdaigle 

Looks like we have a rule that doesn't prefer blank lines in comments between parm descriptions. Do we care about this? Is this something we want to enforce? If so, should we look at the warnings from the linter during code review? If not, should we disable the warning? Does this deserve time on the meeting agenda?

PR #2033, opened 2021-09-02
Move husky install to prepare script

WilcoFiers:
It is no longer recommended to use "postinstall" for Husky ([see docs](https://typicode.github.io/husky/#/?id=install)), and to use "prepare" instead. `postinstall` scripts can not be guaranteed to run, and make it impossible to aria-practices into another project. The following will throw an error, because the postinstall is used rather than prepare:

```sh
$ npm install -D w3c/aria-practices#main
> aria-practices@0.0.0 postinstall /Users/wilco/projects/axe-core/node_modules/aria-practices
> husky install

sh: husky: command not found
```


WilcoFiers:
@jesdaigle Any idea when this could go in?

PR #2046, opened 2021-09-27
Add Sortable table example

jongund:
I needed an example of a sortable table for someone at Illinois and thought I would propose it as an example for the APG.

[Preview Link](https://raw.githack.com/w3c/aria-practices/sortable-table/examples/table/sortable-table.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/2046.html" title="Last updated on Oct 31, 2021, 5:26 PM UTC (511c27d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/2046/7e997a0...511c27d.html" title="Last updated on Oct 31, 2021, 5:26 PM UTC (511c27d)">Diff</a>


mcking65:
As a screen reader user, I find the button descriptions make the state of the table harder to understand. You have to listen very closely or you might think that the columns are are already sorted in the way the button description describes. I am not convinced we need descriptions. I wonder what other screen reader users think?

If there were descriptions, I'd rather have it be clearly distinguishable from sort state, perhaps "Sorts by COLUMN_NAME" and leave out the directionality of the action. The sort direction is conveyed after activation.

examples/table/sortable-table.html
```diff
@@ -0,0 +1,227 @@
...
+      <label>
+        <input type="checkbox"
+          value="show-unsorted-icon"
+          aria-describedby="show-unsorted-icon-desc">
+        Display sort icon in headers of columns that are not sorted
+      </label>
+      <p id="show-unsorted-icon-desc">
+        Adds an <q>♢</q> icon to the header of each column that is sortable but is not currently sorted.
+        Some implementations do this to help users distinguish sortable columns from columns that cannot be sorted.
+        It is important that a different shaped icon is used for users with visual impairments to perceive the difference with the sorted column icon in high contrast modes.
```
mcking65, line 55: What does this mean?  A "different shaped icon". If the icon were the same as the up or down icons, then the icon would be meaningless to all users. I'm not sure whether we should clarify or delete this setence.

examples/table/sortable-table.html
```diff
@@ -0,0 +1,227 @@
...
+                <button>
+                  Favorite Number
+                  <span aria-hidden="true"></span>
+                </button>
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td role="cell">Fred</td>
```
mcking65, line 102: Why declare the role? It is implicit. The spec says not to do this. Pretty certain the role declarations should be removed.

examples/table/sortable-table.html
```diff
@@ -0,0 +1,227 @@
...
+        </table>
+      </div>
+      <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
+    </section>
+
+    <section>
+      <h2>Accessibility Features</h2>
+      <ul>
+        <li>
+          HTML table markup is used to create the rows, caption, and to identify the column headers for the data cells.  The use of HTML supports the ARIA principle of using native element semantics whenever possible.<br/>
```
mcking65, line 139: OK with removing this? It is a coding practice, not accessibility feature. As we add more HTML examples, this language will become superfluous.

examples/table/sortable-table.html
```diff
@@ -0,0 +1,227 @@
...
+          <ul>
+            <li>To make it easier to perceive when a button has focus, the focus indicator encompasses both the column label and sort direction icon.</li>
+            <li>The cursor is changed to a pointer when hovering over the button to help people identify it as an interactive element.</li>
+            <li>To make it easier to perceive that clicking either the column label or the sort direction icon will sort the table, hover styles the button and icon in the same way that focus does.</li>
+            <li>To make activating sort easier for people with visual and movement impairments who are using a pointing device, the click target is maximized not only by making both the column label and sort icon clickable but also by using CSS positioning and sizing to make the button fill the entire header cell area.</li>
+          </ul>
+        </li>
+        <li>To ensure the sorting direction icon is synchronized with the value of the <code>aria-sort</code> attribute, CSS selectors based on the <code>aria-sort</code> value are used to render the sort direction icon.</li>
```
mcking65, line 154: OK with removing this? This is a coding practice, not an accessibility feature. If it were to be documented anywhere, it should be in a comment in the code.

mcking65 (changes requested):

examples/table/sortable-table.html
jongund: @mcking65 I made some edits to the example option description for you to review.

examples/table/sortable-table.html
jongund: @mcking65 removed the role="cell": from the TD elements of the sortable table example

examples/table/sortable-table.html
jongund: @mcking65 I have removed the reference to using `aria-sort` for controlling the rendering of the visual sort icon

jongund:
@howard-e I have addressed @mcking65 comments

examples/table/sortable-table.html
mcking65:
```
          HTML table markup is used to create the rows, caption, and to identify the column headers for the data cells.  The use of HTML supports the ARIA principle of using native element semantics whenever possible.<br/>
```

I don't think we need to say this at all. As we add more and more HTML examples, saying we are using more HTML and less ARIA will be increasingly common and saying this over and over won't add value. Would you be OK with removing thisitem from the list?

examples/table/sortable-table.html
```diff
@@ -0,0 +1,225 @@
...
+                <li>Set on the currently sorted column.  When the sorted column is changed, the <code>aria-sort</code> attribute is removed and set on the newly sorted column.</li>
+                <li>A value of &quot;<code>ascending</code>&quot; indicates the data cells in the column are sorted in ascending order.</li>
+                <li>A value of &quot;<code>descending</code>&quot; indicates the data cells in the column are sorted in descending order.</li>
+              </ul>
+            </td>
+          </tr>
+          <tr data-test-id="span-aria-hidden">
+            <td></td>
+            <th scope="row"><code>aria-hidden="true"</code></th>
+            <td><code>svg</code></td>
```
mcking65, line 189: Shouldn't this be span instead of svg?

mcking65 (changes requested):
I think we are very close to ready to ship. couple of questions left.

jongund:
@mcking65 I made update to the example option description based on your feedback, removed the section on using HTML table markup and changes `svg` to `span` for the `aria-hidden` entry in the roles, properties and states table.

jongund:
@nschonni HTML linting problem with files unrelated to this pull request.  Not sure how to "fix" this issues, should probably be a separate pull request and not fixed in this one.

nschonni:
Yes, it looks like the issues are related to https://github.com/w3c/aria-practices/pull/2055 landing, not this one

nschonni:
@jongund the linting issue is now fixed and green here again

mcking65 (approved):

PR #1277, opened 2019-12-02
Communicating Widget States

zcorpan:
Replaces https://github.com/w3c/aria-practices/pull/1057

PR #2071, opened 2021-10-27
Add guidance section about Communicating Widget States

zcorpan:
Replaces #1277

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/2071.html" title="Last updated on Oct 29, 2021, 12:37 PM UTC (a3e9f6d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/2071/0972a83...a3e9f6d.html" title="Last updated on Oct 29, 2021, 12:37 PM UTC (a3e9f6d)">Diff</a>

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -6250,14 +6250,7 @@ So, As advised by <a href="#naming_rule_avoid_fallback">Rule 4: Avoid Browser Fa
           In most cases, using only <code>aria-selected</code> is the right choice.
           Do not add both <code>aria-selected</code> and <code>aria-checked</code> "just in case",
           since that would increase verbosity for assistive technology users and could cause unnecessary confusion.
-        </p>
-
-        <p>
-          In rare cases, a widget could support both selection and checkedness to represent orthogonal states,
-          although that could be complex and might be confusing for users.
-          For example, a <code>tree</code> widget to represent a changeable sequence of steps
-          could allow selection in order to rearrange the steps,
-          and also allow checking and unchecking them to indicate which steps are active.
+          A valid use of <code>aria-checked</code> over <code>aria-selected</code> is when you need to support a <code>mixed</code> selection state.
         </p>
 
         <p>


mcking65:
There has been a lot of confusion about several of these properties, especially checked and selected. ARIA issues on this topic are some of the most commented. This is one of those topics where the spec language and its implications are particularly challenging for most people to groc, so it is important for APG to make it concrete and understandable. To do that, we need to add some more guidance with specific examples explaining the nuances of the spec.

Even though some of the spec language that supports greater clarity didn't make it into 1.2 and is only now in the editor's draft for 1.3, the updates have already been made to browsers. So, our first release of this section should reflect that reality.

In particular, we need to include thorough explanations of the language regarding checked and selected in [option](https://w3c.github.io/aria/#option) and [treeitem](https://w3c.github.io/aria/#treeitem) added in [ARIA PR 1340](https://github.com/w3c/aria/pull/1340/files). The history and comments in that PR are helpful context.

One of the concepts we considered working into the spec as a requirement but instead left as a non-normative sentence is the preference for checked in multi-selects and selected in single selects. We decided that it is more appropriate at this time for the APG to provide more definitive guidance.

Another topic that needs to be clarified in this section or referenced where appropriate is the difference between focus and selection and the meaning of selection following focus. We might be able to do this more effectively if we first update the [keyboard design section on selection and focus](https://w3c.github.io/aria-practices/#kbd_selection_follows_focus). I'm not sure where the best place to define the difference is; it might be this section. However, the keyboard section currently does not reference any of the completed examples that demonstrate the difference well, e.g., tabs. I think it was written before they existed.

zcorpan:
@mcking65 that's a lot of context... Can you suggest the text you'd like to see added or changed?

Do we need to fix the focus/selection clarification in this PR?

@a11ydoer can we add this to the agenda for today?

zcorpan:
@mcking65 said in today's meeting:

- shouldn't say that aria-selected is required
- should clarify focus vs selection
- should link to selection follows focus

a11ydoer:
general idea - adding section about "selected" and connecting with keyboard section regarding focus management

PR #2081, opened 2021-10-31
Infrastructure: Move Depednabot updates weekly

nschonni:
Removes the limit of 10, so some safe dependencies get PRs and
don't get stuck behind a list of 10 that might be harder to land.
Changes to a weekly schedule to at least there should be less
churn with the removed cap.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,9 +3,9 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: daily
+    interval: weekly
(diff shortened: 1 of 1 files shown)


nschonni:
Miss-read the docs originally https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#open-pull-requests-limit so changed it to `99` rather than removing the line

zcorpan (approved):

PR #2083, opened 2021-10-31
Infrastructure: vnu-jar problem matcher update

nschonni:
Currently the "info" at the start breaks the problem matcher, so these
aren't shown on PRs

### DIFF
--- a/.github/workflows/vnu-jar-problem-matcher.json
+++ b/.github/workflows/vnu-jar-problem-matcher.json
@@ -4,12 +4,12 @@
       "owner": "vnu-jar",
       "pattern": [
         {
-          "regexp": "^\"file:(.*)\":(\\d+).(\\d+)-\\d+\\.\\d+:\\s(error|warning):\\s(.*)$",
+          "regexp": "^\"file:(.*)\":(\\d+).(\\d+)-\\d+\\.\\d+:\\s(info )?(error|warning):\\s(.*)$",
           "file": 1,
           "line": 2,
           "column": 3,
-          "severity": 4,
-          "message": 5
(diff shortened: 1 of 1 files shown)


nschonni:
You can see the few info/warnings on the files tab now

.github/workflows/vnu-jar-problem-matcher.json
```diff
@@ -4,12 +4,12 @@
       "owner": "vnu-jar",
       "pattern": [
         {
-          "regexp": "^\"file:(.*)\":(\\d+).(\\d+)-\\d+\\.\\d+:\\s(error|warning):\\s(.*)$",
+          "regexp": "^\"file:(.*)\":(\\d+).(\\d+)-\\d+\\.\\d+:\\s(info )?(error|warning):\\s(.*)$",
```
zcorpan, line 7: Is it possible to use non-capturing `(?:info )` instead?

nschonni: No, because it only really recognized the Warning/Error levels for the metadata you see on the files tab

zcorpan: I don't understand. Is the `?:` regexp syntax not supported?

nschonni: Sorry, I thought you meant capturing the "info" instead of the warning. Not sure about the updated capture group, but I could try later

nschonni: Looks like it works!

zcorpan (approved):

PR #2116, opened 2021-11-05
Fix a few minor accessibility issues

WilcoFiers:
- The feedDisplay page used an h2 where it should have had an h1

- The bootstrap-accessibility.js file was putting role="presentation" on li element in the navigation for absolutely no reason. This causes the entire menu to end up as `nav > ul > li[role=presentation]`. This seemed unintentional. The fix was just to remove that `.navpills` class, which doesn't seem to be used in styles.

- The "Notes" heading in listbox pages skip a heading level. Going from h2 to h4. Doesn't seem necessary.

### DIFF
--- a/examples/feed/feedDisplay.html
+++ b/examples/feed/feedDisplay.html
@@ -16,7 +16,7 @@
 </head>
 <body>
   <main>
-    <h2>Recommended Restaurants</h2>
+    <h1>Recommended Restaurants</h1>
     <section id="main-content">
       <div id="restaurant-feed" role="feed">
       </div>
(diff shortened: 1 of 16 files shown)


mcking65:
@WilcoFiers 

Don't mean to be a pain, but wonder if you'd mind separating the landmark change into a separate PR? That will enable me to make a more clear change log.

BTW, collapsable listbox is deprecated pattern, and that page doesn't fully follow the template for examples. That notes section doesn't really belong there, but we haven't priritized fixing it due to the deprecation.

PR #2118, opened 2021-11-08
Landmark Example Pages: Restore list structure in page navigation region

WilcoFiers:
The bootstrap-accessibility.js file was putting role="presentation" on li element in the navigation for absolutely no reason. This causes the entire menu to end up as nav > ul > li[role=presentation]. This seemed unintentional. The fix is just to remove that .navpills class, which doesn't seem to be used in styles.

### DIFF
--- a/examples/landmarks/HTML5.html
+++ b/examples/landmarks/HTML5.html
@@ -26,7 +26,7 @@
       <div class="row">
         <div class="col-xs-12 col-sm-3 col-md-3 col-lg-2">
           <nav>
-            <ul class="nav nav-pills nav-stacked">
+            <ul class="nav nav-stacked">
               <li><a href="index.html">Principles</a></li>
               <li class="active"><a href="HTML5.html" aria-current="page">HTML</a></li>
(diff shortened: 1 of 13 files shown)


mcking65 (approved):
Thank you @WilcoFiers! LGTM, but since changes CSS, would like @jongund  to review to ensure there are no side effects I can't see.

jongund (changes requested):
@WilcoFiers @mcking65 Thank you for your interest in the landmark example, it seems the change breaks the `aria-current="page"` visual styling for the example. 

[Preview Link I used](https://raw.githack.com/WilcoFiers/aria-practices/landmark-a11y-fixes/examples/landmarks/banner.html)

ZoeBijl:
I’ve added a fix to @WilcoFiers’ branch.

ZoeBijl:
@jongund I’m seeing the correct style on that link you used, maybe a cache issue?

the banner link highlighted in the menu

jongund (approved):

PR #2119, opened 2021-11-08
Listbox Examples: Correct skipped heading level issue; change some h4 to h3.

WilcoFiers:
Use h3 for these notes, so that the page doesn't skip from an h2 to an h4 heading.


mcking65 (approved):
Thank you @WilcoFiers! LGTM.

PR #2120, opened 2021-11-08
Feed Example Display Page: Change h2 to h1

WilcoFiers:
Replace the h2 with an h1. There's no reason for that first heading to be an h2.


mcking65 (approved):
Thank you @WilcoFiers. LGTM.

PR #2121, opened 2021-11-08
Tree Pattern: Update selection guidance to include aria-checked

mcking65:
ARIA 1.3 clarifies when browsers may provide implicit selection.
Those changes to the ARIA specification enable APG to provide guidance on using ariachecked to represent selection in trees. This change updates the roles, states, and properties section of the tree view pattern accordingly.

[Preview changes to tree pattern in RawGitHack](https://raw.githack.com/w3c/aria-practices/tree-pattern-selected/aria-practices.html#TreeView)

Note: revisions to listbox will be addressed separately.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/pull/2121.html" title="Last updated on Nov 15, 2021, 2:55 AM UTC (780c24b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-practices/2121/59da387...780c24b.html" title="Last updated on Nov 15, 2021, 2:55 AM UTC (780c24b)">Diff</a>

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -3120,7 +3120,7 @@ Some JavaScript and CSS may not function correctly in Internet Explorer.
             <ul>
               <li>
                 Some design systems use <code>aria-selected</code> for single-select widgets and <code>aria-checked</code> for multi-select widgets.
-                In the abscence of factors that would make an alternative convention more appropriate, this is a recommended convention.
+                In the absence of factors that would make an alternative convention more appropriate, this is a recommended convention.
               </li>
               <li>
                 The language of instructions and the appearance of the interface might suggest one attribute is more appropriate than the other.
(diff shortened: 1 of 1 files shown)


JAWS-test:
I haven't followed the entire discussion on the topic, but it seems odd to me that in the paragraph aria-checked and aria-selected can be used arbitrarily. Shouldn't there be some indication that primarily aria-selected should be used. Only if selected and checked are two different states that a treeitem has, both should be used?

mcking65:
@JAWS-test wrote:

> I haven't followed the entire discussion on the topic, but it seems odd to me that in the paragraph aria-checked and aria-selected can be used arbitrarily. Shouldn't there be some indication that primarily aria-selected should be used.

I'm planning to add a note related to the choice of selected or checked before this is ready for review.

>Only if selected and checked are two different states that a treeitem has, both should be used?

In the spec, We included a set of 3 conditions that would have to be met by a tree that uses both checked and selected. Basically, we strongly advise against designing UI that would use both. It would be very difficult to ever meet all three conditions. I'm also planning to add a note about this as well.

jnurthen (approved):

Issue #1811, opened 2021-03-08
Alternatives to adding a PR comment with the output of coverage-report.yml workflow


zcorpan:
#1788 removes the regression-coverage-report check's code that comments on the PR with the output of the regression test coverage check, because the way it's implemented is unsafe.

Relaying a discussion over email with myself, @jesdaigle, @mcking65, @nschonni:

I wanted to learn why the check is posting a comment to the PR:

> In Travis-CI, this was a perma-failing job through “Allowed Failures”, and GitHub does not have an equivalent https://github.com/actions/toolkit/issues/399. I think it was also just an ignored, since it was only there if you clicked through

> the checks tab in a PR is kind of ugly with screen readers. It is tricky to work. Seems to have some kind of scroll-related issues.

So there are several problems here:

1. The regression test coverage report has errors on main.
2. The regression test coverage report is being ignored.
3. GitHub Actions doesn't allow non-blocking failing checks. (Though, allowing them makes the check easier to ignore.)

I think we should address the first point by fixing the errors.

I think we should address the second point by enforcing the check, so the main branch remains in an evergreen no-error state.

Then the third point is a non-issue, since we want it to be a blocking check to not fall back into the state of having errors and the check being ignored.

So then where and how should we present the output of the check? Here are some options:

1. Show the output formatted in the same way as the comment but in the Checks tab. When the check fails, the "Details" link should show the right page.
2. Annotate the "Files changed" tab with review comments on the relevant lines in the source.
3. Have an external page, and link to it from the Checks page.

These aren't necessarily mutually exclusive. A combination of (1) and (2) could be good, if we can get accessibility issues fixed or if we can work around them somehow.

Possible changes we could propose to GitHub:

* Change GitHub UI to fix accessibility issues with the Checks page. (@mcking65 mentioned that it doesn't work well with screen readers. I'd like to learn more about what the issues are specifically.)
* Change GitHub API to allow bots to comment with a comment-only access token, without full write access.
* Change GitHub Actions to allow non-blocking failing checks. https://github.com/actions/toolkit/issues/399

nschonni:
#1976 restricts the token, to resolve the "unsafe" aspect

zcorpan:
The workflow is now enabled again, see https://github.com/w3c/aria-practices/pull/2068

However, it fails for PRs that are from a fork, e.g. https://github.com/w3c/aria-practices/pull/1869

[...]

PR #2123, opened 2021-11-09
Infrastructure: Add regression coverage problem matcher

nschonni:
Fixes https://github.com/w3c/aria-practices/issues/1811

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -1,24 +1,21 @@
 name: Regression Tests Coverage Report
 
 on:
-  pull_request_target:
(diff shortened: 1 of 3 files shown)


nschonni:
As the current version of the job is disable due to the permissions issues, you can see what the output looks like on https://github.com/nschonni/aria-practices/pull/282/files

howard-e (approved):

nschonni:
@howard-e @zcorpan if this lands, you probably need to re-enable the job, because I think it is still turned off because of the current forking permission error

howard-e:
Okay got it

PR #2125, opened 2021-11-10
Switch example using checkbox: Remove browser prefixes from CSS and fix stylelint errors

nschonni:


examples/switch/css/switch-checkbox.css
```diff
@@ -18,17 +18,15 @@ label {
 label .label {
   display: inline-block;
   width: 9em;
-  -webkit-user-select: none;
```
nschonni, line 21: @jongund I think you added these specifically before, but the updated Stylelint PR cleaned out a bunch. If you think it's needed they can be ignored https://stylelint.io/user-guide/rules/list/property-no-vendor-prefix/ or Autoprefixer could be added to the build to add any needed ones based on the supported browsers

mcking65:
@jongund  @zcorpan  @howard-e 

What is best way forward with this. I'm getting errors in main now that are also showing up in the apg-1.2 branch. Do we need to ignore the prefixes in switch?

zcorpan:
Probably we should allow `-webkit-` properties in the stylelint config

https://stylelint.io/user-guide/rules/list/property-no-vendor-prefix/#optional-secondary-options

```
ignoreProperties: ["/^-webkit-/"]
```

jongund:
@zcorpan @mcking65 I think configuring the linter to ignore `-webkit`  is the way to fix it for today, if it works.  Otherwise I guess just take the line with `-webkit-` line out of the CSS and it may break the example in some way for Safari browser.

nschonni:
Opened https://github.com/w3c/aria-practices/pull/2128 draft as an option to use Autoprefixer so the author doesn't need to know which prefixes are still required for the browsers, and only fixes it when publishing the gh-pages branch

jongund:
@nschonni @mcking65 I did some testing on Chrome and Safari and I don't this we need the `-webkit-user-select` anymore, so just delete that line from the CSS file `switch-checkbox.css`.

zcorpan:
@jongund OK. If we can get away with no prefixes, that would be nice IMO. Standardized properties should be supported without prefix, so we can blame browsers when that's not the case.

nschonni:
@mcking65 do you want to land this one, so the other PRs go green again? The regression error appears to just be a hiccup in the actions doing the NPM install

mcking65:
Thank you @jongund  for looking into this; I didn't want to break it in Safari.
Thank you @nschonni for the PR. I'll land now based on this discussion.

nschonni:
Looks like the deploy failed https://github.com/w3c/aria-practices/runs/4200265089?check_suite_focus=true#step:3:44. Not sure if there is an issue with the token, or another GitHub Actions issue related to forks

Issue #2110, opened 2021-11-03
Update the Code Guide documentation and lint everything


zcorpan:
The [Code Guide](https://github.com/w3c/aria-practices/wiki/Code-Guide) wiki page has various recommendations for HTML, CSS, and JavaScript source code style.

We also have various linters, runnable with `npm run lint`.

We should make sure that the Code Guide reflects the code style that we want, and remove everything that is not important or not easily lintable, and then make sure that the linters check for everything that the code guide requires (where possible). We can implement new lint rules in upstream projects (like in #2069).

If we require things that are not lintable, clarify which rules require manual review.

In the last TF meeting we briefly discussed changing the [CSS prefixed properties formatting rule](https://github.com/w3c/aria-practices/wiki/Code-Guide#prefixed-properties): either remove the rule, or change to disallow prefixed properties.

Other possible changes:

- HTML
  - [Practicality over purity](https://github.com/w3c/aria-practices/wiki/Code-Guide#practicality-over-purity): This rule seems vague and hard to review against, and impossible to lint. Suggest removing.
  - [boolean attributes](https://github.com/w3c/aria-practices/wiki/Code-Guide#boolean-attributes): the rule is fine, but it can be one sentence.
  - [reducing markup](https://github.com/w3c/aria-practices/wiki/Code-Guide#reducing-markup): this seems hard to review for and impossible to lint. Suggest removing.  
- CSS
  - [syntax](https://github.com/w3c/aria-practices/wiki/Code-Guide#syntax-1) may need some updates to match current `Stylelint` (https://github.com/w3c/aria-practices/pull/2077#discussion_r741088623).
  - [declaration order](https://github.com/w3c/aria-practices/wiki/Code-Guide#declaration-order): is this rule helpful or is it mostly make-work for contributors and reviewers? Is it readily lintable?
  - [shorthand notation](https://github.com/w3c/aria-practices/wiki/Code-Guide#shorthand-notation): I don't understand the benefit of this rule, and it seems impossible to lint for. Suggest removing.
  - [code comments](https://github.com/w3c/aria-practices/wiki/Code-Guide#comments) is a bit vague, and not lintable. Either remove or clarify.
  - [class names](https://github.com/w3c/aria-practices/wiki/Code-Guide#class-names) has 6 bullet points, only the first one is lintable. Can we simplify? Also remove reference to Sass/Less.
- JS
  - Is the reference to AirBnB JavaScript Style Guide what we should use? Are the linters following it?

[...]

PR #2128, opened 2021-11-10
Infrastructure: Add postcss and autoprefixer to deploy

nschonni:
Since Stylelint is suggesting removal based on the assumption Autoprefixer does a better job, this adds it to the gh-pages deploy so the final version does add any required vendor prefixed values


examples/css/core.css
```diff
@@ -53,6 +53,8 @@ table.data.attributes tbody td {
   font-family: inherit;
   line-height: 1.2;
   hyphens: manual;
+  hyphens: manual;
```
nschonni, line 56: Opps, the staglint is fixing the sample before the checkin

nschonni:
I've marked this as draft for now since it has the commit just to show the CSS changes. I'll rebase it out and flip it to ready after there has been some review.
//cc @jongund @mcking65 @zcorpan @howard-e

.github/workflows/deploy.yml
```diff
@@ -47,11 +47,18 @@ jobs:
       run: |
         curl "https://labs.w3.org/spec-generator/?type=respec&url=https://raw.githack.com/${{ github.repository }}/${{ github.sha }}/aria-practices.html" -o deploy/index.html -f --retry 3
 
+
+    - name: Run PostCSS
+      run: |
+        cd deploy
```
nschonni, line 53: Could also move this section right into the next block too, if the double `cd` might cause an issue with the runner logic

nschonni:
Rebased it out, but you can see the resulting CSS changes in https://github.com/w3c/aria-practices/pull/2128/commits/2e965b94d10b02a0032fdf7ea6c3fe6162cb0221

zcorpan:
Commented in the commit, but we should decide on a policy regarding prefixes in https://github.com/w3c/aria-practices/issues/2110#issuecomment-965749133

nschonni:
@zcorpan this should probably also include a `.broswerlistrc` https://github.com/postcss/autoprefixer#browsers that can be used by stylelint and other tools. Some of the things you're pointing out are likely because there is no current defaults

nschonni:
Looks like only Chrome and Firefox right now https://github.com/w3c/aria-practices/wiki/Code-Guide#browser-support, but I'm wondering if Safari shouldn't be included as well

PR #2150, opened 2021-11-19
Infrastructure: Update stylelint for opacity changes

nschonni:
This was changed during the last major version, but it looks like they moved back from the precentage format quickly


howard-e (approved):
Supersedes #2143, #2142 & #2124.

https://github.com/w3c/aria-practices/pull/2143#issuecomment-969034876 ha! :)

PR #2158, opened 2021-11-23
Infrastructure: Chore: Post-dependabot merge cleanup

howard-e:
- [x] Resolve lockfile inconsistencies with `npm audit fix`. Relevant comment: https://github.com/w3c/aria-practices/pull/2069#discussion_r754746023
- [x] Update `lint-staged` config. Relevant comment: https://github.com/w3c/aria-practices/pull/2154#issuecomment-976093237

PR #2166, opened 2021-12-02
Infrastructure: cleanup husky script

nschonni:
Lint staged v12 recommends dropping the `git add`s.
The husky script is supposed be ignore by `.gitignore` and added by the install script.

/cc @howard-e


nschonni:
Oops, this is a partial dupe of #2158

nschonni:
Removed the parts that overlapped with 2158

PR #2179, opened 2021-12-25
Code guide: swap deprecated getYear method in datepicker dialog example

guiseek:


nschonni:
Reference to the depreciation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getYear

ZoeBijl:
Thank you @guiseek!

PR #2246, opened 2022-02-28
Workflows to support APG Redesign connection with w3c/wai-aria-practices

howard-e:
To facilitate the [APG Redesign](https://github.com/w3c/wai-aria-practices), this PR implements the workflows outlined in https://github.com/w3c/wai-aria-practices/issues/13.

Related PRs:
- https://github.com/w3c/wai-aria-practices/pull/15.

Workflows:
* wai-trigger-deploy.yml
  - Triggers w3c/wai-aria-practices `deploy.yml` workflow to update APG Redesign production URL with latest APG content.
* wai-trigger-pr.yml
  - Triggers w3c/wai-aria-practices `pr-create.yml` workflow to populate PR with preview from APG Redesign.
  - eg. Note `WAI Preview Link` in https://github.com/howard-e/aria-practices/pull/35.
* wai-trigger-cleanup.yml
  - Triggers w3c/wai-aria-practices `remove-branch.yml` workflow to remove branches which are no longer required when PRs have been merged/closed.

### DIFF
--- a/.github/workflows/wai-trigger-cleanup.yml
+++ b/.github/workflows/wai-trigger-cleanup.yml
@@ -1,0 +1,54 @@
+name: Trigger branch cleanup workflow for wai repository
+
+on:
+  delete:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  cleanup-branch-wai:
+    if: ${{ github.event_name == 'delete' && github.event.ref_type == 'branch' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          cache: npm
+
+      - name: Install @octokit/rest
+        run: npm install @octokit/rest
+
+      - name: Run wai-trigger-cleanup script
+        run: |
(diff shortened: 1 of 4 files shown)


howard-e (commented):
This PR makes use of creating workflow dispatch events to https://github.com/w3c/wai-aria-practices and so, depends on a token which [has the repo scope](https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event).

> You must authenticate using an access token with the `repo` scope to use this endpoint.

I’ve assumed `secrets.GH_TOKEN` has the permissions as I’ve seen its usage in `workflows/deploy.yml`, but it’s also being used in a case where it could be ignored. This may need to be verified.

This is required because the GITHUB_TOKEN provided by GitHub in the workflow [won’t work across repositories](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).

> The token's permissions are limited to the repository that contains your workflow.

howard-e:
Closing due to [rules around using secrets from forked repositories](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow).

> ... secrets are not passed to the runner when a workflow is triggered from a forked repository.

PR #2259, opened 2022-03-14
Infrastructure: Add Home Page for new WAI-APG site

alflennik:
Adds a homepage for ARIA Authoring Practices, which will be publicly viewable as part of the [website redesign project](https://github.com/w3c/wai-aria-practices).

The home page content can be previewed here: 

https://deploy-preview-18--wai-aria-practices2.netlify.app/

There is [a second PR](https://github.com/w3c/wai-aria-practices/pull/18) in the redesign repo to consume the new HTML file. This PR will need to be merged before the second PR can be merged.


index.html
```diff
@@ -0,0 +1,239 @@
...
+            href="https://github.com/w3c/aria-practices/wiki/Regression-Tests-for-APG-Example-Pages"
+            aria-label="View regression test documentation"
+          >
+            Learn More
+          </a>
+        </li>
+        <li>
+          <h3>Meetings</h3>
+          <p>
+            The APG Task Force meets on Tuesdays at 14:00 until 15:00 Boston
```
howard-e, line 145: Perhaps an optional observation which could be addressed another time but wondering if `Boston time` conveys the same relative meaning to everyone as they would with GMT/UTC specified times? 

Perhaps we could also use the language in meeting agendas and invites, which state `Eastern Time`?

howard-e (approved):
Looks good @alflennik

index.html
alflennik: This is a good point, that kind of stuck out to me too. I've snuck in that change.

PR #2263, opened 2022-03-28
Infrastructure: update html linting globs

nschonni:
~~Didn't fix any of the issues flagged yet, just updating the CI to start catching new issues again~~


nschonni:
@howard-e I pulled in the fixes for the index, plus another for VNU.
Looks like the failure for the WAI PR creation needs to use `pull_request_target` to access the token

howard-e:
> Looks like the failure for the WAI PR creation needs to use `pull_request_target` to access the token

Right, seems to be the case. Just made the note here too https://github.com/w3c/aria-practices/pull/2262#discussion_r836591645

I'll go ahead and update that PR and remove the `index.html` changes as well

howard-e:
Should be able to rebase on `main` to address the build error here @nschonni

howard-e:
Hmm unsure about this one. I'll pull the `index.html` fix into a separate PR for now to get main green again.

howard-e:
> Hmm unsure about this one. I'll pull the `index.html` fix into a separate PR for now to get main green again.

https://github.com/w3c/aria-practices/pull/2265

howard-e:
```
---
    name: AssertionError
    message: >-
      After sending ENTER on the "previous month" button, the month and year text
      should be not be March 2022
    assertion: not
    values:
      'Value is the same as:': '''March 2022'''
    at: 'test/tests/combobox_datepicker.js:449:9'
  ...
```

@nschonni so taking another look at the example, it would seem there's an edge case not covered by the test where if the day of the current month doesn't exist in the previous month, it returns to the 1st of the current month (+ additional offset in the case of March, see other failing cases below) when the previous month button is used. 

A passing case:
1. Ensure `Mar 28` is selected and open date picker
2. Navigate to previous month button and select
3. Result is `Feb 28`, since `Feb 28` exists.
 
A failing case: 
1. Ensure `Mar 29` is selected and open date picker
2. Navigate to previous month button and select
3. Result is `Mar 1`, since no `Feb 29` (in 2022).

Other failing cases:
1. Assume `Mar 30` selected > open date picker > navigate to previous month button and select > `Mar 2`, since no `Feb 30`.
2. Assume `Mar 31` selected > open date picker > navigate to previous month button and select > `Mar 3`, since no `Feb 31`.
3. Assume `Dec 31` selected > open date picker > navigate to previous month button and select > `Dec 1`, since no `Nov 31`.

Since this PR hasn't caused that issue, I think we could land this and address the edge case in a separate issue/PR

nschonni:
Yeah, I think I fixed a few of the datepicker tests awhile ago, but they really need to pin to specific months (plus add some of those specific cases you mentioned) instead of relying on the current month

PR #2288, opened 2022-04-12
Checkbox and Date Picker Dialog Examples: Editorial change to roles states and properties table markup

### LINKED ISSUE #2285: checkbox and datepicker-dialog have tables with empty headers
`checkbox/checkbox.html` and `dialog-modal/datepicker-dialog.html` both have tables that have empty `th` elements. I suspect these could be `td` elements instead as other parts of the same table do that.

```html
<tr>
  <th></th>
  <th scope="row"><code>aria-live="polite"</code></th>
</tr>
```

image

jongund:
Resolves #2285

___
[WAI Preview Link](https://deploy-preview-38--wai-aria-practices2.netlify.app)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -666,6 +666,7 @@
             <td>
               <ul>
                 <li><a href="accordion/accordion.html">Accordion</a></li>
+                <li><a href="checkbox/checkbox.html">Checkbox  (Two State)</a> (<abbr title="High Contrast Support">HC</abbr>)</li>
                 <li><a href="combobox/combobox-datepicker.html">Date Picker Combobox</a> (<abbr title="High Contrast Support">HC</abbr>)</li>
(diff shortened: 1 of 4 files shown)


examples/index.html
```diff
@@ -666,6 +666,7 @@ <h2 id="examples_by_props_label">Examples By Properties and States</h2>
             <td>
               <ul>
                 <li><a href="accordion/accordion.html">Accordion</a></li>
+                <li><a href="checkbox/checkbox.html">Checkbox  (Two State)</a> (<abbr title="High Contrast Support">HC</abbr>)</li>
```
mcking65, line 669: @jongund 
Does this change mean that the index builder requires the attribute to be in a `th` element when listed in the roles, state, and props table? If yes, I wonder if that code should be a bit more forgiving?

mcking65 (approved):
Thank you for fixes; LGTM.

PR #2404, opened 2022-07-06
Infrastructure: Bump stylelint-config-standard and adjust core.css to rule

howard-e:
* Update `stylelint-config-standard` from v24.0.0 to v26.0.0.
* Adjust `examples/css/core.css` to follow new import-notation rule.
  * NB. We could also ignore the rule if there's any strong feeling/reason for the current url syntax (`@import url("")`).

Supersedes #2398. 
___
[WAI Preview Link](https://deploy-preview-150--wai-aria-practices2.netlify.app)

### DIFF
--- a/examples/css/core.css
+++ b/examples/css/core.css
@@ -1,5 +1,5 @@
 /* Style for syntax highlighting */
-@import url("github.css");
+@import "github.css";
 
 body:not(.toc-inline) {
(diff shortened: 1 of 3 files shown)


howard-e:
Closing as it seems this is no longer a rule being pushed in #2527

PR #2437, opened 2022-08-02
Apply aria-hidden to the font icon for the calendar button in the Date Picker Dialog Example

OwenEdwards:
Apply aria-hidden to the font icon for the calendar button in the Date Picker Dialog Example, to avoid the possibility that it gets announced by screen readers

### DIFF
--- a/examples/dialog-modal/datepicker-dialog.html
+++ b/examples/dialog-modal/datepicker-dialog.html
@@ -62,7 +62,7 @@
             <button type="button"
                     class="icon"
                     aria-label="Choose Date">
-              <span class="fa fa-calendar-alt fa-2x"></span>
+              <span class="fa fa-calendar-alt fa-2x" aria-hidden="true"></span>
             </button>
 
           </div>


mcking65:
There's an aria-label on the button so screen readers will not read any descendant content. If they do, there is a browser bug in the implementation of the accname spec.

It is advisable to avoid using aria-hidden except for in the rare circumstances where it is absolutely necessary.

OwenEdwards:
@mcking65 

> There's an aria-label on the button so screen readers will not read any descendant content. If they do, there is a browser bug in the implementation of the accname spec.

That will work for the naming of the button when tabbing through the page, but will it hide the font icon from being read with the Virtual Cursor?

I see adding `aria-hidden` to a span with a font icon as very similar to including it on an `<svg>` element in this same situation, or including `alt=""` on an icon in an `` element inside a button. Wouldn't this code be flagged as an a11y violation for not having an `alt` on the ``?:

```
<button type="button"
  class="icon"
  aria-label="Choose Date">
  <img src="fa-calendar-alt.png">
</button>
```

OwenEdwards:
@mcking65 you wrote:

> It is advisable to avoid using aria-hidden except for in the rare circumstances where it is absolutely necessary.

Please clarify what you consider "absolutely necessary" - I have seen situations where font icon content is "read" by screen readers, and that seems like a situation where `aria-hidden` *should* be used, as an equivalent to `alt=""` on an `img` element. How is that not "absolutely necessary"?

JAWS-test:
@OwenEdwards 

If an icon would be part of the label according to [Accname](https://www.w3.org/TR/accname-1.2/), but should not be, then it must be marked with aria-hidden=true (e.g. a link that contains text and icon and has no aria-label). But if according to Accname the content of the element does not contribute to the label anyway (because e.g. aria-label is present), then it does not matter if graphics are included (then also an img-graphic does not need an empty alt attribute).

mcking65:
@OwenEdwards 

I'm closing this because I believe we have concluded that aria-hidden is not necessary. If you disagree, please raise an issue where we can discuss the rationale for change.

Since the time when you submitted this PR, we restructured the repository. So, if we are going to make a change, it would need to be done with a new PR.

OwenEdwards:
@mcking65 this point has been rattling around in my brain since I raised it, and I recently found an example that reminded me of it:

````
<!-- Accessibility control on https://www.guidedogs.com/accessibility-statement -->
<div class="accessibility-tools d-none d-lg-block active" aria-expanded="true">
  <button aria-controls="accessibility-tools" id="open-accessibility-panel" aria-label="Open Accessibility Toolbar">
    <img src="/assets/images/aa.svg" alt="Open Accessibility Toolbar">
  </button>
  <button aria-controls="accessibility-tools" id="close-accessibility-panel" aria-label="Close Accessibility Toolbar">
    <em class="fa fa-close"></em>
  </button>
  <div id="accessibility-controls">

    [...]

  </div>
</div>
````

In this case, one button is visually labeled with an `` element, while the other is labeled with a font icon.

Now, in this code, the `alt="Open Accessibility Toolbar"` on the `` element is redundant because the button is already labeled with `aria-label`, right? So, you could argue that the alt should `alt=""` instead.

In your initial reply to this PR, you said :
> There's an aria-label on the button so screen readers will not read any descendant content. If they do, there is a browser bug in the implementation of the accname spec.

In that case, couldn't someone argue that the `` element doesn't need an alt attribute _at all_, because "screen readers will not read any descendant content"? Automatic a11y checkers would flag not having an alt attribute on that `` as a violation, though, irrespective of the fact that it is inside a button with `aria-label`.

In my experience, developers need as few nested "_if a do b but if c do d but if e do f_" rules to follow as possible. My guidance to developers is that they should _always_ include an alt attribute on an `` element, and use a null alt if the `` is decorative; then, for a decorative `<svg>` element, use `aria-hidden="true"`, and also _always_ use `aria-hidden="true"` on the element containing a font icon, to prevent the Unicode character(s) inserted by CSS from being announced by a screen reader, because those characters do not relate to the meaning of the visual representation.

I don't think saying "... unless the img/svg/font icon is inside an element with an AccName that doesn't use that img/svg/font icon, in which case you don't need to mark the img/svg/font icon as decorative" helps developers because they won't reliably understand when that's true and when it isn't.

So I think using `aria-hidden="true"` on the font icon element for _consistency_, even though it isn't strictly necessary, leads to _fewer_ developer errors with font icons and the use of aria-hidden. I think consistency in the rule of when to include `aria-hidden="true"` on elements containing font icons is more important than "avoid using aria-hidden except for in the rare circumstances where it is absolutely necessary," because there is no _harm_ in using it in this situation.

What do you think?

Issue #2418, opened 2022-07-25
Reorganize patterns and examples to reflect site architecture


mcking65:
This issue documents one portion of the [reorganize repository project](https://github.com/w3c/aria-practices/projects/42).

#### Objective
Build a content tree containing the home page, patterns, examples, and index. The practices and about sections of the site will be handled in subsequent steps.

#### Directory Structure
* "content" directory at root level: This will hold all content in a tree that matches the IA of the web site. It has subdirectories for:
    * patterns
    * index
    * shared (for js, css, scripts that need to be referenced by multiple resources. Did not name "common" because we have not yet eliminated aria-common)
* There will be a directory for each pattern that contains:
   * A file for the pattern, e.g., "accordion-pattern.html". Originally considered naming "index.html" but decided a more specific file name could avoid confusion with bulk file operations.
   * An examples directory for the examples.

#### Plan
1. Create subdirectories for patterns under /patterns directory. Note: name the pattern directory the same way it is named in w3c/aria-practices/examples/. For example, the modal dialog pattern directory should be named "dialog-modal", not "dialogmodal". If the new pattern directory name differs from the path name in the currently published wai/aria/apg, add a note about the need for pattern redirect in the redirect notes comment below.
2. For each pattern, create a pattern file from content in aria-practices.html, e.g., `/content/patterns/accordion/accordion-pattern.html`. Copy content from respec file into the new pattern file and edit as needed.
3. In each pattern subdirectory, create an examples directory and move the example files into that directory.
4. Move shared js and css into shared directory.
5. Edit example html files to fix references and minimize scripted content transformation (instructions in comment below).
6. Reconfigure regression tests to run test against examples in new location (regression-tests.sh)
7. Update create-gh-project.js to use new examples location.
8. Update reference-tables.js to use new pattern and example locations and create output in `/content/index/generated-content-index.html`
9. Update coverage-report.js to use new pattern and example locations and create output in /content/about/coverage-reports/
10. Update github actions to utilize new locations
11. Modify build process to:
    * pull pattern content from new pattern content files;
    * pull example content from new example locations;
    * pull index content from new index location;
    * Stop making content additions or modifications to patterns, examples, and index.
13. Configure redirect for pattern if the pattern URL will be different from the URL for that pattern when the WAI site was launched (e.g., for modal dialog because the directory name changed from dialogmodal to dialog-modal.
14. Configure redirect for  examples since they are no longer in the examples-index path.

mcking65:
[...]

PR #2450, opened 2022-08-29
Revise index generation to use new content tree (Issue 2418 task 8)

alflennik:
See https://github.com/w3c/aria-practices/issues/2418 for more information about the task.


mcking65 (changes requested):
Tested, working great!

However, I think we need to do two more things with this PR.

1. Delete the content/index.html file generated by the old version of the script.
2. Change content of scripts/reference-tables.template to match what is in prod on the WAI site. One of our goals of the restructure is for all content in prod to be in the aria-practices repo. The jekyll build should not add or modify content.

alflennik:
@mcking65 Thanks for the feedback, and I implemented the fixes. I had indeed missed the previous index.html file, thanks for noticing it. I also updated the copy to match the production site.

scripts/reference-tables.template
```diff
@@ -11,15 +11,12 @@
   <script src="js/app.js"></script>
```
mcking65, line 11: The paths on the file refs in the script and css don't appear correct

mcking65 (changes requested):
Looking good; just one more thing. I think we need some edits to the head in the template. I don't know which script and css is needed. We should think about both what is needed in prod and what helps people when viewing the content locally in clones of the repo.

alflennik:
@mcking65 Thanks again for pointing out the broken link, I fixed it and several other broken links on this page, and now the JS and CSS running on this page are the same as the other example pages.

mcking65 (approved):

Issue #2424, opened 2022-07-26
Create new pages for practices content


mcking65:
This issue is part of the [Restructure repository Project](https://github.com/w3c/aria-practices/projects/42).

In the new content tree, create a page for each section of aria-practices.html that serves as a source of a practices section page in the web site.

mcking65:
Discussion point: Should we consider having a directory for each practice in the content tree? That would give the option of later having a practice that includes multiple files. It would more closely mirror the structure of the patterns and about branches of the tree.

mcking65:
#2417 will resolve this issue.

As of today, the practices are moved out of aria-ppractices.html and are in the move-examples branch in a practices subdirectory in the new content tree. All links to practices are updated.

@alflennik, the next steps for practices are:

* Apply css to the practices files in the new content tree
* Put the images for the practices cards into the practice subdirectories of the practices directory
* Build content/practices/practices.html (perhaps in a manner similar to the way the index is built)

alflennik:
Sounds good @mcking65!

PR #2451, opened 2022-08-29
Update coverage report script to process new content tree and put output into about (Issue 2418 task 9)

alflennik:
See https://github.com/w3c/aria-practices/issues/2418 for more information about the task.


mcking65 (changes requested):
@alflennik 

Looking good so far.

I am wondering if we might want to break out all the content repairs made to the example files into a separate PR that we would merge to move-examples branch.

We might want to consider holding off on merging the coverage-report script and location changes until after we complete both issues #2418 and #2424. That is because I now see there are a lot of links in the coverage reports to the practices. All those links, which are generated by the script, will have to be updated after #2424 is complete.

I have updated content of the coverage-report.template file.

Will you also please change the coverage-report.js to put the output into `content/about/coverage-and-quality`. Also change the name of the generated html file from `index.html` to `coverage-and-quality-report.html`.

Will this PR also need to include changes to `.github/workflows/coverage-report.yml`?

mcking65 (changes requested):
@alflennik 

There are two issues remaining:
1. The links in the generated output have incorrect href values; I get file not found when following them.
2. The output is different from the output in main. Content has not changed, so the report should be the same.

Compare generated out in main: [coverage/index.html in main](https://github.com/w3c/aria-practices/blob/279111f90c59d281b0359fc1a917e13d23f4211f/coverage/index.html)
To generated out output in move-examples-9: [content/about/coverage-and-quality/coverage-and-quality-report.html in move-examples-9](https://github.com/w3c/aria-practices/blob/f5c246bfc4517a87a036cb6e09ce7396343007fb/content/about/coverage-and-quality/coverage-and-quality-report.html)

It appears that the processing of practices to get information about guidance is not yet updated.

alflennik:
@mcking65 Thanks for the review! You are right about the broken links, and that is now fixed; it occurred when I updated the directory of the coverage report.

About the differences in report content, the reason for that, I think, is that the coverage report GitHub action seems to be disabled and has been disabled for some time (around a year?). [Here is a link to the page where the script is marked as disabled.](https://github.com/w3c/aria-practices/actions/workflows/coverage-report.yml)

mcking65:
@alflennik wrote:

> @mcking65 Thanks for the review! You are right about the broken links, and that is now fixed; it occurred when I updated the directory of the coverage report.

Awesome, thank you!!

> About the differences in report content, the reason for that, I think, is that the coverage report GitHub action seems to be disabled and has been disabled for some time (around a year?). [Here is a link to the page where the script is marked as disabled.](https://github.com/w3c/aria-practices/actions/workflows/coverage-report.yml)

I created branch test-coverage-report from tip of main to test this hypothesis. In that branch, I ran scripts/coverage-report.js against the content in main. That version of the script is listing guidance that is not being listed by the script in move-examples-9.

Compare to [/coverage/index.html in test-coverage-report](https://github.com/w3c/aria-practices/blob/97ab0294f87498cf3e005817fb9e7edcd35d550c/coverage/index.html).

Here are two examples of differences that need to be resolved.

Roles with No Guidance or Examples:
* 29 in the test branch
* 31 in the move-examples-9 branch, including presentation and tooltip, both of which have guidance in the move-examples-9 content tree that should be listed.

Roles with at Least One Guidance or Example:
* 13 in the test-coverage branch
* 17 in the move-examples-9 branch: there are more here because several roles should be listed in the 2 or more table, e.g., alert and cell have guidance in the move-examples-9 branch that is not listed.

alflennik:
The inaccurate coverage numbers were a great catch @mcking65, and thank you for the thorough review. Also thank you for running the coverage report on another branch, having that for reference came in handy. After investigating I found that moving the practices into their own files was not something I had accounted for, because I had opened the PR before that change was made. Now I've updated this branch to load the practices from their own files and I confirmed that the coverage levels are equivalent and that also all the links work on the coverage report.

mcking65 (approved):

PR #2464, opened 2022-09-01
Update GitHub workflows to use new content tree (Issue 2418 task 10)

alflennik:
See https://github.com/w3c/aria-practices/issues/2418 for more information about the task. Note that the build will fail until the changes on the move-examples-6 branch (https://github.com/w3c/aria-practices/pull/2441) are approved and merged.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -6,9 +6,8 @@ on:
       - ".github/workflows/regression.yml"
       - "package*.json"
       - "test/**"
-      - "examples/**"
-      - "!examples/landmarks/**"
-      - "!examples/coding-template/**"
+      - "content/**/examples/**"
+      - "!content/landmarks/examples/**"
   push:
     branches-ignore:
       - main
@@ -17,9 +16,8 @@ on:
       - ".github/workflows/regression.yml"
       - "package*.json"
       - "test/**"
-      - "examples/**"
-      - "!examples/landmarks/**"
-      - "!examples/coding-template/**"
+      - "content/**/examples/**"
+      - "!content/landmarks/examples/**"
 
 jobs:
   regression:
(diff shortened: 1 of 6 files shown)


mcking65 (approved):
LGTM ... although this YML is definitely not something I fully understand.

Would be really nice if @nschonni were to also take a look.

.github/workflows/regression.yml
```diff
@@ -6,9 +6,8 @@ on:
       - ".github/workflows/regression.yml"
       - "package*.json"
       - "test/**"
-      - "examples/**"
-      - "!examples/landmarks/**"
-      - "!examples/coding-template/**"
```
nschonni, line 11: I think this still exists as `content-templates` now

.github/workflows/spelling.yml
```diff
@@ -9,8 +9,7 @@ on:
       - "!**/*.min.js"
       - "!**/*.pack.js"
       - "!**/*.paint"
-      - "!aria-practices-DeletedSectionsArchive.html"
```
nschonni, line 12: This file still exists in the branch

nschonni (commented):
Taking a quick look at the branch structure, it mostly makes sense, except that the `aria-practices-DeletedSectionsArchive.html` still exists (but maybe is supposed to be deleted in a different PR) and the old `coding-templates` I believe are now `content-templates` folder that needs to be ignored

mcking65:
I just removed aria-practices-DeletedSectionsArchive.html from the move-examples branch.

alflennik:
Thanks for the approval @mcking65 and the review @nschonni! Your comment is appreciated nschonni, it's definitely confusing to include references to changes that are actually occurring in other branches. But it sounds like Matt mitigated the issue by removing a file in the base branch.

nschonni:
Yeah, I delete the comment about the Archive file, but I don't think the one about the content templates was addressed

PR #2476, opened 2022-09-15
Add note on APG vs. browser multiselect listbox keyboard pattern

### LINKED ISSUE #2193: Listbox with Multiple Selection: "Recommended selection model" is uncommon
https://w3c.github.io/aria-practices/#Listbox:

Two operating concepts are presented and only one of them is recommended and implemented in the example. I understand that the recommended one is much easier to operate with mouse and keyboard However, the problem is that the non-recommended operating concept is the familiar one, because native multiple select works exactly the same way on the web and desktop. Since an AT user can't tell if it's a native or ARIA listbox, she or he will always try the default operation first, which then doesn't work in the example linked here. As long as the default operation does not match the operation recommended here, a note be added to the page (e.g., "The operation is recommended because it is easier for mouse and keyboard users. However, it has the disadvantage that it does not correspond to the standard and must therefore be learned first. Therefore, operating instructions should be provided to users").

smhigley:
Fixes #2193

Adds a note with the wording we discussed in a meeting on the Rearrangeable Listbox multiselect example.

### DIFF
--- a/content/patterns/listbox/examples/listbox-rearrangeable.html
+++ b/content/patterns/listbox/examples/listbox-rearrangeable.html
@@ -264,9 +264,13 @@ while in the second example, they may select multiple options before activating 
         </tr>
       </tbody>
     </table>
-    <h4 id="kbd_label_multiselect">Multiple selection keys supported in example 2</h4>
+    <h3 id="kbd_label_multiselect">Multiple selection keys supported in example 2</h3>
     <p class="note">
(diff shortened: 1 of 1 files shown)


mcking65 (approved):
Thank you @smhigley!

Looks great. I just made one minor revision to include a link back to the keyboard interaction section of the pattern where the multiple selection model is defined.

mcking65:
Since this is a minor editorial revision and I see a thumbs up from @JAWS-test, who raised the original issue, I am merging without further review.

Note: this is not going directly to main. It is will be merged with the restructuring project. I'm doing it this way so we can avoid fixing it twice. The move-examples branch should be in main in a couple weeks or so.

PR #2479, opened 2022-09-20
Move practices from aria-practices.html to new content tree

mcking65:
SStill WIP.

Starts process of moving the practices content from aria-practices.html into to the new content tree so we can update links from pattterns and examples to practices.


mcking65:
@alflennik 
Please have a look at what I am doing in this PR.

I am thinking about moving practices content at the same time we move patterns and examples. I think it will help with link processing on the build side.

PR #2499, opened 2022-10-04
Consistency pass over source code (Issue 2418 Task)

alflennik:
Following these changes, the examples, patterns and practices all now have a consistent visual style and appear to be part of the same website, with a consistent main landmark, and features such as skipto which are present on all pages.

In addition the source code is consistently formatted, making cross-cutting changes easier to implement, not to mention the fact that the content will be more easily processed when deployed to the WAI website.

Here's the changes made in this PR:

- Added equivalent stylesheets and scripts to all pages. These have been missing from the new patterns and practices pages.
- Added a main tag to all pages. In addition to being required by skipto.js, it seemed like a best practice worth implementing consistently across all pages. Unfortunately this introduces a lot of noise in the PR.
- Ran a HTML formatter on all pages. While this is another noisy change, it mostly touches the same lines as adding the main tag already did. I felt it was necessitated by the fact that currently a lot of the HTML is inconsistently and sometimes incorrectly indented.
- Made sure section ids were consistent, some ids were occasionally missing especially on patterns pages.
- Did a pass on the Example-Template.html since it had fallen out of sync.


mcking65 (changes requested):
@alflennik 

A quick first look tells me most of the changes are great! However, there is one I'd like to undo or change.

The HTML formatter made really long lines.

BTW, I am not a fan of a set line length, e.g., 80 char/line.

Long lines and set line lengths both introduce a similar problem for me, as an editor who uses a screen reader.
They increase cognitive load.

I've been using semantic line breaks in content.
That is, one sentence per line, and if a link has a long href, I put the link on a line by itself.

When editing, I think in sentences.
So, it helps productivity quite a bit to press down arrow and here a single complete sentence, and nothing more.

When a single line contains multiple sentences, the cognitive load of reading is much higher; it takes much more concentration.
It also substantially increases the number of key presses required to edit sentences in the middle or end of a paragraph, list item, or table cell.
And, since Git compares based on lines, multiple sentences on a single line makes reviewing PRs much harder.

When an href is long, it makes understanding the sentence more difficult.
So, that is one situation where I sometimes break a sentence into multiple lines.
That makes it easier to skip past the link to more readily focus on the content that surrounds it.`

alflennik:
@mcking65 Thanks for that info, this was the aspect of the PR that I was the most unsure about. I'll go through and give each sentence its own line. Your point about wrapping lines at 80 characters was interesting, and I like the approach of using semantic breaks as a compromise between auto wrapping and extremely long lines.

alflennik:
@mcking65 the formatting has been updated to give each sentence and long link its own line. This should preserve the current editing experience while also fixing all the indents. (Side note: I noticed the APG has a lot of sentences in it 😜)

mcking65:
Thank you, much better now!

Two questions:
1. What is the intent of including utils.js on example pages where it is not used?
2. What tool did you use for the HTML formatting? Should we include it in our automation?

alflennik:
@mcking65 thanks for the feedback. Regarding the first point about the utils.js file, I'm glad you made that observation, and including it on all pages is not actually necessary. I updated the PR so that utils is only included on pages where it is required.

About the second point of the formatter I used, the answer is that I initially started with Prettier and then made a number of manual changes after it had fixed the indents. I'm unfortunately not aware of any tool that could automatically fix indents without also disrupting conventions like putting each sentence on its own line.

mcking65 (approved):

mcking65:
@alflennik wrote:

> @mcking65 thanks for the feedback. Regarding the first point about the utils.js file, I'm glad you made that observation, and including it on all pages is not actually necessary. I updated the PR so that utils is only included on pages where it is required.

Perfect, looks great!

> About the second point of the formatter I used, the answer is that I initially started with Prettier and then made a number of manual changes after it had fixed the indents. I'm unfortunately not aware of any tool that could automatically fix indents without also disrupting conventions like putting each sentence on its own line.

Wow, manually fixed all that! I thought for sure you had some automation supporting semantic line breaks.

We do have `.editorconfig`, but it doesn't seem to help much. Even though I supposedly have the support for it installed in my eitors, I don't see it doing anything.

I wish prettier would have options to support semantic line breaking.

Issue #2488, opened 2022-09-28
Move alert dialog example under alert dialog pattern


mcking65:
We have two separate patterns for Alert Dialog and Modal Dialog. However, at this time, the alert dialog example is located under the pattern for Modal Dialog.

This is because the alert and modal dialog examples share some code. When I created the new content tree, I didn't separate alert Dialog from modal dialog because I didn't want to dig into figuring out which code is shared. To separate them, some code will need to be copied into the js for alert dialog. Then, we'll need to ensure the alert dialog example has not regressed in any way.

Now that the content tree is complete, we should do this.

Move:
`content/patterns/dialog-modal/examples/alertdialog.html`
to:
`content/patterns/alert-dialog/examples/alertdialog.html`

Of course this will also require revising the href of links from and to the alert dialog example page.

mcking65:
Included in #2418 via #2502

mcking65:
Resolved by #2502

PR #2417, opened 2022-07-25
Create a patterns directory tree with a separate file for each pattern and move all examples inside the patterns directory tree

mcking65:
Resolves #2418. The plan for this PR is described in #2418.

#### Commit Description for Final Squash and Merge
- Moves the HTML source for each pattern from a section in /aria-practices.html into its own file in a directory for the pattern, e.g., `patterns/button/button-pattern.html`.
- moves all example files from the /examples directory into an examples subdirectory under its associated pattern, e.g., `patterns/button/examples/button.html`.
- Moves the HTML source for each practice from a section in /aria-practices.html into its own file in a directory for the practice, e.g., `practices/grid-and-table-properties/grid-and-table-properties-practice.html`
- Moves the HTML source for each section of the about page from a section in /aria-practices.html into `about/about.html`.
- Renames HTML files that were named "index.html" to a more appropriate name, except for the page that serves as the index.
- For consistency, removes a superfluous directory from the path for two of the tree view examples.
- Makes link text more consistent, e.g., links to patterns no longer contain the word "design".
- Makes styling for warnings, advisories, and notes more consistent.
- Removes dependencies on respec for resolution of links to external resources, such as the ARIA specification.
- Revises GitHub workflows and supporting scripts to support the new content structure.

#### Review Instructions and Plan

This PR restructures all content but does not change design, function, or content. So, the objective of review is to ensure that pages look the same in preview as they do in the current production site.

Note: There are a few very minor editorial consistency changes to content. For instance, the word "design" was removed from the text of links to patterns.

We want all 90-some pages of the APG looked at briefly by 2 reviewers. With 6 reviewers, each person need look at only 30 pages, spending 2 to 3 minutes per page. If issues are not found, it should go quickly. If issues are found, please comment in this PR. You do not need to comment on the specific line in the specific file. If your assigned pages look good, please submit an approving review from the files tab.

##### Instructions
When reviewing:
* Visually scan a page in [preview](https://deploy-preview-173--aria-practices.netlify.app/aria/apg/), scan the same page in [production](https://www.w3.org/WAI/ARIA/apg/), and look for any presentation differences.
* On example pages, using a mouse, perform a sniff test of the example's pointer functionality.
* Randomly test a few links; if possible, test links to different types of targets, e.g., other APG pages, ARIA spec references.

That's it. It's a sanity check. We have run automated testing of links and the regression tests on all examples.

#### Reviewer page assignments

Bucket 1: (@jongund and @curtbellew )
[...]

PR #2502, opened 2022-10-06
Move alert-dialog example into alert-dialog pattern (Issue 2418 Task)

alflennik:
Extracts the alert-dialog example from its current position under the dialog-modal pattern directory to the alert-dialog pattern directory. I simplified the source code and removed any dependencies on code in the dialog-modal directory.

Resolves Issue https://github.com/w3c/aria-practices/issues/2488.

See Issue #2418 for more context. Note that the current PR base is https://github.com/w3c/aria-practices/pull/2499. Once that PR is merged the base of this PR should be changed to https://github.com/w3c/aria-practices/pull/2417.


mcking65 (changes requested):
I did a quick test of the example and found that the js is not behaving the same. For instance, focus is not properly trapped. There must be some methods in the dialog.js that were not copied over or replicated.

I am rather concerned that the regression testing did not capture this. We should open an issue to add regression tests that thoroughly test tab key behavior and initial focus in the dialog examples ... or fix them if the tests are there but not complete.

alflennik:
Thanks for the feedback @mcking65, the reason the regression tests did not catch the issue is that there was no regression test for this example. I wrote a regression test and fixed several issues including the one you pointed out where it's possible to tab out of the dialog. Thanks for your patience as I figure out how the examples are authored!

mcking65:
@alflennik, this is working now, and thank you for the additional tests!!!

Unfortunately, since I merged #2499, looks like we have a pile of conflicts.

alflennik:
Great to hear it's working!

PR #2511, opened 2022-10-13
Update link.html

vittorioexp:
Add explanations about "links"

### DIFF
--- a/examples/link/link.html
+++ b/examples/link/link.html
@@ -31,6 +31,7 @@
     The examples below demonstrate three variations of the
     <a href="../../#link">design pattern for link</a>.
      The link pattern is used when it is necessary for elements other than the HTML <code>a</code> element to have link behaviors.
+     If a link's purpose is not clear from its accessible name, it must be described by the link together with its preceding page context. This includes text in the same sentence, paragraph, list item, or table cell as the link, text in a parent list item, and text in the table header cell that's associated with cell that contains the link.
   </p>
   <p>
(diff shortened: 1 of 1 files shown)


mcking65:
Hi @vittorioexp,

Thank you for the suggestion.

Could you please clarify why you think the "About This Example section for the link example should include naming requirements?

I ask because this change is not consistent with the information architecture of the APG. It would thus make the link example different from other APG example pages.

The "About This Example" section is not intended for specific requirements. The requirements are in the pattern, which is linked from the about example section. On the example page, there is also an "Accessibility Features" section that can also be used to explain how the example supports certain types of accessibility requirements and best practices. For instance, the accessibility features section could explain how the name of the link in the example satisfies the WCAG naming requirements that are specific to links.

Generally, the patterns do not include specific naming requirements for all widgets. There are exceptions. For instance, in the listbox pattern, there is a caution against having many list item names that start with the same string. Since WCAG has some specific link naming requirements, there is potential for that to be an exceptional case that justifies covering naming in the pattern. Otherwise, there is an expectation that readers will refer to the APG accessible naming practice for guidance on naming.

mcking65:
@vittorioexp 

Not having received an answer to my above question, I am closing this pull request.

If this is a change that you feel is important to make, please raise an issue where we can discuss the way forward.

Since the time when you submitted this PR, the repository has been restructured. So, any changes would have to be made in a new PR.

PR #2512, opened 2022-10-13
Add note to landmarks section to allow dialogs to not be in landmark

### LINKED ISSUE #1751: Provide some exceptions in landmarks section
While https://www.w3.org/TR/wai-aria-practices-1.1/#general-principles-of-landmark-design effectively encourages landmark usage it occasionally encourages usage which adds noise to the page

An example of this is a page (which uses landmarks effectively) which launches a modal dialog. That modal dialog is outside of the landmarks of the page (as expected) but does not need to contain a landmark itself. This should be clarified in the landmarks section.

jnurthen:
fixes #1751

### DIFF
--- a/aria-practices.html
+++ b/aria-practices.html
@@ -3637,6 +3637,9 @@ Some JavaScript and CSS may not function correctly in Internet Explorer.
 
         <li>Landmark roles can be nested to identify parent/child relationships of the information being presented.</li>
       </ul>
+      <p class="note" title="Dialogs do not need to be landmarks">
+        Modal <code>dialog</code> and <code>alertdialog</code> areas do not need to be part of a landmark as these areas will be perceivable areas in their own right.
+      </p>
 
       <p id="aria_lh_step3"><strong>Step 3: Label areas</strong></p>
 
--- a/examples/landmarks/css/common.css
+++ b/examples/landmarks/css/common.css
@@ -138,6 +138,7 @@ aside {
   padding: 0.5em;
   margin-top: 0.5em;
   margin-bottom: 1em;
+  background-color: #f4f4f4;
   margin-right: 12em;
   width: 100%;
   border-radius: 6px;
--- a/examples/landmarks/css/common.css
+++ b/examples/landmarks/css/common.css
@@ -138,7 +138,6 @@ aside {
   padding: 0.5em;
   margin-top: 0.5em;
   margin-bottom: 1em;
-  background-color: #f4f4f4;
   margin-right: 12em;
   width: 100%;
   border-radius: 6px;


alflennik:
Hi @mcking65, I investigated the notes issue a bit.

In the aria-practices source code, notes are written in the following way. This is taken from the combobox pattern page.

```
<p class="note">Unlike other combobox popups, dialogs do not support <code>aria-activedescendant</code> so DOM focus moves into the dialog from the combobox.</p>
```

After Respec is run, the note HTML is as follows. This is how it is formatted on the production site.

```
<div class="note" role="note" id="issue-container-generatedID-4">
  <div role="heading" class="note-title marker" id="h-note-4" aria-level="5">
    <span>Note</span>
  </div>
  <p class="">Unlike other combobox popups, dialogs do not support <code>aria-activedescendant</code> so DOM focus moves into the dialog from the combobox.</p>
</div>
```

The main benefit of Respec's changes seems to be the addition of a heading with the word "Note." Without that heading the note would not be distinguished from the other content. The "Note" heading is always an h5.

For what it's worth, my recommendation moving forward is to go through the existing notes (there seems to be 42 of them) and format them like this:

```
<div class="note">
  <h3>Note</h3>
  <p>Unlike other combobox popups, dialogs do not support <code>aria-activedescendant</code> so DOM focus moves into the dialog from the combobox.</p>
</div>
```

There are two benefits over the existing notes. One is that the notes can more easily be given descriptive headlines (without knowing Respec tricks) and the other is that the headline level can accurately match its hierarchy inside the page.

I would be curious of any opinions about whether the attribute `role="note"` would be worth including as well.

I tested this inside the template and with some minor styling changes it will behave identically to the current notes both visually and in NVDA and VoiceOver.

Finally thank you for pointing out the preview link issue and I will look at it with @howard-e.

jnurthen:
> The main benefit of Respec's changes seems to be the addition of a heading with the word "Note." Without that heading the note would not be distinguished from the other content. The "Note" heading is always an h5.

This is not correct - notes are always 1 level below the preceding heading. They are not always at level 5.

howard-e:
@mcking65 said:

> * Why is there not a netlify preview for this PR?

The error that was reported by the [wai-aria-practices action](https://github.com/w3c/wai-aria-practices/actions/runs/3245814321/jobs/5323820500) during the `Update site files` step was:
<details>
<summary>Error: no such file or directory, open '/home/runner/work/wai-aria-practices/wai-aria-practices/ARIA/apg/example-index/js/skipto.js'</summary>

```sh
[Error: ENOENT: no such file or directory, open '/home/runner/work/wai-aria-practices/wai-aria-practices/ARIA/apg/example-index/js/skipto.js'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/home/runner/work/wai-aria-practices/wai-aria-practices/ARIA/apg/example-index/js/skipto.js'
}
```
</details>

@jnurthen @alflennik @mcking65 I believe this was resolved by #2402 being merged so this PR needs a rebase onto `main` for the preview link to properly work.

Note that build error messages will also start to be reflected in the PRs now that #2503 has been merged (new and current PRs must have that commit) and once https://github.com/w3c/wai-aria-practices/pull/157 has been merged.

mcking65:
Replaced by #2818

PR #2518, opened 2022-10-24
Checkbox Example (Two State): Activate checkbox only on keyup

### LINKED ISSUE #2425: Checkbox examples should not repeatedly toggle while holding down spacebar
I notice that in [Checkbox Example (Two State)](https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox.html), focusing a checkbox and holding down the spacebar key repeatedly toggles the checkbox. This does not happen with [native HTML checkboxes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox). I wonder if this could be an issue for people with motor disabilities.

If my intuition is correct and this is an issue, the fix should be simple - we just need to add an [event.repeat](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat) check in the [onKeyDown listener](https://github.com/w3c/aria-practices/blob/45e9a6c4b19ea03356a9aa1cda8c8eaaf3272015/examples/checkbox/js/checkbox.js#L35).
_If we really care about parity with native HTML checkboxes, this logic should probably be done in a `keyup` listener, though._

sivakusayan:
Fixes #2425.

I chose to move this logic to a `keyup` handler to maintain consistency with HTML checkboxes. If this fix is okay, we should probably make a similar fix to the [Mixed Checkbox example](https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox-mixed.html).

### DIFF
--- a/examples/checkbox/js/checkbox.js
+++ b/examples/checkbox/js/checkbox.js
@@ -33,7 +33,7 @@ class Checkbox {
 
   /* EVENT HANDLERS */
 
-  // Make sure to prevent page scrolling on spacebar down
+  // Make sure to prevent page scrolling on space down
   onKeydown(event) {
     if (event.key === ' ') {
       event.preventDefault();
--- a/examples/checkbox/js/checkbox.js
+++ b/examples/checkbox/js/checkbox.js
@@ -55,7 +55,6 @@ class Checkbox {
 
     if (flag) {
       event.stopPropagation();
-      event.preventDefault();
     }
   }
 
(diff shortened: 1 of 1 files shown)


mcking65:
@howard-e @alflennik 

Note the failure of workflow
Trigger PR update for WAI-APG site / create-pr-wai

jongund (approved):
I tested the changes and it works great.
I think "spacebar" in the documentation needs to change to "space" to pass the spelling rules

sivakusayan:
Fixed the spell check 🙂 
I also removed the `event.preventDefault()` call in the new `keyup` handler since I realized it's no longer needed. I don't think browsers do anything when the spacebar is released — I'm guessing it was there to prevent the scrolling issue when it was previously a `keydown` handler.

PR #2521, opened 2022-10-31
Move about page and read me first out of aria-practices

mcking65:
@alflennik, if this is helpful, we can merge to move-examples. If you'd rather wait till after move-examples is merged, we can rebase later.

There were only these last bits of content left in aria-practices.html, so this PR completely removes aria-practices.html. We might also be able to delete the respec config?

This PR moves the following from aria-practices.html into the new content tree:
* Moves Introduction to content/about/about.html
* Moves read me first to content/practices/read-me-first/read-me-first-practice.html
* Moves change history to content/about/about.html
* Moves acknowledgements to content/about/about.html
* Puts a copy of generated references content into content/about/about.html


mcking65:
@alflennik

It appears that content/shared/js/notice.html is not being used in production because the copy is different in production. Please make sure the build script consumes the copy from that file.

I updated the hrefs to point to correct location based on new location of read-me-first content under practices. I did so assuming that the relative path needs to be from the location of an example file where the content is included.

I thought that the include was done by one of the scripts that runs on example pages. However, the warning disclosure is not rendered when loading an example file in this repo. So, I can't really test the links to the read first content.

Shouldn't one of those scripts be doing the include?

BTW, "notice.html" is not a very descriptive name, and it doesn't seem like copy should be located in a js directory.
* Should there be a content/shared/templates directory?
* What do you think of renaming notice.html to example-usage-warning.html?

alflennik (approved):

PR #2532, opened 2022-11-02
Move-Examples: Fix notes and warnings

mcking65:
Reformat instances of `class="note"` and `class="warning"`.

For each instance of either class:
* If the class is not on a `div`:
    * Wrap the element with the class in a `div`.
    * Increase indent of the wrapped element two spaces.
    * Delete the class from the wrapped element.
    * Add the class to the `div`.
* In the the `div` with the class, add a heading as the first child:
    * Make the heading level one deeper than the containing section.
    * If the class is "note", set heading text to "Note".
    * If the class is "warning", set heading text to "Warning".

### DIFF
--- a/content/patterns/button/button-pattern.html
+++ b/content/patterns/button/button-pattern.html
@@ -34,13 +34,16 @@
           </li>
           <li>Menu button: as described in the <a href="../menu-button/menu-button-pattern.html">menu button pattern</a>, a button is revealed to assistive technologies as a menu button if it has the property <a href="#aria-haspopup" class="property-reference">aria-haspopup</a> set to either <code>menu</code> or <code>true</code>.</li>
         </ul>
-        <p class="note">
-          The types of actions performed by buttons are distinctly different from the function of a link (see <a href="../link/link-pattern.html">link pattern</a>).
-          It is important that both the appearance and role of a widget match the function it provides.
-          Nevertheless, elements occasionally have the visual style of a link but perform the action of a button.
-          In such cases, giving the element role <code>button</code> helps assistive technology users understand the function of the element.
-          However, a better solution is to adjust the visual design so it matches the function and ARIA role.
-        </p>
+        <div class="note">
+          <h3>Note</h3>
+          <p>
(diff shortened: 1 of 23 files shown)


mcking65:
@jongund,

Here is the draft PR for the notes and warnings changes we need to make. Thank you very much for volunteering to help with this!!

Since the goal is to have the move-examples branch ready for review by Nov 8, we will want to merge this into move-examples before then.

I provided detailed instructions in the top comment. I have also started the work. So, in the first 2 commits, you have several examples of the needed changes to model.

jongund:
@mcking65 
The file "notice.html" seems to be a special case of note, since it is a standalone file with the class=note.
What should be done with that file.

jongund:
@mcking65 @alflennik 
I have completed the changes for notes and warnings, except I was not sure what to do with the `notice.html` file, since it is using `deatils/summary` tags.  So maybe it is ok the way it is.

mcking65 (commented):
thank you @jongund!

You are right that notice.html is a special case and it did not need to be changed ... I don't think. Alex will need to figure that out.

@alflennik please and review and merge if all looks good to you.

alflennik (changes requested):
Thanks for making all these changes @jongund! 

One thing giving me pause right now is a couple of aspects of the styling, for example the notes currently have an empty space at the top and the warnings don't have any styling currently. Would it be possible to add some styling to `shared/css/core.css` which makes the appearance similar between notes and warnings? (And possibly advisements too, see below.)

I looked at the production site and found that the styling was not consistent there either, so there is an opportunity here to improve the visual consistency on the live site.

And then as I was digging through the warnings I encountered `class="advisement"` which felt like the APG has two names for idea of a warning ...? Correct me if I'm wrong. It feels to me like advisements and warnings should be combined, or, if we think they are different concepts maybe I'm just saying that advisements should also be formatted in the same way as notes and warnings.

As for the notice.html file, I took care of that in another PR.

Thanks again for all that manual work getting things consistent.

alflennik:
I decided to merge these changes and address the CSS fixes in the build-fixes branch. Thanks for the help Jon!

PR #2541, opened 2022-11-15
Checkbox Example (Mixed-State):  Activate checkbox only on keyup

sivakusayan:
This is a follow-up PR to #2518 where [Checkbox Example (Two State)](https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox.html) was changed to only toggle on keyup. I think it makes sense to make the same change to [Checkbox Example (Mixed-State)](https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox-mixed.html) for consistency.

### DIFF
--- a/content/patterns/landmarks/landmarks-pattern.html
+++ b/content/patterns/landmarks/landmarks-pattern.html
@@ -1,0 +1,67 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta content="width=device-width, initial-scale=1.0" name="viewport">
+    <title>Landmarks Pattern</title>
+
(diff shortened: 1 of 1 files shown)


mcking65:
@sivakusayan 

Thank you for this contribution! Unfortunately, we need to hold off for a few weeks because we are in the midst of restructuring the entire repository to better support the current site design, and we need to temporarily stabilize the main branch.

Would you mind rebasing this PR after we merge #2417, which I expect will be between December 1 and December 9?

sivakusayan:
Yeah, I don't mind resubmitting this, I'll do that sometime this week 🙂

sivakusayan:
Woops, not sure how this closed. In any case, the [new PR is here](2722).

PR #2543, opened 2022-11-16
Link checker

alflennik:
Adds a link checker which will run in CI. It can also be run locally with `npm run link-checker`.

- In addition to anchor links, it also checks stylesheet, script and image links.
- In addition to checking the HTTP status, it also checks that hash links are pointing to an id on the page which exists.
- It supports the link rewriting which is done for spec links.
- There is a configuration file (".link-checker.js") which supports excluding links on a particular page with a CSS selector (useful for some examples) and a couple other controls.

Huge thanks to @Paul-Clue at Bocoup who worked with me on this feature from beginning to end.


nschonni:
See https://github.com/w3c/aria-practices/pull/2147 for a simpler solution than a custom script

mcking65 (approved):
LGTM, thank you!

@nschonni, did you review the list of things being tested? My understanding was that #2147 does not support all the types of references we need to test.

nschonni:
https://github.com/JustinBeckwith/linkinator seems to have all that. You could also use the extra logic in the `linksToSkip` for the extra skipping you found was needed. My PR was against the current main, and I haven't tried rebasing it against this branch.

alflennik:
Apologies @nschonni I didn't realize you had a PR open which covered the same feature that Paul and I were working on. If we knew that we would have touched base earlier.

We did try Linkinator as we were authoring the PR, but we were a bit irked that it did not check hash links ([see issue](https://github.com/JustinBeckwith/linkinator/issues/79)), which is a limitation off-the-shelf link checkers seem to have, and this was the initial reason that led us to decide to implement a custom link checker.

nschonni:
No problem at all! My main point with other PR was just to leverage as much externally supported code. Only suggestion might be to leverage Linkinator (or something else) if it covers 80% of the cases, to minimize the amount this project has to maintain

mcking65:
I suggest we satisfy our immediate need with this custom script.

Perhaps as a long term maintenance strategy, could we fork linkenator, add missing capability, offer it back in a PR, and then opefully be able to use it in the future? Would be interested in sizing that next year.

Issue #2520, opened 2022-10-26
Mockups for Assistive Technology Support table in Patterns Example's pages


isaacdurazo:
These mockups show how we could display an Assistive Technology Support Table (also known as test reports) from ARIA-AT on an APG Pattern Example page. My intention is to start a discussion to decide what (if not all) of these features we want to implement. 

My thinking is we should start simple, but wanted to include as many possibilities as I could come up with so we have an idea of an end goal.

### Visual Version
![Embedded Test Results](https://user-images.githubusercontent.com/1379244/200435777-3f56b3fa-9721-4132-bfe8-669fa8c8c59d.png)

### Text-based version
A new section with an H2 that reads "Assistive Technology Support" has been added at the end of the page, right after the "HTML Source Code" section. This section has a disclosure component after the heading, indicating the phase in which the test report is. The phase could be Candidate or Recommended. After the disclosure component, we have a table that displays the actual test report from its highest-level view, which has Assistive Technologies with their major version as rows and Browsers as columns. 

For the cells where an AT and a browser intersect, we display a progress bar showing the percentage of passing assertions with a text underneath that reads "X% of passing assertions". When we have no test results to display, we show the word "None".

The cells in the test report that display the progress bar, are clickable. Clicking them will take the user to the ARIA-AT website, specifically to the summary report table for that particular AT and Browser combination they have clicked.

Under the Test Report table, there's a primary button called "View Complete Report", which will take the user to ARIA-AT's main report page for the Pattern Example they are looking at in APG, where they can see the Test Reports for all AT and browser combinations 

Next to the "View Complete Report" button, there's a secondary one called "Embed". Clicking it will open the "Embed Pattern Test Report " modal. See the next mockup for more details.

### Visual Version
![Embed Modal](https://user-images.githubusercontent.com/1379244/198105683-b3ad6e10-8955-4221-8b86-1c5a59784894.png)

### Text-based version
The "Embed Pattern Test Report modal" has a heading that reads "Embed [pattern name] Test Report" followed by the test report table for the pattern example the user is looking at. Underneath the table, the user has different filters to hide or show data in the table, copy the code and embed the test report somewhere else.

On the left-hand side, the embed filters are split up into two lists:

1. Filter Assistive Technologies. Which is a checklist of all ATs for the user to hide or show.
2. Filter Browsers. Which is a checklist of all Browsers for the user to hide or show.

[...]

PR #2569, opened 2022-12-21
Button Example: Add AT support tables from ARIA-AT

alflennik:
First implementation of #2520

See https://github.com/w3c/aria-at-app/pull/476 for the PR which adds support for ARIA-AT support level iframes.

This PR should not be merged until that branch has been deployed to production, because the iframe included in this PR is still referring to the ARIA-AT sandbox, not production.

[Preview Button Example Page](https://deploy-preview-206--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/)

___
WAI Preview Link [failed to build](https://github.com/w3c/wai-aria-practices/runs/12316080194?check_suite_focus=true) on 'Update site files' step. _(Last tried on Mon, 27 Mar 2023 21:42:40 GMT)._


mcking65:
https://deploy-preview-180--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/#supportlevels

gives me

>502 Bad Gateway

Under the Support Levels heading.

content/patterns/button/examples/button.html
```diff
@@ -195,6 +195,18 @@ <h2 id="sc1_label">HTML Source Code</h2>
           sourceCode.make();
         </script>
       </section>
+
+
+      <section>
```
mcking65, line 200:
```suggestion
      <section id="at-support">
```

content/patterns/button/examples/button.html
```diff
@@ -195,6 +195,18 @@ <h2 id="sc1_label">HTML Source Code</h2>
...
+        <h2>Support Levels</h2>
```
mcking65, line 201:
```suggestion
        <h2>Assistive Technology Support</h2>
```

mcking65 (commented):
Suggested minor changes based on issue #2520

content/patterns/button/examples/button.html
```diff
@@ -195,6 +195,18 @@ <h2 id="sc1_label">HTML Source Code</h2>
...
+        <iframe
```
mcking65, line 202: Need an H3 here for "Command Button"

content/patterns/button/examples/button.html
```diff
@@ -195,6 +195,18 @@ <h2 id="sc1_label">HTML Source Code</h2>
...
+          class="support-levels-command-button"
```
mcking65, line 203: Curious about the name of this class. I didn't look at the class definition, but assume that it could be used for all AT support iframes so should not include -command-button?

content/patterns/button/examples/button.html
```diff
@@ -195,6 +195,18 @@ <h2 id="sc1_label">HTML Source Code</h2>
...
+          src="https://aria-at-app-sandbox.bocoup.com/embed/reports/command-button"
+          height="500"
+          allow="clipboard-write"
+          style="border-style: none; width: 100%;">
+        </iframe>
+      </section>
```
mcking65, line 209: Need an h3 for "Toggle Button" followed by another iframe for that table.

mcking65 (commented):
We need a table for command button and a table for toggle button.

Now I recall this is why we decided to do the link page before the button page ... we need the JAWS 2023 results before we can publish toggle button. Nevertheless, let's get the toggle results into this PR. We can have a separate PR for the link page.

content/patterns/button/examples/button.html
```diff
@@ -195,6 +195,18 @@ <h2 id="sc1_label">HTML Source Code</h2>
           sourceCode.make();
         </script>
       </section>
+
+
+      <section>
+        <h2>Support Levels</h2>
+        <iframe
+          class="support-levels-command-button"
+          src="https://aria-at-app-sandbox.bocoup.com/embed/reports/command-button"
```
howard-e, line 204:
```suggestion
          src="https://aria-at.w3.org/embed/reports/command-button"
```
As https://github.com/w3c/aria-at-app/pull/476 has been merged and available in production.

This should also address @mcking65 comment that:

> https://deploy-preview-180--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/#supportlevels gives me
> >502 Bad Gateway

As the sandbox environment was experiencing performance issues.

content/patterns/button/examples/button.html
howard-e: This seems to be needed. Based on the usage from the ARIA-AT App at [aria-at-app/server/handlebars/public/script.js L30:L34](https://github.com/w3c/aria-at-app/blob/1885ac1fce9c38a0889801b6d1a6d0f8056805b3/server/handlebars/public/script.js#L30:L34), it posts a message to the APG example webpage for each `iframe` (with their unique class name) that's used to maintain a fixed height which prevents a scrollbar on the `iframe`s. This is seen at [aria-practices/content/shared/js/app.js L51:L60](https://github.com/w3c/aria-practices/blob/575926a48cb14d8497ff3bdba1889d20791b1c6e/content/shared/js/app.js#L51:L60).

It could also be general class name such as `support-levels-pattern` loop through those elements to fix the height but it would mean that when multiple `iframe`s are included, some would be evaluated again as it happens for each `iframe` that's included.

content/patterns/button/examples/button.html
howard-e: Addressed in https://github.com/w3c/aria-at-app/pull/479.

content/patterns/button/examples/button.html
howard-e: Toggle Button has been added

howard-e:
@mcking65 the feedback has been addressed and should now be available at https://deploy-preview-180--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/#supportlevels.

mcking65 (changes requested):
I'm adding this to the agenda for discussion during the January 17 APG TF meeting.

I propose we make the following changes:
1. Change warning copy as described below.
2. Make the warning a disclosure like it is on the aria-at site.
3. Change "None" in table cells to "No Data". The word none implies there is no support. While that is the case for JAWS or NVDA with Safari, it is not the case for VoiceOver with Chrome.
4. Order rows and columns alphabetically.
5. Change the accessible names of the buttons:
    1. Change "View Complete Report" to "View Complete Report in New Window"
    2. Change "Embed" to "Copy Embed Link"

#### Proposed changes to warning copy

The warning is meaningfully different from the production aria-at  site and has a gramatical error. Also, reading it in context, I now realize that it is not very clear about the providence of the data. This is particularly problematic if someone uses the embed link in another context.

The [copy in this PR preview](https://deploy-preview-180--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/#assistivetechnologysupport) is:

>Candidate Report
>The information in this report generated from candidate tests. Candidate aria-at tests are in review by assistive technology developers and lack consensus regarding:

The [aria-at pprod site copy](https://aria-at.w3.org/report/1097) copy is:

>Warning! Unapproved Report
>The information in this report is generated from candidate tests. Candidate ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:

I propose that we:
1. Use the language "Warning! Unapproved Report" instead of "Candidate Report"
2. Add the missing word "is" before the word "generated".
3. Add language that describes the source of the data as the aria-at project with a link to the aria-at home page.

My proposed copy is:

>Warning! Unapproved Report
>The information in this report is generated from candidate tests developed and run by the [ARIA-AT Project](https://aria-at.w3.org/). Candidate ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:

content/patterns/button/examples/button.html
mcking65: Ah, understand now. thank you.

jscholes:
Accessibility feedback:

1. The iframes don't have accessible names, via a `title` attribute.
    * Note: when the iframe embed code is copied to the clipboard, the result should also have an accessible name.
3. The first text within each iframe, e.g. Command Button" and "Candidate Report", are marked up as separate level 3 headings.  There should either be a single heading of "Command Button Candidate Report" or similar, or "Candidate Report" shouldn't be a heading at all.
4. The first column of the table, with the assistive technology names and versions, doesn't have a column header.
5. The control to navigate to the full report would be more usable/accurate as a link, e.g. which would allow people to easily copy the URL to the clipboard.  (in addition to the wording updates already suggested by @mcking65).
6. The "Embed" button doesn't copy a link to the clipboard, it copies the mark-up for embedding the full iframe.  As such, the label should be updated in line with @mcking65's feedback, but not using the word "link".  E.g. "Copy embed code".
    * Note: the wording of the resulting alert should also be updated accordingly, to reference embed code rather than a link.

mfairchild365:
In the ARIA-AT CG call on 1/19, we discussed the wording of the cells. Currently, it says something like "100% of passing assertions". We agreed that it would be clearer to say "100% support" (or whatever the percentage is).

jscholes:
Additional accessibility note: when reviewing the cells in the first column of the table, the assistive technology name and version are on separate lines of the virtual buffer with NVDA.  When reading sequentially, users have to read the name, then arrow down to read the version.

mcking65:
Feedback from Joe Humbert: the tables have both horizontal and vertical scroll bars when viewed in Firefox but nothing is scrollable b/c all content is visible. Can they be suppressed?

The "None" text does not pass CCR reqs.

The table border colors  are not as good contrast as other tables on the page.

Cell text: change to X% supported

jscholes:
Technical notes:

1. The embed URL includes a `title` parameter, e.g. `title=Command+Button`.  Changing the string changes the heading text at the top of the embedded report, which seems like an unnecessary use case that could result in misleading embeds if someone simply calls their embedded report whatever they want.
2. On a similar note, it would be wonderful for the embed URL to include a parameter for setting the base heading level.  It's currently set to 3 because that matches the APG page layout, but that won't be the heading level that's appropriate for all embedded contexts.

jscholes:
Final comment (I think!): I ran the page through Axe, and it noted that the same element ID, `at-version`, is being used on several elements.  I don't think this has a user-perceivable impact, but valid mark-up is just a good idea.

mcking65:
@jscholes can you explain the advantage or benefit of adding a title attribute on the iframe? Generally, I think it is better if iframes do not have an accessible name. To me, it seems like it is a much better experience if the screen reader experience is more like the visual experience where the iframe boundaries are not distracting.

During the APG meeting, we discussed link verses button for the view full report element. While there are some marginal advantages to a link with an href, there are also some visual presentation considerations. We don't want to style a link like a button nor a button like a link. The copy embed code element needs to be a button. Several members of the task force prefer the current design with two buttons over exploring a design with a link and a button. Since the benefits of the change are so minimal, I would prefer to keep the focus on all the highest priority changes. We can revisit this later.

mcking65:
@howard-e @alflennik 

We collected a fair bit of feedback from both the APG TF meeting and the ARIA-AT CG meeting this week. Most of it is represented in the comments above. However, to ensure we have a clear and prioritized list of changes for you to work on, I'm going to consolidate and prioritize feedback in this comment.

#### Requested Changes
1. Increase the color contrast ratio of the text in the cells that contain the word "None" to the required 4.5 or better.
1. Change copy in cells with data to "%X Supported", e.g. "96% Supported".
1. In cells that contain the word "None", change copy to one of the two options described below.
1. Change copy of the text above the table (warning text) as described below.
1. Remove heading markup from the warning and make it a disclosure that is collapsed on load like it is on the aria-at site.
1. Change the accessible name of the button "View Complete Report" to "View Complete Report in New Window".
1. Change the visible name of the button "Embed" to "Copy Embed Code" and revise the confirmation announcement to be "Copied code for embedding".
1. Order rows and columns alphabetically.
1. Increase the color contrast of the table borders (make them match other tables on the page).
1. If possible, suppress the visibility of the table scrollbars when they are not needed.
1. Remove the embed parameter that allows people to change the heading text.
1. Add an embed parameter that enables people to specify a heading level.

#### Proposed changes to warning copy

The warning is meaningfully different from the production aria-at  site and has a gramatical error. Also, reading it in context, I now realize that it is not very clear about the providence of the data. This is particularly problematic if someone uses the embed link in another context.

The [copy in this PR preview](https://deploy-preview-180--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/#assistivetechnologysupport) is:

>Candidate Report
>The information in this report generated from candidate tests. Candidate aria-at tests are in review by assistive technology developers and lack consensus regarding:

The [aria-at pprod site copy](https://aria-at.w3.org/report/1097) copy is:

>Warning! Unapproved Report
>The information in this report is generated from candidate tests. Candidate ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:

Requested changes:
1. Use the language "Warning! Unapproved Report" instead of "Candidate Report"
4. Add the missing word "is" before the word "generated".
5. Add language that describes the source of the data as the aria-at project with a link to the aria-at home page.

Requested new copy:

>Warning! Unapproved Report
>The information in this report is generated from candidate tests developed and run by the [ARIA-AT Project](https://aria-at.w3.org/). Candidate ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:

Note: as requested above, "Warning! Unapproved Report" should not be a heading. Instead, it should be summary text in a summary/details or otherwise made into a disclosure button label.

#### Options for cells that do not contain data
We have two reasons for a cell to not have data:
1. We have not yet tested the AT/Browser combination, e.g., JAWS/Firefox.
2. The AT/Browser combination does not exist, e.g., JAWS/Safari.

If it is readily feasible to have a different string for each case, please use the following Option A for the copy, otherwise use Option B.

Option A for the copy in cells with no data is:
1. If the AT/Browser combination will be tested in the future, use the string "Data Not Yet Available"
2. If the AT/Browser combination does not exist, use the string "Not Applicable".

Option B for the copy in cells with no data is "No Data".

jscholes:
@mcking65 

> can you explain the advantage or benefit of adding a title attribute on the iframe?

Iframes are user-addressable containers that are conveyed by the screen reader, and can be directly navigated to.  In the case of this preview page, there are also two of them, without any means of telling them apart without reading the content.

Hence, in line with https://www.w3.org/WAI/WCAG21/Techniques/html/H64, I would currently consider these frames to be failing against [4.1.2 Name, Role, Value (level A)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value) were I auditing the page.

mcking65:
@jscholes wrote:

> Iframes are user-addressable containers that are conveyed by the screen reader, and can be directly navigated to. In the case of this preview page, there are also two of them, without any means of telling them apart without reading the content.
> 
> Hence, in line with https://www.w3.org/WAI/WCAG21/Techniques/html/H64, I would currently consider these frames to be failing against [4.1.2 Name, Role, Value (level A)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value) were I auditing the page.

James, I don't understand your interpretation. My reading of 4.1.2 and H64 lead me to a different conclusion. I wonder what you think of the following line of reasoning?
how you are interpreting 4.1.2.

From the applicability section of [H64: Using the title attribute of the iframe element | WAI | W3C](https://www.w3.org/WAI/WCAG21/Techniques/html/H64):

>This technique relates to:
>* 2.4.1: Bypass Blocks (Sufficient, together with H70: Using frame elements to group blocks of repeated material when used with Grouping blocks of repeated material in a way that can be skipped, using one of the following techniques: ) 
>* 4.1.2: Name, Role, Value (Sufficient when used with G108: Using markup features to expose the name and role, allow user-settable properties to be directly set, and provide notification of changes)

In my reading of the H64 page, we are not using frames in a way where the technique is applicable. We are not using frames to group blocks or structure the page content. For all practical purposes, the iframes are exactly equal in function and semantics to the section and div elements that wrap other content on the page. We do not intent the frames to have any accessibility semantics.

The default role of an iframe is presentation in most browsers. Giving the iframe a name overrides role presentation. The iframe element is similar to the section element in this way. The section element has role generic by default. However, if you name a section, it becomes a landmark region element.

We have not named the section elements because we don't want to overload the page with landmark regions; we have a well-defined document information architecture that is conveyed by headings. If we were to name the sections, it would reduce users ability to perceive the overall architecture of the page and navigate it efficiently.

Similarly, I believe we want the iframes to be presentational because they fit into that same document information architecture, which is communicated via headings. Naming the iframes would disrupt the information architecture without adding navigation value. This will be especially the case when we move the H3 elements outside the iframes. I've requested that the table be named. The consumer of the iframe, especially the APG pages, should be in full control of the heading structure and content. On pages that have only one example, we do not want an H3 heading. We just want an appropriately named table. This will match the structure of the rest of the page.

jscholes:
@mcking65 

> For all practical purposes, the iframes are exactly equal in function and semantics to the section and div elements that wrap other content on the page.
> ...
> The default role of an iframe is presentation in most browsers. Giving the iframe a name overrides role presentation. The iframe element is similar to the section element in this way.

I think this is the crux of the issue.  If that is how iframes function and are/are not exposed to users, I agree there is no need for the containers to have an accessible name.  However, while I did not carry out exhaustive testing:

* NVDA, with both Chrome and Firefox, does expose the frame boundaries to users, despite it having no name.
* JAWS, with the same browsers, does not expose the frame semantics while navigating sequentially, but users can jump to the content of the frame by pressing M/Shift+M.

We could consider this to be a quirk of NVDA, and leave it as is.  But both JAWS and NVDA do at least recognise the role of the frame in some way, and so I'm not sure the browsers are even following the rules as you described them.  Unless the browsers are giving the frame a generic/presentational role, and JAWS/NVDA are detecting the semantics in some other fashion.

alflennik:
@mcking65 A preview including fixes for the above items is available here: https://deploy-preview-206--aria-practices.netlify.app/

Note that item 3 has not been fixed - we discussed it previously at an APG meeting and decided that if it was too heavy a lift it would be preferable to put it aside and implement the other fixes first.

mcking65:
> Note that item 3 has not been fixed - we discussed it previously at an APG meeting and decided that if it was too heavy a lift it would be preferable to put it aside and implement the other fixes first.

We have two options for item 3: option A and option B. We agreed on option B if option A is too heavy a lift for now.

Per option B described above, The copy should be "No Data" instead of "None".


Also, the warning text is missing some words. Please double check against the copy listed above. The phrase before the list is missing the word "Candidate".

alflennik:
@mcking65 I saw you added headers for each of the iframes, thank you for fixing that. Also I think you were asking about why the preview didn't update automatically: the reason is that this feature requires changes across both the aria-practices and wai-aria-practices repositories, and the automated preview system is intended for changes to aria-practices which don't require changes in wai-aria-practices.

The issues you mentioned have been fixed, take a look here: https://deploy-preview-206--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/

Additionally, following our conversation in the task force meeting last week I added support for recommended reports in addition to candidate reports.

mcking65:
@alflennik 

The two issues I raise in [this comment above](https://github.com/w3c/aria-practices/pull/2569#issuecomment-1462670773) are not addressed.

1. In the tables, it still says "none" instead of "no Data"
2. Above the table, when the disclosure is expanded, the copy says "ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:". It is missing the word "Candidate". It should say "Candidate ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:"

Neither of the above issues is visible at this moment because the iframe is returning:

>Error: ENOENT: no such file or directory, open '/home/aria-bot/aria-at-report/handlebars/views/main.hbs'

charmarkk (commented):
- There are some wrapping issues with the first bullet in the "About This Example" section. Shrinking the screen eventually fixes it, until the mobile breakpoint is triggered and it happens again. The bullet is slightly overlapping the button icon.

image

- There is no visual focus indicator on the example Print Page and Mute buttons - should there be?

- The text wrapping/breaks in the HTML Source Code section is really odd and hard to visually parse. It looks like it's similar to most other examples, but because there are lots of single words and elements not strictly related to ARIA (notably \<q\> and \<code\>) it it is a little hard to read it. Should that be looked at? 
image

charmarkk:
@alflennik fair! Can do.

With respect to the tables, I can +1 @mcking65's previous comments, too.

alflennik:
@mcking65 I believe the issues you saw have been fixed, could you check once more? Thank you!

alflennik:
@mcking65 I've updated this PR and the [website PR](https://github.com/w3c/wai-aria-practices/pull/206) to use the production version of ARIA-AT following our deploy of that feature today. It can be previewed here: https://deploy-preview-206--aria-practices.netlify.app/aria/apg/patterns/button/examples/button/

mcking65 (approved):

PR #2613, opened 2023-02-13
Infrastructure: Move to Stylelint 15

nschonni:
https://stylelint.io/migration-guide/to-15/ recommends migrating to Prettier for all the formatting parts directly. This updates the related hooks and dependencies to implement that approach, along with fixing some new minor violations

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,6 @@
 # Ignore external copied w3c files
 common
+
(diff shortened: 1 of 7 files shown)


howard-e (approved):
Whoa [76 rules deprecated](https://stylelint.io/user-guide/rules#deprecated)!

Looks good to me

PR #2614, opened 2023-02-13
Infrastructure: Run Prettier on commit of all files

nschonni:
Now that all the linting relies on prettier, the commit hook can now run on all filetypes.
Ran `npx prettier -w .` and ignored some filetypes like HTML and 3rd party code.
This doesn't add any specific CI to double check for regressions though for those that don't commit with the Husky hooks enabled


.prettierignore
```diff
@@ -4,3 +4,12 @@ common
 # Vendored theme files
 bootstrap.css
 bootstrap-theme.css
+
+# Ignore HTML formatting as the diffs are not great
+*.html
+
+# Ignore thire party JavaScript
```
nfxpnk, line 11: thir**d**

nschonni:
```suggestion
# Ignore third party JavaScript
```

howard-e:
This made me notice there are some outdated paths in the `package.json` as well. Will submit a PR for those changes

PR #2619, opened 2023-02-14
Infrastructure: Update script that provides skipto menu on all pages to version 5.1

jongund:
Changes included in SkipTo version 5.1:

* Supports including headings and landmarks in custom web components.
* Rewritten to use Javascript  modules.
* Change configuration options to allow customization of container element to be customized; default container is `nav` element.
* The APG implementation of SkipTo sets the container to `div` so it does not create a navigation landmark. 
      
___
[WAI Preview Link](https://deploy-preview-195--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 28 Feb 2023 19:04:02 GMT)._

### DIFF
--- a/content/shared/js/skipto.js
+++ b/content/shared/js/skipto.js
@@ -14,104 +14,107 @@
  * Report Issues: https://github.com/skipto-landmarks-headings/page-script-5/issues
  * ======================================================================== */
 
-/*
-*   debug.js
-*
-*   Usage
-*     import DebugLogging from './debug.js';
-*     const debug = new DebugLogging('myLabel', true); // e.g. 'myModule'
-*     ...
-*     if (debug.flag) debug.log('myMessage');
-*
-*   Notes
-*     new DebugLogging() - calling the constructor with no arguments results
-*                   in debug.flag set to false and debug.label set to 'debug';
-*                   constructor accepts 0, 1 or 2 arguments in any order
-*                   @param flag [optional] {boolean} - sets debug.flag
-*                   @param label [optional] {string} - sets debug.label
-*   Properties
-*     debug.flag    {boolean} allows you to switch debug logging on or off;
-*                   default value is false
-*     debug.label   {string} rendered as a prefix to each log message;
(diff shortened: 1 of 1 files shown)


mcking65:
@jongund 

Please edit the opening comment and list changes since V4.2. This will both help with testing and enable us to have a descriptive commit statement in the history.

mcking65 (changes requested):
There are some differences in the attributions in the copyright statements. Please more consistently mirror the language used in the prior version statement. We promised to keep it consistent.

This version:

~~~
* Copyright (c) <2022, 2023> (version 5.1) Jon Gunderson, University of Illinois and PayPal
* All rights reserved.
~~~

Prior version:

~~~
/*! skipto - v4.2.0 - 2022-06-16
 * https://github.com/paypal/skipto
 * Copyright (c) 2022 Jon Gunderson; Licensed BSD
 * Copyright (c) 2021 PayPal Accessibility Team and University of Illinois; Licensed BSD */
~~~

mcking65:
I believe it would be better if the script did not wrap the button in a navigation region. If a page author wants to put the button inside of an existing region, perhaps they can do that. I don't think the script should presume that the author wants to add another region to the page.

Personally, I find regions containing a single element to be pretty annoying. On rare occasion they become necessary because there is an important element in the middle of the page that does not belong in the prior or next region -- that is an extreme edge case and often represents undesirable design.

Because the button is the first element, it is trivial to get to it and almost impossible for screen reader users to overlook. The region is thus adding verbosity that does not provide utility.

If we really wanted it to be inside of a region, which I don't believe is necessary as it is not actual page content, we could make it the first element in the banner. However, I believe it is better that it is outside the banner. If it were persistently visible to all users, I would advocate that it be the first element inside the banner.

jongund:
@mcking65 
I updated the copyright information and removed the navigation landmark wrapping the button and menu.

mcking65 (approved):

a11ydoer (approved):

Issue #2636, opened 2023-03-01
Infrastructure: Remove temp pinning to Node 16 to work around tests failing with Node 18 in CI


howard-e:
Initially reported by @jongund through an email thread and further supported by the latest set of [dependabot updates](https://github.com/w3c/aria-practices/pulls?q=is%3Apr+is%3Aopen+dependabot), all the regression tests are failing with a similar error for each test, when ran with Node 18 in the GitHub Actions:

```
not ok 1 - accordion_accordion › before hook
  ---
    name: AssertionError
    message: Rejected promise returned by test
    values:
      'Rejected promise returned by test. Reason:': |-
        Error {
          message: 'Timed out while waiting for WebDriver server',
        }
    at: 'Timeout.poll [as _onTimeout] (test/util/start-geckodriver.js:34:16)'
  ...
```

The error message is `Timed out while waiting for WebDriver server`.

howard-e:
The difference between the recent failing PRs and the passing PRs is the passing PRs were using `node v16.x.x` while the failing ones are now making use of `node v18.x.x`, as `actions/setup-node@v3` is now using the latest LTS as we're not pinning the node version.

This issue was identified in https://github.com/SeleniumHQ/selenium/issues/10970 (ISSUE 2). The builds should pass by merging https://github.com/w3c/aria-practices/pull/2635 as this bug was addressed in [selenium-webdriver v4.5.0](https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/node/selenium-webdriver/CHANGES.md#v450) but that dependabot PR is also failing.

howard-e:
May be potentially related to a regression with how `child_process.spawn` works in CI for node 18.

howard-e:
> The difference between the recent failing PRs and the passing PRs is the passing PRs were using `node v16.x.x` while the failing ones are now making use of `node v18.x.x`, as `actions/setup-node@v3` is now using the latest LTS because we're not pinning the node version.

#2640 now pins the version back to `v16` and is a temporary fix for this, while we investigate.

mcking65:
@howard-e, will the bugs associated with Node 18 effect only CI or could it also effect local execution of tests?

howard-e:
> will the bugs associated with Node 18 effect only CI or could it also effect local execution of tests?

@mcking65 it only affects the tests when running in the GH CI from what I can see. At least, no unexpected errors have been thrown locally for me with node18.

howard-e:
[...]

PR #2640, opened 2023-03-02
Infrastructure: Bump `ava` and `selenium-webdriver`

howard-e:
See #2636.

Bump `ava` and `selenium-webdriver`.

Also _temporarily_ pin the node version to `v16` when executing the `Run regression tests` job. `actions/setup-node@v3` has recently defaulted to using `v18` but that it seems to break the regression tests running in CI in this repository. 

This should be temporary because we'd ideally want to solve what's happening for the latest node LTS version here.
___
[WAI Preview Link](https://deploy-preview-203--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 02 Mar 2023 21:58:20 GMT)._

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -35,7 +35,6 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16
           cache: npm
 
       - name: Install npm dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/regression.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v3
         with:
+          node-version: 16
```
mcking65, line 38: If this line is intended to be temporary, do we have an issue tracking the underlying issue that will remind us to remove this line?

howard-e: I've renamed #2636 to be more specific to this issue so we can continue tracking there.

PR #2653, opened 2023-03-30
Add AT support tables to example pages for alert, link, and radio using activedescendant

mcking65:
Updates three example pages to include a section for AT support:
* Alert
* Link
* Radio Using aria-activedescendant
___
[WAI Preview Link](https://deploy-preview-209--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 30 Mar 2023 19:05:07 GMT)._

### DIFF
--- a/content/patterns/alert/examples/alert.html
+++ b/content/patterns/alert/examples/alert.html
@@ -143,6 +143,18 @@
           sourceCode.make();
         </script>
       </section>
+      
+      <section id="at-support">
+        <h2>Assistive Technology Support</h2>
+        <iframe
+          class="support-levels-alert"
+          src="https://aria-at.w3.org/embed/reports/alert"
+          height="100"
+          allow="clipboard-write"
+          style="border-style: none; width: 100%;">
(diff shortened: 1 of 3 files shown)


alflennik:
Hi @mcking65, the slug-style names originate [here in the aria-at repo](https://github.com/w3c/aria-at/tree/master/tests), so for link, the `link-span-text` name is the one to use, and for radio the name in question is `radiogroup-aria-activedescendant`. The reason there are unique class names for each iframe is that message-passing script which fixes the iframe heights when the iframe height changes (like on page resize or when the details box opens and closes) needs a way to uniquely refer to the iframes when there are multiple instances on the page at once. I think in the near-to-medium-term future, both of these irritations will be resolved by having a "copy embed code" button directly in the aria-at-app. Hopefully that helps. I'll update the code in this PR to use the accurate names.

alflennik (approved):
Great to see more ARIA-AT data in the APG! I tested and they work well.

PR #2682, opened 2023-04-26
Remove duplicate SkipTo library

alflennik:
Removes the skipto from the aria-practices repo, since this is liable to cause conflicts with the skipto added by the wai-aria-practices library when the website is built.

Fixes issue 2676.
___
[WAI Preview Link](https://deploy-preview-216--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 04 May 2023 17:28:55 GMT)._


mcking65:
@alflennik 

I'm a bit confused. This is turning out differently from what I anticipated. we might need to discuss.

I actually don't understand why the build process had to de-dupe. I'm wondering if the build process was doing the duplication?

One of our overall goals with the repo structure is for aria-practices to be the source of all content. wai-aria-practices should contain only code that builds jekyll files and the output of that build process. Everything else should have aria-practices as the original source.

Along with all other shared scripts, the skipto.js as content that should be sourced from aria-practices. So, even if we were to change the html in aria-practices to remove it from the head as is done in this PR, the build process should still source skipto.js from `w3c/aria-practices/content/shared/js`.

That said, I don't understand why including skipto.js in the head of the aria-practices html files would be the cause of duplication. It appears that `scripts/pre-build/library/formatForJekyll.js` is appropriately including skipto.js in the head of the markdown files just as it is also including the other scripts referenced in head of the html files. The other scripts don't have a duplication problem. It seems like the duplication would be somehow happening in wai-aria-practices and it should be removed from there.

Note that I am assuming everything in `w3c/wai-aria-practices/content-assets/wai-aria-practices/shared/js` is being pulled from `w3c/aria-practices/content/shared/js`. I am further assuming that if we change those files in aria-practices, those changes will propagate via the build process to wai-aria-practices.

Where is my understanding incorrect?

alflennik:
Hi there @mcking65, sorry to cause any confusion.

You were asking why the build process had to de-dupe, and whether the build process was doing the duplication. You're right that the build process was causing duplication and could easily be altered to prevent any duplication. The reason I removed skipto from all the content files was not because it was the easiest way to fix the issue, although it does fix the issue, but because I felt like it was the most prudent course of action given that skipto isn't needed while authoring content, and the fact that removing it at the source would simplify the build scripts.

You also mentioned that content should originate in the aria-practices repo and wai-aria-practices should be limited to more presentational concerns like generating the jekyll files. I definitely agree with that. I think in this case I was mentally categorizing skipto as a presentational element but I'm 100% happy to treat it as content and move it back to aria-practices. I've updated the PR.

Also in your comment you described some investigation you did about possible causes and you were right to single out the formatForJekyll file, this is definitely the place the duplication was introduced. I think it was an artifact from a point last year that there were different skipto libraries in operation in aria-practices vs wai-aria-practices. But given that I elected to remove skipto from the content files it's no longer duplication anymore, it's more like singlication ...? In other words, it is now the singular and unique place that the skipto library is attached to website pages. 

Hopefully that helps. Let me know if I can clarify anything or if you want to tweak the direction this PR is taking.

mcking65:
@alflennik 

Why manage skipto.js differently from the other shared scripts?

The following shared scripts:
* app.js
* examples.js
* highlight.pack.js

All appear to have the same build needs as skipto.js. Each of the above:

* Has a script tag in the head of html files in the aria-practices repo
* Is sourced from aria-practices/content/shared/js
* Has a script tag in the markdown files in wai-aria-practices/aria/apg
* Is copied during build into wai-aria-practices/content-assets/wai-aria-practices/shared/js

It seems to me that we would be keeping the build process as simple as possible if it were to treat skipto.js like those other shared scripts.

Yes, all the scripts, except those that transform our html into jeckyll markdown files are content. All words, behaviors, images, etc. that are not supplied by the WAI templates are content and should be sourced from aria-practices.

mcking65:
Closing per discussion.

Issue #2628, opened 2023-02-28
Feed Display Example Broken Star Images


alflennik:
There are broken links in the [feed example](https://www.w3.org/WAI/ARIA/apg/patterns/feed/examples/feed/) on the [dedicated display page](https://www.w3.org/WAI/ARIA/apg/patterns/feed/examples/feed-display.html): there should be images representing the number of stars each restaurant has received.

We discussed the issue in the task force and we think the issue is caused by the fact that the links to the images are generated by JavaScript, and we think a good course of action would be to use an inline SVG instead of an external link.

a11ydoer:
@mcking65 How about having the discussion about updating feed example in the APG meeting? [Issue 565](https://github.com/w3c/aria-practices/issues/565) is present about design idea.

frozenzia:
The "correct" path for the star images, btw, is e.g. https://www.w3.org/WAI/content-assets/wai-aria-practices/patterns/feed/examples/imgs/rating-4.png

Hopefully that's not a "yeah, we already knew that" sort of thing, but thought I'd throw it out there.

mcking65:
Hi @frozenzia, Thank you for the PR. Per Howard's comment, I am going to close that PR.

However, I am wondering if you would be willing to work on the solution proposed at the top of this issue, which is to change the example code to use inline SVG for the ratings stars?

frozenzia:
@mcking65, I don't want to make any promises, but I'm definitely interested in giving it a shot. As I just mentioned elsewhere, though, I'm not very experienced with contributing to projects - may need a bit of hand-holding or at least don't be surprised if I end up asking "dumb" questions along the way.

frozenzia:
@mcking65, hey, sure, sign me up for this issue. If I've understood things properly, the next step here will be that "An editor will confirm there are no conflicting plans and, if needed, provide guidance."

mcking65:
@frozenzia wrote:

> @mcking65, hey, sure, sign me up for this issue. If I've understood things properly, the next step here will be that "An editor will confirm there are no conflicting plans and, if needed, provide guidance."


OK, super! As editor, I can confirm there are no conflicting plans, and this is a good time to move forward with a PR for this issue.

The proposed solution that the Task Force is landed on is to use inline SVG for the rating star images. You might take inspiration from [content/patterns/radio/examples/radio-rating.html](https://github.com/w3c/aria-practices/blob/main/content/patterns/radio/examples/radio-rating.html).

[...]

PR #2705, opened 2023-05-24
Fix path to ratings star images

frozenzia:
This has not been tested in any way by myself, and it's possible there's a problem with "backing up" the directory tree that much, but as best as I can tell, this should fix these paths.
___
[WAI Preview Link](https://deploy-preview-223--aria-practices.netlify.app/ARIA/apg) _(Last built on Wed, 24 May 2023 08:33:37 GMT)._


mcking65:
Closing per above comment and need for a different solution as described in #2628.

PR #2722, opened 2023-06-08
Checkbox Example (Mixed-State): Toggle on keyup to prevent continuous toggling

sivakusayan:
This is a follow-up PR to https://github.com/w3c/aria-practices/pull/2518 where [Checkbox Example (Two State)](https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox.html) was changed to only toggle on keyup. I think it makes sense to make the same change to [Checkbox Example (Mixed-State)](https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox-mixed.html) for consistency.

(Just a head's up, this is a resubmitting of #2541)
___
[WAI Preview Link](https://deploy-preview-231--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 13 Jun 2023 18:40:15 GMT)._

### DIFF
--- a/content/patterns/checkbox/examples/js/checkbox-mixed.js
+++ b/content/patterns/checkbox/examples/js/checkbox-mixed.js
@@ -16,6 +16,7 @@ class CheckboxMixed {
     this.checkboxNodes = domNode.querySelectorAll('input[type="checkbox"]');
 
     this.mixedNode.addEventListener('keydown', this.onMixedKeydown.bind(this));
+    this.mixedNode.addEventListener('keyup', this.onMixedKeyup.bind(this));
     this.mixedNode.addEventListener('click', this.onMixedClick.bind(this));
     this.mixedNode.addEventListener('focus', this.onMixedFocus.bind(this));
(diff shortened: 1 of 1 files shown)


sivakusayan:
It seems the failing test has to do with the [toolbar example](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/examples/toolbar/), where we try to click the font family button in the toolbar and [no dropdown menu appears from the button](https://github.com/w3c/aria-practices/blob/fec7032399028f3c3a17dbd05253e3140e0805d2/test/tests/toolbar_toolbar.js#L367).

Is this test flaky? I guess I would be surprised if my checkbox change caused this to fail.

mcking65:
@ccanash, will you please assign someone to investigate the cause of the failing test? I re-ran all tests, and it continues to fail.

andreancardona (approved):

mcking65 (approved):
Thank you @sivakusayan, the changes are working as expected. This is a good improvement to the behavior.

mcking65:
With this commit, #2425 is now fully resolved.

PR #2741, opened 2023-06-29
Add <select> to treegrid's focusable element list

### LINKED ISSUE #2740: Treegrid getFocusableElements() skips <select> elements
In the Treegrid pattern examples, there is a function called `getFocusableElements()` that is used to find interactive child elements of a selected row and toggle their `tabindex` values. The function specifies which elements should be focusable, noting that `<textarea>` has been left out deliberately:

`root.querySelectorAll('a,button,input,td>[tabindex]');`

The list should include the `<select>` element as well.

stephanieleary:
Adds <select> to the list of elements that should be able to receive focus in a treegrid row.

Fixes #2740.

### DIFF
--- a/content/patterns/treegrid/examples/js/treegrid-1.js
+++ b/content/patterns/treegrid/examples/js/treegrid-1.js
@@ -51,7 +51,7 @@ function TreeGrid(treegridElem, doAllowRowFocus, doStartRowFocus) {
     // textarea not supported as a cell widget as it's multiple lines
     // and needs up/down keys
     // These should all be descendants of a cell
-    var nodeList = root.querySelectorAll('a,button,input,td>[tabindex]');
+    var nodeList = root.querySelectorAll('a,button,input,select,td>[tabindex]');
     return Array.prototype.slice.call(nodeList);
   }
 


mcking65:
Hi @stephanieleary,

I am uncertain that the treegrid example could include a select in a cell with only this change. A task force member who is an engineer will need to review the code and test it.

From my limited study of the code, it does not appear to implement all features of the pattern. If I understand the code correctly, it does not support cells that include widgets that require arrow keys to operate.

I assume that when tabindex is set for elements in a cell on line 33, that navigation of the tree grid will break if tabindex is set to 0 on any element that consumes up/down/left/right. A select uses up/down. So, I am assuming that the change in this PR will cause treegrid nav to fail when `var focusable = getFocusableElements(cell)[0] || cell;` is executed because getFocusableElements would include the select.

Per the pattern definition, if a child of a cell consumes arrow keys, then focus needs to be set on the cell when navigating the grid. If the user is going to operate the child widget, the treegrid needs to provide a  mechanism for moving focus inside the cell, which disables grid navigation, and then back to the cell. In some implementations, which we don't have an example of in the APG, pressing enter on the cell opens the contents into a dialog.

Another approach that can be simpler when it is feasible is to make the cell content a button. Activating the button changes the content, e.g. replaces it with an input. The user action of completing the input with enter or canceling the input with Escape restores the button. This approach is illusttrated is in the [data grid example 2: Sortable Data Grid With Editable Cells](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/#examples). In example 2, note how the values in the description column are editable. The type column includes a widget similar to a select; it can't be opened with down arrow though.

PR #2776, opened 2023-08-10
Navigation Treeview Example: Correct typo: change 'an' to 'and' in accessibility features

sdvg:
Incorrect text:

>Because transparent borders are visible on some systems with operating system high contrast settings enabled, transparency cannot be used to create a visual difference between the element that is focused an other elements.

Corrected text:

>Because transparent borders are visible on some systems with operating system high contrast settings enabled, transparency cannot be used to create a visual difference between the element that is focused and other elements.

___
[WAI Preview Link](https://deploy-preview-249--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 10 Aug 2023 07:28:42 GMT)._

### DIFF
--- a/content/patterns/treeview/examples/treeview-navigation.html
+++ b/content/patterns/treeview/examples/treeview-navigation.html
@@ -354,7 +354,7 @@
             To support operating system high contrast settings:
             <ul>
               <li>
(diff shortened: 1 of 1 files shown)


mcking65 (approved):
Thank you @sdvg for the correction!

mcking65:
@sdvg this is now merged and will be included in the next publication, which is typically at the end of the month.

PR #2797, opened 2023-09-06
ReadMe: Remove section about coverage and quality report

howard-e:
Update link to APG Coverage Report in README.md 

The previous path (https://raw.githack.com/w3c/aria-practices/main/coverage/index.html), no longer exists following the APG restructure.

https://raw.githack.com/w3c/aria-practices/main/content/about/coverage-and-quality/coverage-and-quality-report.html seems to be the new location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,18 +106,3 @@ See the [APG Editorial Style Guidelines](https://github.com/w3c/aria-practices/w
 for information about writing prose for the APG.
 The [ReadMe for the ARIA specification](https://github.com/w3c/aria/)
 contains additional useful editorial guidance.
-
-## ARIA Roles, Properties and States Referenced in Guidance and Examples(Also known as APG Coverage Report)
-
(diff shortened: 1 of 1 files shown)


howard-e:
This build will fail until #2796 is merged

mcking65:
@howard-e 

The coverage reports have been moved inside the APG about section:
https://www.w3.org/WAI/ARIA/apg/about/coverage-and-quality/

As a result, we no longer need to surface them in the readme. We only had them there before because there wasn't anywhere else we could provide a path to them. They are not directly related to contributing to the APG, so I am repurposing this PR to remove that section from the readme.

howard-e:
> The coverage reports have been moved inside the APG about section: https://www.w3.org/WAI/ARIA/apg/about/coverage-and-quality/
> 
> As a result, we no longer need to surface them in the readme. We only had them there before because there wasn't anywhere else we could provide a path to them. They are not directly related to contributing to the APG, so I am repurposing this PR to remove that section from the readme.

@mcking65 okay got it, thanks for sharing that and cleaning up this PR!

Issue #2702, opened 2023-05-16
Simplify build of patterns and practices pages


mcking65:
#### The problem

Currently the [patterns](https://www.w3.org/WAI/ARIA/apg/patterns/) and [practices](https://www.w3.org/WAI/ARIA/apg/practices/) pages are built by scripts that process the content of all the individual pattern and practice pages. This auto-generation adds complexity and constrains editorial control.

The complexity of auto-generation sometimes leads to build failures, such as the failure mentioned in [this comment by James on PR 2670](https://github.com/w3c/aria-practices/pull/2670#issuecomment-1550188786).

Because the patterns and practices pages are built by scripts, editors can't control some important aspects of the content. They can't manipulate the order in which practices are presented. They also have limited control of the abstract that describes each pattern and practice on the patterns and practices pages. The abstract is pulled  from the first sentences of the about section of a pattern or practice. Generally, this works well, but there are times when the abstract could be improved by highlighting information that that can't be easily crammed into the first sentence without degrading readability.

The patterns and practices pages do not need to be updated frequently. So, the effort and time saved by auto-generation is fairly small.

Simplifying the build process will improve the long term sustainability of the APG by reducing the need for engineering support.

#### Solution Proposal

1. Source patterns page content from content/patterns/patterns.html
2. Source practices page content from content/practices/practices.html
3. Make the build of home, patterns, practices, and about as simple and consistent as is reasonable.
4. Document a clear set of classes and DOM structure requirements that editors can use when building and revising content for those pages.

#### Proposal for approach

* Start small and iterate
* Draft a PR that moves the practices page content into practices.html. This should also ensure that all images used on the practices page are sourced from content/images.
* After practices build is working well, repeat with another PR for the patterns page.
* Look for opportunities to simplify by better aligning the way home, patterns, practices, and about are coded and processed.

ccanash:
We will triage this task this week.

ccanash:
Hi @mcking65 , we triage this and the next step is clear for us but we won't work on this until we reprioratize APG work

ccanash:
Closing as this was deployed to prod on dec 12 https://github.com/w3c/wai-aria-practices/pull/280

PR #2807, opened 2023-09-19
Infrastructure: update skipto.js to version 5.2.1

jongund:
Updated skip.js to use public distribution.
This requires scripts referencing `skipto.js` to include the following data attribute:
```
  <script data-skipto="colortheme: aria; displayoption: popup" src="../../shared/js/skipto.js"></script>
```
 
This is related to [issue 259](https://github.com/w3c/wai-aria-practices/issues/259) in wai-aria-practices repository.
___
[WAI Preview Link](https://deploy-preview-260--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 06 Feb 2024 22:25:28 GMT)._


mcking65 (changes requested):
@jongund rather than add colortheme and displayoption to cspell, I think we should change the syntax to use either color-theme and display-option or colorTheme and displayOption. I am pretty sure either would get past cspell. What is your preference?

Also, a navigation region is getting wrapped around the skipto button. We don't want that. We need an option to suppress. Or, better yet, the default should be to not add it. Skipto is the first thing on the page. The nav region is not useful in practice, and it adds extra verbosity that is distracting when tabbing.

jongund:
The keywords are case insenstative, so the following updated configuration should fix the cspell issue and the navigation landmark problem is fixed by adding the `containerElement` option.

```
  <script data-skipto="colorTheme:aria; displayOption:popup; containerElement:div" src="../../shared/js/skipto.js"></script>
```

mcking65 (approved):

mcking65:
Thanks @jongund. That solved the problems.

This PR now provides the updated skipto with appropriate behaviors on all pages except for three:
1. [APG Home](https://deploy-preview-260--aria-practices.netlify.app/aria/apg/)
2. [Patterns | APG | WAI | W3C](https://deploy-preview-260--aria-practices.netlify.app/aria/apg/patterns/)
3. [Practices | APG | WAI | W3C](https://deploy-preview-260--aria-practices.netlify.app/aria/apg/practices/)

Since #2702, is not yet done, we will need to address the update of skipto for these three pages in wai-aria-practices via w3c/wai-aria-practices#259. We need to determine how that will be resolved before we merge this.

@a11ydoer, I'm assigning you as a reviewer to look at the visual presentation and mouse behavior of this version to make sure there are no regressions or unexpected changes.

a11ydoer (approved):
Navigation with keyboard only or mouse works.

howard-e:
@mcking65 @a11ydoer @jongund this PR has been updated with the suggested change to `patterns.html` and `practices.html` to include the `data-skipto` attributes in the `skipto` script import for those pages.

The [preview link](https://deploy-preview-260--aria-practices.netlify.app/aria/apg/) has also been rebuilt.

PR #2812, opened 2023-09-25
Infrastructure: link-checker.js: Check if any of the known `pageIds` include the hash being checked for

### LINKED ISSUE #2809: link-checker incorrectly says a fragment does not exist
In #2734, the link-checker is failing on the link:

https://github.com/w3c/aria-practices#code-conformance

It says the fragment `#code-conformance` was not found. I wonder if it is added dynamically and the link-checker is getting incomplete content?

this test failure is blocking progress on #2734.

howard-e:
Closes #2809. Follow up on my thoughts in https://github.com/w3c/aria-practices/issues/2809#issuecomment-1733823225.

Should this explicitly check for `user-content-` when checking GitHub links instead?
___
[WAI Preview Link](https://deploy-preview-261--aria-practices.netlify.app/ARIA/apg) _(Last built on Mon, 02 Oct 2023 13:59:20 GMT)._

### DIFF
--- a/scripts/link-checker.js
+++ b/scripts/link-checker.js
@@ -227,7 +227,7 @@ async function checkLinks() {
 
         let matchesHash = true;
         if (hash) {
-          matchesHash = !!matchingPage?.ids.includes(hash);
+          matchesHash = !!matchingPage?.ids.some((id) => id.includes(hash));
         }
 
         const isLinkBroken = !(
@@ -274,7 +274,11 @@ async function checkLinks() {
             hrefOrSrc.match(pattern)
           );
 
-        if (!isHashCheckingDisabled && hash && !pageData.ids.includes(hash)) {
+        if (
+          !isHashCheckingDisabled &&
+          hash &&
+          !pageData.ids.some((id) => id.includes(hash))
+        ) {
           consoleError(
             `Found broken external link on ${htmlPath}:${lineNumber}:${columnNumber}, ` +
               'hash not found on page'
--- a/scripts/link-checker.js
+++ b/scripts/link-checker.js
@@ -33,6 +33,23 @@ async function checkLinks() {
     return getLineNumber;
   };
 
+  const checkPathForHash = (hrefOrSrc, ids = [], hash) => {
+    // On some websites, the ids may not exactly match the hash included
+    // in the link.
+    // For e.g. GitHub will prepend client facing ids with their own
+    // calculated value. A heading in a README for example could be
+    // 'Foo bar', navigated to with https://github.com/foo/bar#foo-bar,
+    // but GitHub calculates the actual markup id included in the document
+    // as being 'user-content-foo-bar' for its own page processing purposes.
+    //
+    // See https://github.com/w3c/aria-practices/issues/2809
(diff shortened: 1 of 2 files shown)


scripts/link-checker.js
```diff
@@ -227,7 +227,7 @@ async function checkLinks() {
 
         let matchesHash = true;
         if (hash) {
-          matchesHash = !!matchingPage?.ids.includes(hash);
+          matchesHash = !!matchingPage?.ids.some((id) => id.includes(hash));
```
alflennik, line 230: Since it's not checking that the ID is exact I think an explanatory comment might be warranted, especially one that explains that this reproduces the behavior already present in GitHub.

howard-e: Done in https://github.com/w3c/aria-practices/pull/2812/commits/56d93dbc56487d3dfbe11c948a5b63f993dda70a

alflennik (approved):
That was a really helpful comment, I didn't know that!

scripts/link-checker.js
```diff
@@ -227,7 +227,16 @@ async function checkLinks() {
...
+          // On some websites, the ids may not exactly match the hash included
+          // in the link.
+          // For e.g. GitHub will prepend client facing ids with their own
+          // calculated value. A heading in a README for example could be
+          // Foo bar, navigated to with https://github.com/foo/bar#foo-bar,
+          // but GitHub calculates the actual markup id included in the document
+          // as being user-content-foo-bar for its own page processing purposes.
+          //
+          // See https://github.com/w3c/aria-practices/issues/2809
+          matchesHash = !!matchingPage?.ids.some((id) => id.includes(hash));
```
mcking65, line 239: I think this condition is not sufficiently restrictive. It makes it too easy for me (and others) to create href attributes that are not valid. I really want to be confident that our content does not contain invalid fragment ids. Even though there is the possibility of more maintenance to this code, I'd rather have the condition be specific to GitHub pages and the "user-content-" prefix. It does not seem likely we will encounter this problem on other pages referenced by the APG.

Over time, if it turns out GitHub is wishywashy in there approach or if this turns out to be a problem with other sites, we can decide how to address those other scenarios when they arise.

mcking65 (changes requested):
@howard-e 

Thank you for the investigation and the Fix! I would like to tighten up the conditions. see my comment in the code.

scripts/link-checker.js
howard-e: @mcking65 I've updated the code to account specially account for this GitHub scenario.

I also set it up to handle any others in the future using the `.link-checker.js` file which seems to handle some config settings.

howard-e: cc @alflennik I also removed `/^https:\/\/github\.com\/.*\/wiki\//` from `ignoreHashesOnExternalPagesMatchingRegex` as the same situation would happen on those wiki pages.

mcking65 (approved):
Thank you @howard-e! Looks great.

alflennik (approved):

mcking65:
Thank you @howard-e and @alflennik

PR #2831, opened 2023-10-06
Rating Slider Example: Redesign as an input for 10-value satisfaction scale

### LINKED ISSUE #2501: Rating slider is not a good pattern
In my opinion, the [rating slider](https://www.w3.org/WAI/ARIA/apg/example-index/slider/slider-rating.html) is not a pattern that we should recommend:
* if my blind colleague tells me that he operates the slider and I don't see any slider on the page, then it is confusing
* with the mouse I do not realize that I can select half values. If I click in the right half of a star, I won't notice that I can also click in the left half and give a different rating.
* Currently, it seems unclear whether or not the rating stars must meet the WCAG requirement for target size (see https://github.com/w3c/wcag/issues/2713). If yes, then they do not fulfill it

jongund:
Resolve #2501 by redesigning rating slider to be an input for an 10 point satisfaction scale.

Features:

* `0` value of the slider means no satisfaction rating has been selected, this is the value when rating slider is initialized.
* User can only select values between 1-10.

#### Preview link

[Preview Updated Rating Slider Example](https://deploy-preview-271--aria-practices.netlify.app/aria/apg/patterns/slider/examples/slider-rating/)

### Review checklist

*Reviewers:* To learn what needs to be covered by each review, Follow the link for the type of review to which you are assigned.

* [x] [Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review) by @curtbellew 
* [x] [Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review) by @ariellalgilmore 
* [x] [Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review) by @ariellalgilmore
* [x] [Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review) by @andreancardona 
* [x] [Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review) by @a11ydoer 
* [x]  [Color contrast] color contrast by @a11ydoer
___
[WAI Preview Link](https://deploy-preview-271--aria-practices.netlify.app/ARIA/apg) _(Last built on Mon, 18 Dec 2023 04:17:54 GMT)._


jongund:
@mcking65 
I made the changes to the rating slider example, should be ready for review

jongund:
@mcking65 
The link checker errors on files that I did not touch, maybe do to some of the problems with the slow website response.

mcking65:
Thank you @jongund, this is great!!

When I tried with VO on chrome, the VO speech is kind of wonky at the ends of the scale. I hope that is a VO/Chrome issue. It is worth investigating. If you press right arrow until you get to 10, VO says 9. Then pressing left it says 10. then pressing right is says 9. I get similar behavior at the low end of the scale. I don't have a similar problem on Windows, so I'm guessing it is not related to the JavaScript unless there is a js interop bug in the mix.

jongund:
@mcking65 

What do you think of the 1 rating being labelled "Not At All Satisfied", instead of "Extremely Dissatisfied"?
"Extremely Dissatisfied" doesn't sound quite right to me, but I can live with it for the example.

jongund:
@mcking65 '
i could not replicate the problem with VO saying 9 when on the value 10 on Chrome, Safari or Firefox
Not sure what is going on, it works fine for me.
The HOME and END keys are a different story with VO, they do not work for me.

mcking65:
@jongund asked:

> What do you think of the 1 rating being labelled "Not At All Satisfied", instead of "Extremely Dissatisfied"? "Extremely Dissatisfied" doesn't sound quite right to me, but I can live with it for the example.


I think extremely dissatisfied is the most appropriate label. It is the inversion of extremely satisfied, and it is common on such scales.

jongund:
@mcking65 
I have updated the rating slider based on the visual styling of the [Survey Monkey Customer Satisfaction Survey Template](https://www.surveymonkey.com/mp/customer-satisfaction-survey-questions/). 

The updated visual styling of the example has connected rectangles around each rating value and the selected rating value is larger than the other rating values, has a different forground/background colors and the value is bolded.  The larger rectangle and bolding is to meet the WCAG requirements of more than just changing the foreground/background colors of the selected value.   Also increased the size of the operable area to meet WCAG 2.2 minimum size requirements.

curtbellew:
Reviewed this using IOS, MacOS and windows using Chrome, Firefox, and Safari.  Keyboard and mouse worked great.  When I tested it in Safari with VO on MacOS I was duplicating the issue that @mcking65 mentioned above.  The issue seems to be specific to MacOS VO using Chrome because it worked well when I switched to Safari.

jongund:
@curtbellew 
Do you think the visual design looks like something someone would see and understand in a real satisfaction survey question?

curtbellew:
@jongund , It looked very natural to me.  There's good context around it with the group label and the labels for 'Extremely Unsatisfied' -'Extremely Satisfied' labels.  The rectangles made sense right away.

ariellalgilmore:
Hi @jongund! The new updates look great and I agree with @curtbellew on the rectangles making more sense right away! I noticed on mobile when viewing the slider it goes off the screen and creates a horizontal scrollbar on the page

Screenshot 2023-10-30 at 2 33 45 PM

jongund:
@curtbellew @ariellalgilmore 

I have updated the width of the rating slider so it should work fine in a iphone screen width.

a11ydoer:
> @curtbellew @ariellalgilmore
> 
> I have updated the width of the rating slider so it should work fine in a iphone screen width.

iphone12
@jongund I tested with my iphone12. It is almost but the rating scale is still outside of mobile view port.

a11ydoer:
@alastc @mbgower This APG example is trying to meet new WCAG criteria, target size for mobile device but it seems to be challenging. Any suggestions?

content/patterns/slider/examples/slider-rating.html
```diff
@@ -74,46 +74,46 @@ <h2 id="ex_label">Example</h2>
             <svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="520" height="96">
               <rect class="focus-ring" x="23" y="2" width="54" height="90" rx="5" />
               <g class="rating">
-                <rect class="value" height="24" width="48" x="13" y="30"></rect>
-                <text class="label" x="32" y="47">1</text>
-                <text class="description"  x="13" y="20">Extremely Unsatisfied</text>
+                <rect class="value" height="24" width="34" x="12" y="30"></rect>
```
a11ydoer: @jongund Is it possible to have the width to "24" instead of "34"?

jongund:
What drive are you testing with?

a11ydoer:
> What drive are you testing with?

I tested with my iphone 12.

a11ydoer:
Test review: Rating Slider example needs test ids.

Other slider examples also missing below test ids

- slider/examples/slider-color-viewer.html:
   - group-aria-labelledby
- slider/examples/slider-temperature.html:
   - aria-valuetext

a11ydoer:
suggestions from Alastair Campbell
@alastc

"If you have 10 items in one row you have a theoretical min-width of 32px each (assuming no other padding/margin on containing elements).
 
So, you could either:
Reduce the width of the items, keeping above 23px width.
Wrap the items if there is not sufficient width.
 
It might be best to allow the items to wrap if needed anyway, and make sure the containing elements have minimal padding/margin.
 
Or, you could also consider using flexbox to vary the width of each item between a min and max settings (e.g. 24px to 40px)."

mbgower:
I believe these are the two problems that have been flagged:
1. If the slider is too fine grained, it will be impossible to fit it in the width of the viewport without making each of the points smaller than 24x24 without spacing.
2. If the slider extends beyond the viewport to allow the targets to be wide enough, it will not pass Reflow since a user would need to horizontally scroll.

Feel free to correct my problem states, but I'll respond to those for now.

### 1. Target Size problem
There is actually language in the SC that excepts this situation
> Targets that allow for values to be selected spatially based on position within the target are considered one target for the purpose of the success criterion. Examples include sliders, color pickers displaying a gradient of colors, or editable areas where you position the cursor.

Not only does any slider get excepted by this, but this slider in particular only takes one tabstop, so it seems easy to argue that the slider is one target and entirely meets the use case for this exception.

### 2. Reflow problem
Given the exception mentioned for target size, I'm not sure this should ever need to be considered, but one could at least try to make the case that it meets the exception:
> Except for parts of the content which require two-dimensional layout for usage or meaning.

However, I'm going to argue that is not actually the case.
First, there's no reason this _has_ to be displayed in a horizontal slider.
Second, the WCAG 2.x Task Force has been discussing a number of ways content can have its own mechanisms for repositioning horizontal content that works equally for all users that should _not_ be considered "scrolling. The mechanism should be familiar to many people who consume movies on Netflix, etc. Rows of suggested movie titles exceed the width of the screen for all users. The user views offscreen titles by (depending on the implementation) either advancing the entire row by 'screen full' (that is if there are 5 titles visible, the next 5 titles take their place at the same time) or by incrementing the visible titles (the title on the left moves off screen, and a new title appears from the right side of the screen).
In your slider example, if for whatever reason you perceived a need to have each slider be large, you could provide a similar mechanism. Typically arrows are provided at each end of the visible list as affordances for sighted users who are using a pointer device; keyboard users never need to use these since the list advances as they navigate.

Incidentally, such a mechanism also prevents this kind of interaction from any potential of failing Dragging Movements.

There are a couple of other things I thought worth mentioning. The first is that there seem to be non-visible targets at either end of the slider which allow a user to reposition to the extremes (or to put it another way, the ultimate values have an enlarged hit area that extends beyond the horizontal width of the slider). It's an interesting implementation, and I see equivalent keyboard operation is provided with Home/End. (At the moment there don't appear to be keyboard operators documented.) However I do wonder about the lack of visual affordance for this, as well as the lack of pointer shape or mouse over effects on the entire slider. Combined with the decision to compartmentalize each of the numbered segments, it doesn't lead to a very clear pointer user interaction.

jongund:
@a11ydoer 

I am using an iPhone 12 and I get the full range of the slider on my screen.
[Rating Slider Preview](https://deploy-preview-271--aria-practices.netlify.app/aria/apg/patterns/slider/examples/slider-rating/)

jongund:
@a11ydoer @mcking65 
I updated the updated rating slider example to be responsive, it wasn't as hard as I thought it would be.

a11ydoer:
@jongund I checked with another iphone 12 and it looks ok as you mentioned. I had the larger font size setting on the phone. Flex box idea by Alastair may be the option to cover those cases. I will also review the code and leave the comment. 
@mbgower Thanks so much for your post. I read through your comments and I need more time to learn and understand by looking at other wcag docs. I will keep in touch.

jongund:
@mcking65 @a11ydoer 

I have improved high contrast (e.g. forced-colors) support by adding a media query for `forced-colors` and using CSS `canvas`, `canvastext` and `linktext` color values when forced colors is enabled.  This required adding `canvastext` and `linketext` to `cspell.json` file since making them camel case was causing a linting error for the CSS validator.

content/patterns/slider/examples/css/slider-rating.css
```diff
@@ -14,84 +14,88 @@
...
   fill: currentcolor;
   fill-opacity: 1;
 }
 
-.rating-slider[aria-valuenow="1.5"] svg .star-1 .star,
-.rating-slider[aria-valuenow="1.5"] svg .star-2 .fill-left {
-  fill: currentcolor;
-  fill-opacity: 1;
+.rating-slider svg .current .label {
+  fill: white;
```
jnurthen, line 45: Why is this one hardcoded to white? Isn't there a named color we could use here?

content/patterns/slider/examples/slider-rating.html
```diff
@@ -60,41 +58,63 @@ <h2 id="ex_label">Example</h2>
         </div>
         <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
         <div id="ex1">
-          <div id="id-rating-label" class="label">Rating</div>
+          <div id="id-rating-label" class="label">Rate your satisfaction with the service you received?</div>
```
andreancardona, line 61:
```suggestion
          <div id="id-rating-label" class="label">Rate your satisfaction with the service you received</div>
```

content/patterns/slider/examples/slider-rating.html
```diff
@@ -108,11 +128,10 @@ <h2>Accessibility Features</h2>
...
-            For example, the color of the star borders is set to match the foreground color of high contrast mode text by specifying the CSS <code>currentcolor</code> value for the <code>stroke</code> property of each inline SVG <code>polygon</code> element.
-            To enable the high contrast background color to be the used as the contrasting color when a star is not fully or partially filled, the <code>fill-opacity</code> attribute of the <code>polygon</code> is set to zero.
+            To ensure the borders of the rating values and focus ring have sufficient contrast with the background when high contrast settings invert colors, the color of the borders are synchronized with the color of the text content using a CSS media query selector  (e.g. <code>@media (forced-colors: active)</code>) .
+            For example, the color of the rating value borders is set to match link foreground color of high contrast mode text by specifying the CSS <code>canvas</code> and <code>linkText</code> values for the <code>stroke</code> and <code>fill</code> properties of each inline SVG <code>rect</code> and <code>text</code> elements.
```
andreancardona, line 134:
```suggestion
            For example, the color of the rating value borders is set to match the link foreground color of high contrast mode text by specifying the CSS <code>canvas</code> and <code>linkText</code> values for the <code>stroke</code> and <code>fill</code> properties of each inline SVG <code>rect</code> and <code>text</code> elements.
```

jongund:
@mcking65 @a11ydoer @andreancardona @ariellalgilmore @jnurthen 
I made some changes in the JS, CSS and HTML to fix the problem of pointer interaction outside the slider rail effecting the rating value.  Only interaction within the slider rail will change the value.

andreancardona (approved):

a11ydoer:
@mcking65 The review of forced color regarding color contrast was done. The documentation is also updated in [accessibility review section](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility-review)

mcking65:
@a11ydoer @jnurthen @jongund 

There is an unresolve comment by @jnurthen on line 46 of `content/patterns/slider/examples/css/slider-rating.css`.

mcking65 (approved):
I have completed editorial review.

@jongund @a11ydoer 

I made quite a few changes to content in all sections of the example page in addition to changing links on 5 other pages. My changes should be easy to distinguish by looking at the list of commits; I tried to make the commit titles descriptive. I'm re-requesting reviews from you to double-check my work.

mcking65:
Friendly nudg to @ariellalgilmore for accessibility review and  @andreancardona for code review.

mcking65:
@jongund, since this is your PR, I can't request review from you, but I'd like it if you were to double-check the changes I made in the last 10 commits.

ariellalgilmore (approved):
accessibility looks good to me!!

a11ydoer (approved):
The mouse and touch (without a screen reader running) are working as expected on Windows, macOS, iOS, and Android.

content/patterns/slider/examples/slider-rating.html
a11ydoer, line 1: I see the clarification on " less friendly" concept via revised commit. - "significantly increase efficiency for users who rely on keyboard navigation to perceive options and make a selection."
Great job and this is approved.

a11ydoer (approved):
The clarification for "user freindly" was done via this commit.

curtbellew (commented):
Tested with mouse and on mobile and found no issues.  Looks great.

PR #2860, opened 2023-11-13
suggestion: small update to WAI-ARIA Roles, States, and Properties

schalkneethling:
I added a small update suggestion for the second line under the heading, "WAI-ARIA Roles, States, and Properties". The line states that the author can use `menu` or `true`, but it might be good to highlight that that is because `true` is an alias for `menu`.

I would like to know your thoughts.
___
[WAI Preview Link](https://deploy-preview-274--aria-practices.netlify.app/ARIA/apg) _(Last built on Mon, 13 Nov 2023 16:38:50 GMT)._

### DIFF
--- a/content/patterns/menu-button/menu-button-pattern.html
+++ b/content/patterns/menu-button/menu-button-pattern.html
@@ -53,7 +53,7 @@
         <h2>WAI-ARIA Roles, States, and Properties</h2>
         <ul>
           <li>The element that opens the menu has role <a href="#button" class="role-reference">button</a>.</li>
-          <li>The element with role <code>button</code> has <a href="#aria-haspopup" class="property-reference">aria-haspopup</a> set to either <code>menu</code> or <code>true</code>.</li>
+          <li>The element with role <code>button</code> has <a href="#aria-haspopup" class="property-reference">aria-haspopup</a> set to either <code>menu</code> or <code>true</code> (<code>true</code> is an alias for <code>menu</code>)</code>.</li>
           <li>
             When the menu is displayed, the element with role <code>button</code> has <a href="#aria-expanded" class="state-reference">aria-expanded</a> set to <code>true</code>.
             When the menu is hidden, it is recommended that <code>aria-expanded</code> is not present.


patrickfox:
An additional thought - knowing that `aria-haspopup` was updated in [WAI-ARIA v1.1](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup) to support additional roles, the use of `true`/`false` should arguably be avoided.

My question: should there be an effort to encourage developers to support the newer standard by noting that the use of `true` is an anti-pattern?

Example:

> The element with role `button` has [aria-haspopup](https://w3c.github.io/aria/#aria-haspopup) set to `menu` (Note: `true` is an alias for `menu` and is valid for legacy support, but use of `true` is an anti-pattern).

content/patterns/menu-button/menu-button-pattern.html
```diff
@@ -53,7 +53,7 @@ <h2>Keyboard Interaction</h2>
         <h2>WAI-ARIA Roles, States, and Properties</h2>
         <ul>
           <li>The element that opens the menu has role <a href="#button" class="role-reference">button</a>.</li>
-          <li>The element with role <code>button</code> has <a href="#aria-haspopup" class="property-reference">aria-haspopup</a> set to either <code>menu</code> or <code>true</code>.</li>
+          <li>The element with role <code>button</code> has <a href="#aria-haspopup" class="property-reference">aria-haspopup</a> set to either <code>menu</code> or <code>true</code> (<code>true</code> is an alias for <code>menu</code>)</code>.</li>
```
mcking65, line 56: The ARIA specification does not define true as an alias for menu or vice versa. The specification requires browsers to treat both values as having the same meaning but only when specified as a value for aria-haspopup. The value of true can be applied to many other properties where you cannot specify menu in its place.

If an explanation is really necessary, than perhaps it might say something like:

When processing aria-haspopup, browsers treat menu and true as equivalent.

The question the APG TF needs to answer: If the APG TF were to include explanations of this nature more consistently  throughout the guide, would the additional words add clarity and increase understanding?

I have not done a survey of the guide to find out how many places we'd need language like this to be consistent. Over time, we are striving to have consistent editorial practice. It is a challenge. We have a backlog of content that needs attention.

More often than not, we lean toward trying to create greater clarity by reducing rather than adding words.

mcking65:
@patrickfox wrote:

> An additional thought - knowing that `aria-haspopup` was updated in [WAI-ARIA v1.1](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup) to support additional roles, the use of `true`/`false` should arguably be avoided.
> 
> My question: should there be an effort to encourage developers to support the newer standard by noting that the use of `true` is an anti-pattern?
> 
> Example:
> 
> > The element with role `button` has [aria-haspopup](https://w3c.github.io/aria/#aria-haspopup) set to `menu` (Note: `true` is an alias for `menu` and is valid for legacy support, but use of `true` is an anti-pattern).

I am not aware of any ARIA spec language that would support this. There are open ARIA issues with discussion threads that suggest otherwise.

content/patterns/menu-button/menu-button-pattern.html
schalkneethling: This makes complete sense to me, thank you for your response @mcking65 🙏

> It is a challenge. We have a backlog of content that needs attention.

Can anyone help with these? I would be happy to contribute if you point me in the right direction. Thanks again.

mcking65: Yes, anyone can help, and we'd love to have your help.

Please read [Contributing to the APG | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/about/contributing/).

We are working on some backlog grooming so that we can have a query that will show a prioritized list of bugs. For now, just searching for issues with the "bug" label is a good place to start. Our use of the "good first issue" label is inconsistent so it may not yet be very helpful.

mcking65:
Closing based on discussion in the comments of this PR.

PR #2871, opened 2023-12-01
Amend implicit vs explicit label advice in naming document

patrickhlauke:
Explicit `<label>`s are preferred over implicit ones, as they have better AT support.

Closes https://github.com/w3c/aria-practices/issues/2870

Issue #2870, opened 2023-11-30
Naming Form Controls with the Label Element recommends incorrect technique


dotjay:
On the [Providing Accessible Names and Descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) practices page under the heading [Naming Form Controls with the Label Element](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_with_labels), the text describes two techniques, commonly known to accessibility professionals as implicit and explicit labeling.

Implicit:

> Wrapping the checkbox and the labeling text in a `label` element as follows gives the checkbox an accessible name.

Explicit:

> A form control can also be associated with a label by using the `for` attribute on the `label` element.

The guidance goes on to state the following, which I believe to be incorrect:

> When possible, use the above encapsulation technique for association instead of the following for attribute technique.

However, _explicit_ labeling using `for` and `id` attributes has [long been established as the preferable technique](https://www.tpgi.com/html5-accessibility-chops-form-control-labeling/) due to better support in browsers and assistive technology software, which is echoed by the WAI's own [tutorial on Labeling Controls](https://www.w3.org/WAI/tutorials/forms/labels/) (with my emphasis):

> Associating labels implicitly
> In some situations, form controls cannot be labeled explicitly. For example, a content author might not know the id of a form field generated by a script, or that script might not add an id at all. In this case, the label element is used as a container for both the form control and the label text, so that the two are associated implicitly. Generally, _explicit labels_ are better supported by assistive technology.

This is backed up by experts [reporting issues with voice control software like Dragon Naturally Speaking](https://a11ysupport.io/tech/html/label_element) not working with implicit labeling techniques, and [more recent testing with screen readers](https://dev.to/thejaredwilcurt/the-truth-behind-implicitexplicit-form-labels-2e10).

I suggest that the advice on the [Providing Accessible Names and Descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) page be reworded to something like:

> When possible, use the following `for` attribute technique for association instead of the above encapsulation technique.

patrickhlauke:
Pull request submitted here @dotjay https://github.com/w3c/aria-practices/pull/2871

mcking65:
PR #2882, which resolves this issue is merged. The change is scheduled for publication tomorrow, December 19.

Thank you @dotjay and @patrickhlauke for raising the issue and shaping the resolution.

PR #2882, opened 2023-12-06
Accessible Names Practice: Recommend association via 'for' attribute when using 'label' element

mcking65:
Inspired by #2871, this is an alternative proposal for resolving #2870.

@patrickhlauke, I originally was looking at writing this as a suggestion in #2871, but it became too complicated. I appreciate the time you took to make a PR; I do not intend to be dismissive of your work on the issue.

In addition to reducing word count, here are the notable editorial differences and my reasons for proposing them. Feedback is welcome on all of them.

1. It describes the preferred syntax first. It seems beneficial to present the most valuable and important information first and enable people to skip over information they may not use.
2. It uses more concrete language for explaining the reason for the preference. In particular, help people understand that it is due to specific AT/Browser combinations not calculating the correct name.
3. It emphasizes the preference by mentioning it two times.
#### Preview

[Preview of the proposed changes to the names and description practice](https://deploy-preview-283--aria-practices.netlify.app/aria/apg/practices/names-and-descriptions/#namingtechniques)

___
[WAI Preview Link](https://deploy-preview-283--aria-practices.netlify.app/ARIA/apg) _(Last built on Sun, 17 Dec 2023 08:02:10 GMT)._

### DIFF
--- a/content/practices/names-and-descriptions/names-and-descriptions-practice.html
+++ b/content/practices/names-and-descriptions/names-and-descriptions-practice.html
@@ -307,8 +307,8 @@
             However, unless the text is programmatically associated with the checkbox, assistive technology users will experience a checkbox without a label.
           </p>
           <p>
-            HTML provides two syntaxes for associating a label with a form control.
-            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>id</code> of the control.
+            HTML provides two ways of associating a label with a form control.
+            The one that provides the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>id</code> of the control.
             This way of associating the label with the control is often called explicit association.
           </p>
 
@@ -316,7 +316,7 @@
 &lt;label for="subscribe_checkbox"&gt;subscribe to our newsletter&lt;/label&gt;</code></pre>
 
           <p>
-            The other syntax, which is known as implicit association,  is to wrap the checkbox and the labeling text in a <code>label</code> element.
+            The other way, which is known as implicit association,  is to wrap the checkbox and the labeling text in a <code>label</code> element.
             Some combinations of assistive technologies and browsers fail to treat the element as having an accessible name that is specified by using implicit association.
           </p>
 
--- a/content/practices/names-and-descriptions/names-and-descriptions-practice.html
+++ b/content/practices/names-and-descriptions/names-and-descriptions-practice.html
@@ -301,25 +301,28 @@
         <section id="naming_with_labels">
           <h3>Naming Form Controls with the Label Element</h3>
           <p>
-            The HTML <code>label</code> element enables authors to identify content that serves as a label and associate it with a form control.
+             The HTML <code>label</code> element enables authors to identify content that serves as a label and associate it with a form control.
             When a <code>label</code> element is associated with a form control, browsers calculate an accessible name for the form control from the <code>label</code> content.
           </p>
           <p>
             For example, text displayed adjacent to a checkbox may be visually associated with the checkbox, so it is understood as the checkbox label by users who can perceive that visual association.
             However, unless the text is programmatically associated with the checkbox, assistive technology users will experience a checkbox without a label.
-            Wrapping the checkbox and the labeling text in a <code>label</code> element as follows gives the checkbox an accessible name.
+            HTML provides two syntaxes for specifying this relationship.
+            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>ID</code></code> of the control.
           </p>
-          <pre><code>&lt;label&gt;
-  &lt;input type="checkbox" name="subscribe"&gt;
-  subscribe to our newsletter
-&lt;/label&gt;</code></pre>
-          <p>
-            A form control can also be associated with a label by using the <code>for</code> attribute on the <code>label</code> element.
-            This allows the label and the form control to be siblings or have different parents in the DOM, but requires adding an <code>id</code> attribute to the form control, which can be error-prone.
-            When possible, use the above encapsulation technique for association instead of the following <code>for</code> attribute technique.
-          </p>
+
           <pre><code>&lt;input type="checkbox" name="subscribe" id="subscribe_checkbox"&gt;
 &lt;label for="subscribe_checkbox"&gt;subscribe to our newsletter&lt;/label&gt;</code></pre>
+
+          <p>
+            The other syntax, which doesn't return the correct accessible name in as many combinations of assistive technologies and browsers, is to wrap the checkbox and the labeling text in a <code>label</code> element.
+          </p>
+
+<pre><code>&lt;label&gt;
+&lt;input type="checkbox" name="subscribe"&gt;
+subscribe to our newsletter
+&lt;/label&gt;</code></pre>
+
           <p>
             Using the <code>label</code> element is an effective technique for satisfying <a href="#naming_rule_visible_text">Rule 2: Prefer Visible Text</a>.
             It also satisfies <a href="#naming_rule_native_techniques">Rule 3: Prefer Native Techniques</a>.
--- a/content/practices/names-and-descriptions/names-and-descriptions-practice.html
+++ b/content/practices/names-and-descriptions/names-and-descriptions-practice.html
@@ -308,7 +308,7 @@
             For example, text displayed adjacent to a checkbox may be visually associated with the checkbox, so it is understood as the checkbox label by users who can perceive that visual association.
             However, unless the text is programmatically associated with the checkbox, assistive technology users will experience a checkbox without a label.
             HTML provides two syntaxes for specifying this relationship.
-            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>ID</code> of the control.
+            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>id</code> of the control.
           </p>
 
           <pre><code>&lt;input type="checkbox" name="subscribe" id="subscribe_checkbox"&gt;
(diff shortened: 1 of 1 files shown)


content/practices/names-and-descriptions/names-and-descriptions-practice.html
```diff
@@ -301,25 +301,28 @@ <h4>Warning</h4>
...
-          <pre><code>&lt;label&gt;
-  &lt;input type="checkbox" name="subscribe"&gt;
-  subscribe to our newsletter
-&lt;/label&gt;</code></pre>
-          <p>
-            A form control can also be associated with a label by using the <code>for</code> attribute on the <code>label</code> element.
-            This allows the label and the form control to be siblings or have different parents in the DOM, but requires adding an <code>id</code> attribute to the form control, which can be error-prone.
-            When possible, use the above encapsulation technique for association instead of the following <code>for</code> attribute technique.
+            HTML provides two syntaxes for specifying this relationship.
+            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>ID</code> of the control.
```
patrickhlauke, line 311:
```suggestion
            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>id</code> of the control.
```

content/practices/names-and-descriptions/names-and-descriptions-practice.html
```diff
@@ -301,25 +301,28 @@ <h4>Warning</h4>
...
           </p>
+
           <pre><code>&lt;input type="checkbox" name="subscribe" id="subscribe_checkbox"&gt;
 &lt;label for="subscribe_checkbox"&gt;subscribe to our newsletter&lt;/label&gt;</code></pre>
+
+          <p>
+            The other syntax, which doesn't return the correct accessible name in as many combinations of assistive technologies and browsers, is to wrap the checkbox and the labeling text in a <code>label</code> element.
```
patrickhlauke, line 318: the "in as many" sounds like we actually went around counting the totals. i'd generalise it more into something more like... "The other syntax is to wrap […]. However, this syntax is not supported by all combinations of assistive technologies and browsers." (if possible, i'd even be more specific and say which ones, at time of writing, have problems)

patrickhlauke:
while the implicit/explicit naming isn't in the HTML spec per se, it's a fairly common way of referring to these two syntaxes https://www.google.com/search?q=implicit+explicit+html+label ... including in actual WAI material https://www.w3.org/WAI/tutorials/forms/labels/#:~:text=Associating%20labels%20implicitly,-In%20some%20situations&text=In%20this%20case%2C%20the%20label,better%20supported%20by%20assistive%20technology.

mcking65:
@howard-e @ccanash 

The preview link is still missing on this PR even though none of the checks are failing. Is the issue with the build workflow resolved? Is it possible to provide a preview for this PR?

content/practices/names-and-descriptions/names-and-descriptions-practice.html
mcking65: I made some revisions to address this. Rather than just saying it is not supported, I tried to use words that help the reader understand what is not happening. In general, we avoid naming AT in the APG. We are plannig to provide support data that can show AT specifics via the aria-at data using the AT support tables on the example pages. Over time, that data will be able to surface in more places.

mcking65:
@patrickhlauke, @a11ydoer, There is a possibility we will do a publication this week (Dec 12). If so, I'll have to merge this tomorrow (Dec 11) morning if hthis were to be included. Otherwise, it can be published on Dec 19.

I'll wait for approving reviews from both of you before merging.

a11ydoer:
@mcking65 I reviewed the changes, and it looks good. I fully understand @patrickhlauke's kind intention to help readers with non-supported browsers and AT combinations, but it would be hard to update all the details as technology advances. Furthermore, the AT project will fulfill that need.

content/practices/names-and-descriptions/names-and-descriptions-practice.html
a11ydoer, line 1: @mcking65 this is easy to read and understand. Thanks for the update.

patrickhlauke:
i would still like to see the "implicit vs explicit" mention though, based on https://github.com/w3c/aria-practices/pull/2882#issuecomment-1843557219 ... developers do use that terminology, so why try to avoid it for some "the spec itself doesn't call it that"?

a11ydoer:
> i would still like to see the "implicit vs explicit" mention though, based on [#2882 (comment)](https://github.com/w3c/aria-practices/pull/2882#issuecomment-1843557219) ... developers do use that terminology, so why try to avoid it for some "the spec itself doesn't call it that"?

Hi @patrickhlauke, 

Actually, I suggested a similar thing at the APG meeting last week. Use "implicit vs. explicit" concepts and link those definitions to somewhere, even in the ARIA spec. I see that WAI material used implicit and explicit concepts from your WAI reference. Can you help us to find the definitions of "explicit" and "implicit" so that we can use the words?

All the reasoning behind this is that APG is aiming for easy reading, targeting a 6th-grade reading level, aligning with WCAG 3 efforts. This approach is intended for **both developers and beginners in ARIA**. I've checked, and it seems that "implicit" and "explicit" may not be at a 6th-grade vocabulary level.

mcking65:
@a11ydoer wrote:

> All the reasoning behind this is that APG is aiming for easy reading, targeting a 6th-grade reading level, aligning with WCAG 3 efforts. This approach is intended for **both developers and beginners in ARIA**.

I didn't realize that we had 6th-grade level is a target. I thought that there was so much dispute over the meaning of such measurement systems and the bias built into them that we were explicitly not adopting them.

I have simply been striving for clear language that is not more complex than necessary.

>I've checked, and it seems that "implicit" and "explicit" may not be at a 6th-grade vocabulary level.

I'm a little confused. Is this reason not to use the words "implicit" and "explicit"? If we were to add links to their definitions, it wouldn't change the grade-level of the prose.

The APG text itself provides the special contextual meaning of "implicit labeling" and "explicit labeling". So, I suggest that if we add these terms to the text that we would not add definition links. The links would be more distracting than helpful because they would point to another place that provides essentially the same information that is in the APG prose.

I'm not sure I understand the value of including the implicit/explicit terminology, but I will add it and people can decide if the content is or is not better with that terminology included.

patrickhlauke:
i'll admit that i seriously doubt some of the more inherently complex and technical concepts of ARIA will be (easily or at all) conveyable in simple language.

in terms of adding a definition, I was thinking along the lines of my original PR - where the concept itself is explained, and then at the end adding `This is often called <em>implicit</em> labeling`. This, I'd argue, then doesn't need it own dictionary definition of what "implict" means, because if a reader doesn't understand/know that term, they still understand the concept because that explanation was not reliant on it.

shawna-slh:
+1 for keeping the language as simple as feasible
+1 for not getting too "hung up" on calculated reading levels
+1 to where the concept itself is explained, and then at the end adding This is often called <em>implicit</em> labeling. -- for us old birds for whom implicit and explicit labels makes sense :-)

thanks for tagging me to weigh in

mcking65:
I updated the text to include the terms explicit association and implicit association.

daniel-montalvo (approved):

content/practices/names-and-descriptions/names-and-descriptions-practice.html
```diff
@@ -303,23 +303,28 @@ <h3>Naming Form Controls with the Label Element</h3>
...
           </p>
-          <pre><code>&lt;label&gt;
-  &lt;input type="checkbox" name="subscribe"&gt;
-  subscribe to our newsletter
-&lt;/label&gt;</code></pre>
           <p>
-            A form control can also be associated with a label by using the <code>for</code> attribute on the <code>label</code> element.
-            This allows the label and the form control to be siblings or have different parents in the DOM, but requires adding an <code>id</code> attribute to the form control, which can be error-prone.
-            When possible, use the above encapsulation technique for association instead of the following <code>for</code> attribute technique.
+            HTML provides two syntaxes for associating a label with a form control.
```
patrickhlauke, line 310: If we're aiming for simplicity/understandability (regardless of reading level etc), maybe instead of referring to "syntaxes" we could just talk about "ways" here and subsequently

content/practices/names-and-descriptions/names-and-descriptions-practice.html
```diff
@@ -303,23 +303,28 @@ <h3>Naming Form Controls with the Label Element</h3>
...
+            The syntax with the broadest browser and assistive technology support is to set the <code>for</code> attribute on the <code>label</code> element to the <code>id</code> of the control.
+            This way of associating the label with the control is often called explicit association.
           </p>
+
           <pre><code>&lt;input type="checkbox" name="subscribe" id="subscribe_checkbox"&gt;
 &lt;label for="subscribe_checkbox"&gt;subscribe to our newsletter&lt;/label&gt;</code></pre>
+
+          <p>
+            The other syntax, which is known as implicit association,  is to wrap the checkbox and the labeling text in a <code>label</code> element.
+            Some combinations of assistive technologies and browsers fail to treat the element as having an accessible name that is specified by using implicit association.
```
patrickhlauke, line 320: might be prudent to qualify this statement with "Currently, some combinations..." or similar, as the situation may well change

content/practices/names-and-descriptions/names-and-descriptions-practice.html
a11ydoer: Thanks for considering understandability and simplicity, @patrickhlauke !

mcking65: Revised acordingly.

content/practices/names-and-descriptions/names-and-descriptions-practice.html
mcking65: If the bugs are fixed and this text is still present, the text will be equally inaccurate whether or not it includes the word "currently". The only way to avoid that problem would be to include a phrase like "as of December 2023", but I'm reluctant to start including such temporal prose in the APG.

So, it is somewhat risky to include any information of this nature. We have tried to avoid doing so. The long term strategy for addressing AT support is interop data from [aria--at](https://aria-at.w3.org/)., which  will be updated with each new release of a covered AT

patrickhlauke (approved):

PR #2910, opened 2024-01-25
Sortable Table Example: Improve CSS so it works when all columns are sortable

### LINKED ISSUE #2904: Sortable table fails without at least one column that is not sorted
Sortable table is a great example.  But it seems to fail unless (at least?) one column is marked non-sortable. If everything is sortable then the top header row is shrunken in height and the active zone is very small.

Is this a known problem?  Seems surprising that it hasn't been noticed before, since sorting all columns seems common.

I didn't find the issue in the issue database, which is surprising to me.  I'm happy to generate an example if this is not a known problem. 

-- Malcolm

kentslaney:
When all the columns are sortable, none of the header buttons create space in the layout, causing the th row's height to collapse. Fixes #2904.
___
[WAI Preview Link](https://deploy-preview-294--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 25 Jan 2024 23:33:33 GMT)._

### DIFF
--- a/content/patterns/table/examples/css/sortable-table.css
+++ b/content/patterns/table/examples/css/sortable-table.css
@@ -24,7 +24,6 @@ table.sortable th:nth-child(5) {
 }
 
 table.sortable th button {
-  position: absolute;
   padding: 4px;
   margin: 1px;
   font-size: 100%;


a11ydoer (approved):
The suggested change - removing position;absolute - solves the overlapping table header problem. Thanks so much for the PR, @kentslaney

curtbellew (approved):
Tested this change out via codepen and it works well

shirsha (commented):
Tested and review completed

Issue #2907, opened 2024-01-23
Infrastructure: Failing link-checker for GitHub links


ccanash:
First noted at https://github.com/w3c/aria-practices/pull/2775#issuecomment-1894400224 (although it does affect any new contributions as well, at this time)

The [failing build's](https://github.com/w3c/aria-practices/actions/runs/7546522687/job/20787981692?pr=2775) error message is `Found broken external link on content/about/contributing/contributing.html:69:15, hash not found on page`, which points to https://github.com/w3c/aria-practices#code-conformance.

ccanash:
Hi @mcking65  this issue came up today during the APG is it ok if I set the priority to P1?

howard-e:
An [additional failing check](https://github.com/w3c/aria-practices/actions/runs/7882635938/job/21508158856) here first seen yesterday, is the following:

<details>
<summary>error stack</summary>

```bash
Found broken external link on content/patterns/landmarks/examples/at.html:198:21
  FetchError: maximum redirect reached at: https://chrome.google.com/webstore/detail/skipto-landmarks-headings/fjkpbfcodhflpdildjbmdhhmcoplghgf
    at ClientRequest.<anonymous> (/home/runner/work/aria-practices/aria-practices/node_modules/node-fetch/lib/index.js:1590:15)
    at ClientRequest.emit (node:events:529:35)
    at HTTPParser.parserOnIncomingClient (node:_http_client:700:27)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:119:17)
    at TLSSocket.socketOnData (node:_http_client:541:22)
    at TLSSocket.emit (node:events:517:28)
    at addChunk (node:internal/streams/readable:368:12)
    at readableAddChunk (node:internal/streams/readable:341:9)
    at Readable.push (node:internal/streams/readable:278:10)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:190:23)
Link checker found 1 broken link
```

<details

PR #2931, opened 2024-02-08
Fix failing link checker on GitHub README link

evmiguel:
This PR addresses #2907, where the link checker is failing due to a reference to a GitHub README link. The link checker fails because the page uses a `react-partial` element to wrap the README content.

Questions for reviewers:
- This is a first pass solution, and I am looking for more feedback on how to make this solution less brittle, in case GitHub decides to change how they handle READMEs in the future. Thoughts?

### DIFF
--- a/.link-checker.js
+++ b/.link-checker.js
@@ -1,3 +1,5 @@
+const HTMLParser = require('node-html-parser');
+
 module.exports = {
   filesToIgnore: [
     // For example:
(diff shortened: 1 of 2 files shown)


scripts/link-checker.js
```diff
@@ -142,7 +142,18 @@ async function checkLinks() {
...
+            const ssr = html
+              .querySelectorAll('react-partial')
+              .filter(
+                (partialElement) =>
+                  partialElement.getAttribute('partial-name') ===
+                  'repos-overview' // This is the partial that handles the READMEs
+              )
+              .flatMap((element) => element.getElementsByTagName('script'))
+              .map((element) => JSON.parse(element.innerHTML))[0];
+            return { ok: response.ok, status: response.status, ids, ssr };
```
alflennik, line 156: Would it be possible to add an option for this work in the `.link-checker` file? If we can avoid dumping this complexity in the main script that would be a big win.

.link-checker.js
```diff
@@ -18,13 +20,27 @@ module.exports = {
     {
       name: 'github',
       pattern: /^https:\/\/github\.com\/.*/,
-      matchHash: (ids, hash) =>
-        ids.includes(hash) || ids.includes(`user-content-${hash}`),
+      matchHash: (ids, hash, ssr) => {
```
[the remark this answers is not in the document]
alflennik: I agree with Howard's point about adding an object here. 

I would say though that I don't think this is properly called `ssr`. After all, if GitHub was doing SSR properly then we would be dealing with correctly formatted HTML already and none of these workarounds would be necessary.

As you can see in my other comment, I think we need to add another option to the configuration file to contain the complexity that GitHub is forcing us to handle. When you do that maybe you can try out another name and/or way of passing data to this function that makes sense based on how you tackle it.

alflennik (commented):
I left a couple comments but my first bit of feedback is a biggie, which is I verified that this does indeed fix the issue with the page. Nice!

There is one link failing though. If you merge in the latest changes from the main branch there is another PR, merged yesterday, which fixes it.

evmiguel:
@howard-e @alflennik thank you for your feedback! I've made some changes based on them. Looking forward to your thoughts.

.link-checker.js
```diff
@@ -1,3 +1,24 @@
+const HTMLParser = require('node-html-parser');
+
+const getAttributeValue = (obj, attribute) => {
```
howard-e, line 3: Could you add a brief comment here on what this does for future context? Something like:

`// Checks object for attribute and returns value. If not found on first pass, recursively checks nested objects and arrays of nested object(s) until attribute is found. If not found, returns undefined.`

howard-e (approved):
Looks good to me! Thanks for addressing my feedback and thanks for the research put into this work!

We'll want to pay close attention to the link-checker checks, to determine how this holds up with any future GitHub changes -- but having this a bit more structured is great and makes it easier to think about adding future support. Not only for GitHub, but also if other domains are changing unexpectedly.

alflennik (approved):
Thanks for addressing my feedback, I'm happy with how this turned out!

.link-checker.js
evmiguel: Added

PR #3010, opened 2024-05-01
Explicitly sort results of reference-tables and coverage-report by path

howard-e:
Supersedes #2981 

* Bumps [glob](https://github.com/isaacs/node-glob) from 8.1.0 to 10.3.12.
* Update scripts for `reference-tables` and `coverage-report` to maintain previous sort order.


mcking65 (approved):
the output looks good

howard-e:
Updated the top comment to show #3037 is being superseded instead.

PR #2853, opened 2023-11-01
Infrastructure: Bumps stylelint from 15.1.0 to 15.11.0 and use recommended rules

howard-e:
Supersedes #2846

This updates `stylelint` which has a newly enforced rule, `declaration-block-no-redundant-longhand-properties` which is forcing instances of `top: a, right: b, bottom: c, left: d` to now be in the shorthand format of `inset: a b c d`.

For example, instances of the following:

```css
top: 0;
right: 0;
left: 0;
bottom;
```

has been changed to:

```css
inset: 0;
```

Following [the Code Guide on shorthand notation](https://github.com/w3c/aria-practices/wiki/Code-Guide#shorthand-notation), this seems to be the appropriate change to make.

PR #3014, opened 2024-05-03
Infrastructure: Use Stylelint GitHub Action matcher

nschonni:
Saw on a different repo that they've added their own built-in formatter. Could't just add `npm run lint:css -f github` because it confused Prettier, so just split the tasks in the CI config.

### DIFF
--- a/.github/workflows/lint-css.yml
+++ b/.github/workflows/lint-css.yml
@@ -33,10 +33,11 @@ jobs:
         with:
           cache: npm
 
-      - uses: xt0rted/stylelint-problem-matcher@v1
-
       - name: Install npm dependencies
         run: npm ci
 
       - name: Stylelint
-        run: npm run lint:css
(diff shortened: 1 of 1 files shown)


mcking65:
@nschonni 

I don't know anything about the new commands. What are the reasons for and benefits of this diff?

nschonni:
Similar to the old extra problem matcher, it will show a comment on the Files tab of PRs when there are issues. The benefit of swapping out the current one, is that one supported by Stylelint directly, should be kept up to date if they change any of their output format. This doesn't affect you running anything locally

howard-e (approved):
@nschonni thanks! Looks good to me

> Could't just add `npm run lint:css -f github` because it confused Prettier, so just split the tasks in the CI config.

Was wondering why not do this, so thanks for sharing.

nschonni:
@howard-e feel free to pull the changes into your other Stylelint PR #2853 if you prefer

PR #3068, opened 2024-07-28
Infrastructure: Group Minor/Patch Dependabot updates

nschonni:
Not sure how the commit message/PR title will show up for this one. The idea is that all the minor patch updates will be in one PR, but the major version updates will continue to be opened as individual PRs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,13 @@ updates:
     open-pull-requests-limit: 99
     commit-message:
       prefix: "Infrastructure"
+    groups:
+      dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
   - package-ecosystem: github-actions
     directory: "/"
     schedule:


howard-e (commented):
@nschonni interesting!

> Not sure how the commit message/PR title will show up for this one.

Ran this on my fork since I was also curious. Here's [a PR example with 10 dependencies included](https://github.com/howard-e/aria-practices/pull/83). Overall, I like the thought and this would cause a whole lot less 'noise'! There may be minor version bumps which will still require some code modification but figuring out what's changed and pulling that bump into its own PR shouldn't be an issue based on how the grouped PR is presented.

I wonder if there is also a review benefit in grouping together [like dependencies](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-3) which already need (at times) to move forward together, such as `eslint*`, `stylelint*`, but also require some code modification at times. Being able to view that isolated but 'grouped' change _could_ be useful but I don't feel strongly. Any thoughts on that?

Also -- if this is merged, it will require someone to enable Grouped security updates under https://github.com/w3c/aria-practices/settings/security_analysis. I don't have access to that.

howard-e:
> but figuring out what's changed and pulling that bump into its own PR shouldn't be an issue based on how the grouped PR is presented

Looking again at the grouped PR, notes on `prettier` and `selenium-webdriver` were left off at the end. Unsure why.

nschonni:
> I wonder if there is also a review benefit in grouping together [like dependencies](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-3) which already need (at times) to move forward together, such as `eslint*`, `stylelint*`, but also require some code modification at times. Being able to view that isolated but 'grouped' change _could_ be useful but I don't feel strongly. Any thoughts on that?

Yeah, that might be cleaner, but since the update PRs here aren't usually as urgent to review and land, as with an app, I think it's fine to keep them as the big lump 😄

nschonni:
> Looking again at the grouped PR, change notes on `lint-staged`, `prettier` and `selenium-webdriver` were left off at the end. Unsure why.

I wonder if the size of the changelogs might have hit a limit on the API or the bot template

howard-e:
> > Looking again at the grouped PR, change notes on `lint-staged`, `prettier` and `selenium-webdriver` were left off at the end. Unsure why.
> 
> I wonder if the size of the changelogs might have hit a limit on the API or the bot template

Possibly! I'll check if that's the case

howard-e (approved):
@nschonni said:
> Yeah, that might be cleaner, but since the update PRs here aren't usually as urgent to review and land, as with an app, I think it's fine to keep them as the big lump 😄

Thanks again for this @nschonni. I agree with the above point!

My sole concern on [additional details disclosures being left off](https://github.com/w3c/aria-practices/pull/3068#issuecomment-2260744700) is because of the 16-bit limit GitHub also applies. There is a `Description has been truncated` comment attached to the bottom of my example PR which happens because of a [PR message max length](https://github.com/dependabot/dependabot-core/blob/99fc1412ae2a026d149744ef42f2e0ef1553f935/common/lib/dependabot/pull_request_creator/message_builder.rb#L162:L170) calculation that's done, the typical max of [65536](https://github.com/dependabot/dependabot-core/blob/99fc1412ae2a026d149744ef42f2e0ef1553f935/common/lib/dependabot/pull_request_creator/github.rb#L20). Not an issue since the table with the list of repositories is there and knowing which repository to check out is all that's needed (not that there's anything else we can do!)

PR #3072, opened 2024-07-30
Sortable Table Example: Separate off-screen portion of caption with parenthesis instead of comma

### LINKED ISSUE #3066: Comma on Sortable Table Example
Hi!

On the [Sortable Table Example](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) page, there's this example:

```
    <caption>
      Students currently enrolled in WAI-ARIA 101
      <span class="sr-only">
        , column headers with buttons are sortable.
      </span>
    </caption>
```

I understand that the idea here is that the first part is shown visually to everyone, and the second part is only "shown" to accessibility software. However, if you ignore the markup, it's [comma splice](https://en.wikipedia.org/wiki/Comma_splice) in terms of English grammar and style, which is already a bit of a problem. And when you need to make this text localizable, it becomes even more of a problem.

The reason I'm bringing up localization is that this example was actually used almost verbatim by frontend designers and developers in a translatable string in the project in which I provide support to the localizers. Translating this is a bit hard because this looks like one sentence, but its two parts don't sound so related. It would be nice to split it into two translatable strings, but splitting a translatable string in the middle of a sentence is not a great practice.

Would it be possible to split it without a comma? For example:
```
    <caption>
      Students currently enrolled in WAI-ARIA 101
      <span class="sr-only">
        (column headers with buttons are sortable)
      </span>
    </caption>
```

This looks more like two parts that can be separated cleanly: if you ignore the markup, it becomes "Students currently enrolled in WAI-ARIA 101 (column headers with buttons are sortable)", which is a good English sentence without comma splice, and it can also be split into two translatable strings without hurting understanding or structure.

I suggested this to our frontend developers, but they were reluctant about it because they wanted to follow W3C's example. Usually, following W3C's example is great, but perhaps this particular example could be changed? :)

Thanks!

ariellalgilmore:
Fixes #3066

updating sr-only text to parenthesis instead of a comma

#### Preview

[Preview Sortable Table Example in compare branch](https://deploy-preview-339--aria-practices.netlify.app/aria/apg/patterns/table/examples/sortable-table/)

___
[WAI Preview Link](https://deploy-preview-339--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 27 Aug 2024 02:57:06 GMT)._

### DIFF
--- a/content/patterns/table/examples/sortable-table.html
+++ b/content/patterns/table/examples/sortable-table.html
@@ -65,7 +65,7 @@
           <table class="sortable">
             <caption>
               Students currently enrolled in WAI-ARIA 101
-              <span class="sr-only">, column headers with buttons are sortable.</span>
+              <span class="sr-only">(column headers with buttons are sortable).</span>
             </caption>
             <thead>
               <tr>
--- a/content/patterns/table/examples/sortable-table.html
+++ b/content/patterns/table/examples/sortable-table.html
@@ -65,7 +65,7 @@
           <table class="sortable">
             <caption>
               Students currently enrolled in WAI-ARIA 101
-              <span class="sr-only">(column headers with buttons are sortable).</span>
+              <span class="sr-only"> (column headers with buttons are sortable).</span>
             </caption>
             <thead>
               <tr>
--- a/content/patterns/table/examples/sortable-table.html
+++ b/content/patterns/table/examples/sortable-table.html
@@ -65,7 +65,7 @@
           <table class="sortable">
             <caption>
               Students currently enrolled in WAI-ARIA 101
-              <span class="sr-only"> (column headers with buttons are sortable).</span>
+              <span class="sr-only">&puncsp;(column headers with buttons are sortable).</span>
             </caption>
             <thead>
               <tr>


content/patterns/table/examples/sortable-table.html
```diff
@@ -65,7 +65,7 @@ <h2 id="ex_label">Example</h2>
           <table class="sortable">
             <caption>
               Students currently enrolled in WAI-ARIA 101
-              <span class="sr-only">, column headers with buttons are sortable.</span>
+              <span class="sr-only">(column headers with buttons are sortable).</span>
```
mcking65, line 68: Important to have a space before the opening parend. Without the space, the there is no separation between the visible text and the off-screen text.

```suggestion
              <span class="sr-only"> (column headers with buttons are sortable).</span>
```

mcking65 (changes requested):
Suggestion for One small change.

content/patterns/table/examples/sortable-table.html
```diff
@@ -65,7 +65,7 @@ <h2 id="ex_label">Example</h2>
           <table class="sortable">
             <caption>
               Students currently enrolled in WAI-ARIA 101
-              <span class="sr-only">, column headers with buttons are sortable.</span>
+              <span class="sr-only"> (column headers with buttons are sortable).</span>
```
mcking65, line 68: Very strange ... when I look at the preview, the space preceding the parenthesis is not there. What would be causing it to get trimmed? Something about the class? Should we do something like include a `<br>` prior to or inside the `<span>`. I don't understand what is happening. This should be simple.

ariellalgilmore: hmm, I am seeing a space in VoiceOver output.... what screenreader are you using? Maybe we can discuss in the meeting tomorrow. Could also try adding `&nbsp;`

Screenshot 2024-08-12 at 8 34 31 PM

adampage:
```suggestion
              <span class="sr-only">&puncsp;(column headers with buttons are sortable).</span>
```

Heya @ariellalgilmore, the breaking space entity I was trying to remember is [`&puncsp;`](https://html.spec.whatwg.org/multipage/named-characters.html#:~:text=%CF%88-,puncsp%3B,-U%2B02008) — it might be worth a shot. It looks like there are other white space entities that might work well too: `&ensp;`, `&thinsp;`, etc.

I also found the related w3c/accname#225, which cross-references a CSS proposal named [`white-space-trim`](https://www.w3.org/TR/css-text-4/#propdef-white-space-trim).

cc: @mcking65

ariellalgilmore: Thank you @adampage! I am having trouble testing this on my devices. I am not the strongest JAWS user, maybe we can sync on how I can test this locally one time. I committed the suggestion in the meantime though, so it can be tested in the deploy preview.

mcking65 (approved):
Works now, thanks @ariellalgilmore!

PR #3082, opened 2024-07-31
Infrastructure: Bump Stylelint v16 and use MJS config file

nschonni:
Alternate to https://github.com/w3c/aria-practices/pull/2853
@howard-e I bumped the packages, but ignored the new failing rules in this one, to see if that makes it easier to land (and fix separately). I also converted the config file so comments could be added for why rules might be disabled.


.stylelintrc.mjs
```diff
@@ -0,0 +1,21 @@
...
+    'font-family-no-missing-generic-family-keyword': [
+      true,
+      {
+        ignoreFontFamilies: ['Font Awesome 5 Free'],
+      },
+    ],
+    'selector-class-pattern': null,
+    'selector-id-pattern': null,
+    // Fixable Stylelint 16 rules
+    'declaration-block-no-redundant-longhand-properties': null,
```
howard-e, line 17: Thinking further on this, it is better to ignore instead of what I was doing in #2853, as using `top right bottom left` is still so popularly used.

howard-e (approved):
@nschonni thanks! This is great and a plus over #2853

Issue #2735, opened 2023-06-26
Regression tests fail on macOS due to Windows-based assumptions


mcking65:
@frozenzia wrote:

In `toolbar_toolbar.js`, there are a couple of tests where CTRL+A are currently used to select all the text in a textarea. This does nothing on a Mac, and copying similar code from `alertdialog_alertdialog.js`, I've added an extra line to do the selecting in case the test is run on a Mac. After making this one and only change, running the tests still results in 2 test failures (+ 6 known failures):
```
combobox_grid-combo › content/patterns/combobox/examples/grid-combo.html [data-test-id="popup-key-home"]: Home from focus on list puts focus on grid and moves cursor

test/tests/combobox_grid-combo.js:880

 879:                                                           
 880:     t.true(                                               
 881:       await confirmCursorIndex(t, ex.comboboxSelector, 0),

Cursor should be at index 0 after one ARROW_HOME key

Value is not `true`:

false

› test/tests/combobox_grid-combo.js:880:11



menu-button_actions-active-descendant › content/patterns/menu-button/examples/menu-button-actions-active-descendant.html [data-test-id="menu-end"]: "end" on role="menu"

Rejected promise returned by test. Reason:

AssertionError {
  actual: 'mi3',
  code: 'ERR_ASSERTION',
  expected: 'mi4',
  generatedMessage: false,
  operator: 'strictEqual',
  message: 'aria-activedescendant should be set to mi4 for item: #ex1 [role="menu"]',
}

› assertAriaSelectedAndActivedescendant (test/util/assertAriaActivedescendant.js:24:10)
› async test/tests/menu-button_actions-active-descendant.js:471:3

─

2 tests failed
6 known failures
```

If I DON'T add those 2 lines in the toolbar test, I actually end up with 5 tests failed:
```
combobox_grid-combo › content/patterns/combobox/examples/grid-combo.html [data-test-id="popup-key-home"]: Home from focus on list puts focus on grid and moves cursor

test/tests/combobox_grid-combo.js:880

 879:                                                           
 880:     t.true(                                               
 881:       await confirmCursorIndex(t, ex.comboboxSelector, 0),

Cursor should be at index 0 after one ARROW_HOME key

Value is not `true`:

false

› test/tests/combobox_grid-combo.js:880:11



menu-button_actions-active-descendant › content/patterns/menu-button/examples/menu-button-actions-active-descendant.html [data-test-id="menu-end"]: "end" on role="menu"

[...]

PR #3115, opened 2024-09-16
Infrastructure: Address failing macOS regression tests

stalgiag:
This PR adds two small utility methods: `isMacOS()` and `translatePlatformKeys()`. The latter utility takes a single `Key` or an array of `Key`s and returns the keys with any platform-specific `Key`s translated. For example, `Keys.Control` becomes `Keys.Meta` on macOS.

These utilities are then used in the toolbar_toolbar test to ensure that the OS-specific modifier key is used in the "select all text" chord. 

This PR also adds a small delay to a `disclosure_navigation_hybrid` test to handle a scroll to focus event interfering with an assertion.

see #2735 
___
[WAI Preview Link](https://deploy-preview-352--aria-practices.netlify.app/ARIA/apg) _(Last built on Wed, 11 Dec 2024 17:19:43 GMT)._


howard-e (approved):
@stalgiag great job on getting this work started!

Found 2 additional failures when running on MacOS:
1. `disclosure_navigation_hybrid › content/patterns/disclosure/examples/disclosure-navigation-hybrid.html [data-test-id="link-aria-current"]: "aria-current" attribute on links`
2. `combobox_grid-combo › content/patterns/combobox/examples/grid-combo.html [data-test-id="popup-key-home"]: Home from focus on list puts focus on grid and moves cursor`

<details>
<summary>They were also described in #2735</summary>

```bash
─

  disclosure_navigation_hybrid › content/patterns/disclosure/examples/disclosure-navigation-hybrid.html [data-test-id="link-aria-current"]: "aria-current" attribute on links

  Rejected promise returned by test. Reason:

  ElementNotInteractableError {
    remoteStacktrace: `RemoteError@chrome://remote/content/shared/RemoteError.sys.mjs:8:8␊
    WebDriverError@chrome://remote/content/shared/webdriver/Errors.sys.mjs:193:5␊
    ElementNotInteractableError@chrome://remote/content/shared/webdriver/Errors.sys.mjs:353:5␊
    webdriverClickElement@chrome://remote/content/marionette/interaction.sys.mjs:167:11␊
    interaction.clickElement@chrome://remote/content/marionette/interaction.sys.mjs:136:11␊
    clickElement@chrome://remote/content/marionette/actors/MarionetteCommandsChild.sys.mjs:205:29␊
    receiveMessage@chrome://remote/content/marionette/actors/MarionetteCommandsChild.sys.mjs:85:31␊
    `,
    message: 'Element <a href="#mythical-page-content"> could not be scrolled into view',
  }

  ElementNotInteractableError: Element <a href="#mythical-page-content"> could not be scrolled into view
      at Object.throwDecodedError (/Users/howard/Projects/aria/aria-practices/node_modules/selenium-webdriver/lib/error.js:521:15)
      at parseHttpResponse (/Users/howard/Projects/aria/aria-practices/node_modules/selenium-webdriver/lib/http.js:514:13)
      at Executor.execute (/Users/howard/Projects/aria/aria-practices/node_modules/selenium-webdriver/lib/http.js:446:28)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async thenableWebDriverProxy.execute (/Users/howard/Projects/aria/aria-practices/node_modules/selenium-webdriver/lib/webdriver.js:742:17)
      at async /Users/howard/Projects/aria/aria-practices/test/tests/disclosure_navigation_hybrid.js:85:9



  combobox_grid-combo › content/patterns/combobox/examples/grid-combo.html [data-test-id="popup-key-home"]: Home from focus on list puts focus on grid and moves cursor

  test/tests/combobox_grid-combo.js:880

   879:                                                           
   880:     t.true(                                               
   881:       await confirmCursorIndex(t, ex.comboboxSelector, 0),

  Cursor should be at index 0 after one ARROW_HOME key

  Value is not `true`:

  false

  › test/tests/combobox_grid-combo.js:880:11

  ─
```

</details>

I do like this approach of starting the OS specific test utilities here and targeted a specific test. It should make resolving those other 2 issues easier to handle in the future.

I'm okay with approving and pushing for this work to get merged. We should also make sure the original issue is updated with a comment so it's clear what the 2 remaining issues are.

test/tests/combobox_grid-combo.js
```diff
@@ -877,10 +877,14 @@ ariaTest(
     // Send key "ARROW_HOME"
     await combobox.sendKeys(Key.HOME);
 
-    t.true(
-      await confirmCursorIndex(t, ex.comboboxSelector, 0),
-      'Cursor should be at index 0 after one ARROW_HOME key'
-    );
+    // On macOS, the HOME key deselects the grid popup
+    // but it doesn't move the cursor.
```
stalgiag, line 881: On macOS, pressing HOME does not move the cursor in the selector input. This was tested on multiple systems manually and with this test. Pressing the key does modify the `aria-activedescendant` as expected but it doesn't have the secondary effect of moving the cursor to the beginning of the input. This likely warrants a TODO or separate follow-up to determine if this should be addressed in the pattern implementation.

test/tests/disclosure_navigation_hybrid.js
```diff
@@ -82,7 +82,11 @@ ariaTest(
 
       if (links.length > 0) {
         await buttons[b].click();
+
         await links[0].click();
+        // Add a small delay here to ensure that the scroll to focus event
+        // has time to complete and doesn't interfere with the next assertion
+        await t.context.session.sleep(300);
```
stalgiag, line 89: This test was giving intermittent failures. I determined that this was due to the scroll to focus event fired by the link click interfering with the assertion (`buttons[1]` always failed). This small delay fixes this issue.

test/util/translatePlatformKey.js
```diff
@@ -0,0 +1,26 @@
...
+  [Key.CONTROL]: Key.META,
+};
+
+/**
+ * Translates a key or key combination for the current OS
+ *
+ * @param {string|string[]} keys - The key(s) to translate
+ * @returns {string[]} - The translated key(s) as a flat array ready for spreading
+ */
+function translatePlatformKey(keys) {
```
stalgiag, line 14: This function is potentially prematurely complex. I wanted to support instances when a platform key translation may be represented by chord -> single key or single key -> chord. We do not need this functionality now since we don't use any examples of this. I am open to reverting to a map if that is preferred.

howard-e: It potentially is but I could see the need for this as we move to ensuring tests are platform agnostic. Perhaps providing a clear example in the JSDoc will also be helpful.

howard-e (commented):
@stalgiag thanks so much for doing so much research into what's happening here! Left some inline comments

test/tests/toolbar_toolbar.js
```diff
@@ -1115,7 +1116,8 @@ ariaTest(
   'toolbar-button-enter-or-space',
   async (t) => {
     let textarea = await t.context.session.findElement(By.css('textarea'));
-    await textarea.sendKeys(Key.chord(Key.CONTROL, 'a'));
+    let modifierKey = translatePlatformKey(Key.CONTROL);
```
howard-e, line 1119: This being `modifierKey` as an array could be confusing. Renaming to `modifierKeys` works? Although maybe just as confusing given a single value was passed. I suppose this goes back to having the function provide a clear example of usage

.github/workflows/regression.yml
```diff
@@ -21,10 +21,11 @@ on:
 
 jobs:
   regression:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, macos-latest]
```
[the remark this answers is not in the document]
stalgiag: Thanks for this! I [decided to go with number 3](https://github.com/w3c/aria-practices/pull/3115/commits/a4688c03add445fecbe3243c8f93a2cd6d679246) since you already have a well-written example.

With these updates, I see that there are a handful of tests that do not fail on my local machine but do fail on macOS in the Github workflows. Unfortunately, many of these appear to be flakey failures. I see two options:

1. Revert the changes that add mac regression testing to CI
2. Address the new set of tests that only fail on CI in this PR as well

I am open to either. I just don't want to merge new workflows in that stand to impede development with flakey failures. Curious what your thoughts are.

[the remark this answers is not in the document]
howard-e: Yep, definitely flaky. I prefer going with number 1 here. I'd rather there be some dedicated focus and surety so as to not introduce additional CI flakiness. Would rather a follow up PR to introduce these workflows after some robust testing.

Also re-ran the tests multiple times on my macOS and not seeing those failures, so some dedicated effort on figuring out why it's happening in CI would be preferred.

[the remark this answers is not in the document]
stalgiag: I agree. Reverted!

test/tests/toolbar_toolbar.js
stalgiag: Good idea! I addressed with [aa905f6](https://github.com/w3c/aria-practices/pull/3115/commits/aa905f638bbea694cde78fc7b386647bdcc4ac8c).

howard-e (approved):
@stalgiag thanks for addressing the feedback. These changes LGTM now and should also make future instances of platform based issues easier to handle!

PR #3141, opened 2024-10-08
updated SkipTo.js version to a version that supports "instant" scrolling

jongund:
This version of skipto supports scrolling, but it is off by default.
Feedback that I have received indicates people with motion distractions want an opt in, rather than default feature,  So need to discuss at an upcoming meeting.
___
[WAI Preview Link](https://deploy-preview-364--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 17 Oct 2024 00:08:33 GMT)._


jongund:
There is a newer version available for update

PR #2991, opened 2024-04-15
Add Practice Page for Supporting Color Contrast Settings

mcking65:
Part of the resolution of issue #2864.

#### Preview Links

* [Preview revised practices page in compare branch](https://deploy-preview-480--aria-practices.netlify.app/aria/apg/practices/)
* [Preview color contrast settings practice page in compare branch](https://deploy-preview-480--aria-practices.netlify.app/aria/apg/practices/color-settings/)

___
[WAI Preview Link](https://deploy-preview-480--aria-practices.netlify.app/ARIA/apg) _(Last built on Fri, 21 Aug 2026 05:52:26 GMT)._

PR #3188, opened 2024-12-03
Add `@csstools/stylelint-formatter-github` as Stylelint's GitHub formatter

howard-e:
Stylelint's GitHub formatter was deprecated in [v16.8.0](https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md#1680). Running [`npx stylelint "**/*.css" -f github`](https://github.com/w3c/aria-practices/blob/b0c04c4bf4bae3d2de4b1fed893d250ec27f5de8/.github/workflows/lint-css.yml#L40) from the workflow now gives the following error and is being thrown in #2991:

```sh
[stylelint:004] DeprecationWarning: "github" formatter is deprecated
Please use "stylelint-actions-formatters" instead.
```

This PR uses [`--custom-formatter @csstools/stylelint-formatter-github`](https://github.com/csstools/postcss-plugins/tree/main/plugins-stylelint/formatter-github#csstoolsstylelint-formatter-github), which should be a drop in replacement for `-f github`

### DIFF
--- a/.github/workflows/lint-css.yml
+++ b/.github/workflows/lint-css.yml
@@ -37,7 +37,7 @@ jobs:
         run: npm ci
 
       - name: Stylelint
-        run: npx stylelint "**/*.css" -f github
+        run: npx stylelint "**/*.css" --custom-formatter @csstools/stylelint-formatter-github
 
       - name: Prettier
         run: npx prettier --check "**/*.css"
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
       "devDependencies": {
         "@babel/core": "^7.25.2",
         "@babel/eslint-parser": "^7.25.1",
+        "@csstools/stylelint-formatter-github": "^1.0.0",
         "ava": "^6.1.3",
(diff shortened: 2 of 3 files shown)


howard-e:
cc @nschonni @jongund

jongund:
@howard-e 
What does this mean for me as an author?
I get alot of CSS linting errors in the high contrast pull request.
What is the best way to fix them?

howard-e:
> @howard-e What does this mean for me as an author?

Looking again at your latest build, this error may be more of a warning instead of forcing a process exit so maybe nothing actually! If I'm wrong on that, you could approve and merge this work into `main` then merge `main` into your branch.

> I get alot of CSS linting errors in the high contrast pull request. What is the best way to fix them?

Running `npm run fix` and committing should do the trick. But I just double-checked that the `lint-staged` setup should be automatically reformatting when committing. Is it possible your git commit actions has a `--no-verify` happening? Whether from your cli or some tool?

nschonni (approved):
I think this makes sense. Other thought would be to just install it during CI, but that might lead to random failures

PR #3234, opened 2025-02-11
Infrastructure: MacOS Regression Testing

howard-e:
Follow-up from #3115.

Ensuring that #2735 doesn't become an issue in the future, should we also include macOS in the regression testing?
___
[WAI Preview Link](https://deploy-preview-385--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 11 Feb 2025 18:57:16 GMT)._


howard-e:
The errors reflected here aren't thrown when running `npm run regression` locally

PR #3258, opened 2025-04-15
Vertical Temperature Slider Example: Eliminate redundant AX tree rendering of value

### LINKED ISSUE #3257: Vertical Temperature Slider Example: Duplicative screen reader text following label
In the [Vertical Temperature Slider Example](https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-temperature/), a textual representation of the current value is rendered inside the SVG before the element with `role="slider"`.  IN the following example markup, the text is "25.0°C" (the default at page load):

```
<div id="id-temp-label" class="label">Temperature</div>
<svg role="none" class="slider-group" ...>
	<text class="temp-value" x="28" y="35">25.0°C</text>
	<rect class="rail" x="60" y="47" width="8" height="300" rx="5" aria-hidden="true"></rect>
	<g role="slider" .../>
	...
/svg>
```

The `<text>` element with `class="temp-value"` is exposed to screen readers, meaning that users:
* Hear the value twice, once from this text and once from the `aria-valuetext` attribute on the semantic slider element; and
* have to move their screen reader cursor passed the text that is of no use to them.

For example, here's a speech log from NVDA with speaking of command keys enabled, starting from just before the label:

> down arrow
> Temperature
> down arrow
> 25.0°C
> down arrow
> slider    25.0 degrees Celsius

In addition, it's not clear why the slider visually shows "°C" as the unit of measurement, while the `aria-valuetext` attribute explicitly expands it to "degrees Celsius".  Either all users should be expected to understand the common "°C" notation, or it should be spelled out for everyone.  Giving screen reader users more explanation encourages inequitable over-labelling, has a disproportionate impact in braille, etc.

CC @IsaDC, @mcking65

jongund:
Resolves #3257 

Changes the vertical temperature slider to eliminate redundant announcement of the slider value by screen readers. This change makes the value displayed visually above the slider a child of the slider so its content becomes presentational.

[Vertical Slider Preview](https://deploy-preview-400--aria-practices.netlify.app/aria/apg/patterns/slider/examples/slider-temperature/)

___
[WAI Preview Link](https://deploy-preview-400--aria-practices.netlify.app/ARIA/apg) _(Last built on Mon, 26 May 2025 02:02:22 GMT)._

### DIFF
--- a/content/patterns/landmarks/examples/resources.html
+++ b/content/patterns/landmarks/examples/resources.html
@@ -53,8 +53,8 @@
                             <li><a href="https://www.w3.org/TR/WCAG20-TECHS/aria#ARIA11">ARIA11: Using ARIA landmarks to identify regions of a page</a> (WCAG 2.0 Technique)</li>
                             <li><a href="https://www.w3.org/TR/WCAG20-TECHS/aria#ARIA20">ARIA20: Using the region role to identify a region of the page</a> (WCAG 2.0 Technique)</li>
                             <li><a href="https://alistapart.com/column/wai-finding-with-aria-landmark-roles">WAI-finding with ARIA Landmark Roles</a> (A List Apart)</li>
-                            <li><a href="https://www.tpgi.com/using-wai-aria-landmarks-2013/">Using WAI-ARIA Landmarks – 2013</a> (Paciello Group)</li>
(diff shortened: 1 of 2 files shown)


mcking65:
@a11ydoer  are you able to review? Hoping this good to merge this week in time for publication next week.
Note: we won't have a meeting next week though.

PR #3291, opened 2025-06-07
Remove explicit row="role" declaration from treegrid example

### LINKED ISSUE #1049: Treegrid example has <tr role="row"> which is not allowed per ARIA in HTML spec
http://w3c.github.io/html-aria/#tr says in the column "Implicit ARIA semantics - SHOULD NOT be used" for element `tr`:

> role=row, may be explicitly declared when child of a table element with role=grid

so for role=table or role=treegrid, `<tr role=row>` should not be declared explicitly.

The treegrid example in aria-practices uses `<tr role=row>`.

http://w3c.github.io/aria-practices/examples/treegrid/treegrid-1.html

Should the example omit `role=row`, or should ARIA in HTML be changed to allow it in `role=treegrid`?

cc @spectranaut @stevefaulkner @mcking65

adampage:
Resolves #1049.

### DIFF
--- a/content/about/coverage-and-quality/coverage-and-quality-report.html
+++ b/content/about/coverage-and-quality/coverage-and-quality-report.html
@@ -600,7 +600,6 @@
                 <li><a href="../../../content/patterns/combobox/examples/grid-combo.html">Editable Combobox with Grid Popup</a></li>
                 <li><a href="../../../content/patterns/grid/examples/layout-grids.html">Layout Grid</a></li>
                 <li><a href="../../../content/patterns/table/examples/table.html">Table</a></li>
-                <li><a href="../../../content/patterns/treegrid/examples/treegrid-1.html">Treegrid Email Inbox</a></li>
             </ul>
 </td>
           </tr>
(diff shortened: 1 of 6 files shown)


adampage:
Hi @mcking65 / @howard-e, it seems [the validator is still failing](https://github.com/w3c/aria-practices/actions/runs/15505047582/job/43658751392?pr=3291#step:5:22) on the absence of an explicit `tr[role="row"]` declaration.

Aside from that, the updated example appears to work just fine. I’ve only benchtested with macOS VoiceOver + Chrome so far, but the experience is identical with and without the explicit role declaration.

adampage:
Heya @mcking65, it looks like we can [remove 4 of our vnurc rule filters](https://github.com/w3c/aria-practices/pull/3291/files#diff-f31e5215b8463974a5c98ba7f89bea24281211c978a586a49c5d10345881d0ee), but [the “_`tr` is missing `role`_” error still remains](https://github.com/w3c/aria-practices/actions/runs/15569099984/job/43840405821?pr=3291).

PR #3302, opened 2025-07-01
Accordion Pattern: Remove extraneous colon character from heading

### LINKED ISSUE #3300: AutoScroll issues with Hash urls
I am attempting to reference a URL in my codebase, but I've encountered an issue when the URL includes a hash.

For example, consider the following link: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties:

When you click on this URL, it fails to direct you to the specific section indicated by the hash. The problem arises because the browser removes the : from the URL upon resolution. As a result, the autoscroll feature does not function as intended, and users are left at the top of the page.

To resolve this, I recommend removing the : from the heading "WAI-ARIA Roles, States, and Properties" to ensure that autoscrolling to the href works correctly.

mcking65:
Resolves #3300 by removing the trailing colon (:) character at the end of the heading for roles, states, and properties.
___
[WAI Preview Link](https://deploy-preview-408--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 01 Jul 2025 07:45:31 GMT)._

### DIFF
--- a/content/patterns/accordion/accordion-pattern.html
+++ b/content/patterns/accordion/accordion-pattern.html
@@ -74,7 +74,7 @@
       </section>
 
       <section id="roles_states_properties">
-        <h2>WAI-ARIA Roles, States, and Properties:</h2>
+        <h2>WAI-ARIA Roles, States, and Properties</h2>
         <ul>
           <li>The title of each accordion header is contained in an element with role <a href="#button" class="role-reference">button</a>.</li>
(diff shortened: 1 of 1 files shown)


a11ydoer (approved):
I confirmd that the colon was removed. I randomly checked other patterns, and they do not have any extranous colon.

PR #3362, opened 2025-09-11
Carousel Pattern: Fix inconsistency in use of digits and number words

mehrazmorshed:
In one place, found using slide one (word form), in another slide 2 (digit). 
Best to keep consistent (preferably digits in technical writing: slide 1, slide 2).
___
[WAI Preview Link](https://deploy-preview-435--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 11 Sep 2025 08:34:14 GMT)._


daniel-montalvo (approved):
The relevant line that I could find under our [Resoure style](https://www.w3.org/WAI/EO/wiki/Style) guide is:

> Use words for numbers that start a sentence, and for numbers up to one hundred (guidance based on documents where descriptive or narrative text is predominant and numbers are not the significant focus)

Agree that these numbers are not the significant focus.

Issue #3361, opened 2025-09-09
Infrastructure: `cspell` error with string 'invalid'


howard-e:
Happening in #3328, getting an error "Unknown word (inval)" when running cspell for the following files:

* https://github.com/w3c/aria-practices/blob/3066ace4b04cbfab832090c113f37878518b8586/content/patterns/spinbutton/examples/css/quantity-spinbutton.css#L84
* https://github.com/w3c/aria-practices/blob/3066ace4b04cbfab832090c113f37878518b8586/content/patterns/spinbutton/examples/css/quantity-spinbutton.css#L88
* https://github.com/w3c/aria-practices/blob/3066ace4b04cbfab832090c113f37878518b8586/content/patterns/spinbutton/examples/css/quantity-spinbutton.css#L116
* https://github.com/w3c/aria-practices/blob/3066ace4b04cbfab832090c113f37878518b8586/content/patterns/spinbutton/examples/quantity-spinbutton.html#L342

nschonni:
Might be a collision with some of the regex in  "ignoreRegExpList" array in the cspell.json

PR #3365, opened 2025-09-16
Infrastructure: Remove 'id=' from cspell's ignoreRegExpList

howard-e:
Address #3361

Removes `"id=\"(?:[^\\\"]+|\\.)*\""` from `cspell.json`'s `ignoreRegExpList` array. Before this, consider the case where a line is:

```html
<th scope="row"><code>aria-invalid="true"</code></th>
```

The portion, `id="true"` would be excluded leaving cspell to evaluate the following:

```html
<th scope="row"><code>aria-inval</code></th>
```

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -78,6 +78,7 @@
     "Dubnium",
     "Dušek",
     "Dvořák",
+    "Ebene",
     "Église",
     "Elasmobranchii",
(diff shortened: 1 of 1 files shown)


adampage (approved):
Look great, thank you @howard-e! cc: @mcking65

PR #3372, opened 2025-10-06
Add Experimental Example of Scrollable Listbox with Actions on Options

### LINKED ISSUE #3193: Request for aria-actions example with role=listbox/option
Where each option's has 1-2 inner action buttons that become visible only on focus. When the parent option is not focused, the actions are hidden but the option would still use aria-actions="" to indicate that actions are possible.

Chrome will use this example to work with the AT vendors.

mcking65:
Resolve #3193.

#### Review checklist

*Reviewers:* To learn what needs to be covered by each review, Follow the link for the type of review to which you are assigned.

Status|Asignees|Type of Review
---|---|---
WIP | @mcking65 |[Editorial review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#editorial_review)
WIP | @jongund |[Functional review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#functional_review)
WIP | @ariellalgilmore |[Visual design review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#design_review)
? | @ _|[Accessibility review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#accessibility_review)
Done | @joelamyman |[Code review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#code_review)
? | @ |[Test review](https://github.com/w3c/aria-practices/wiki/Pull-Request-Review-Process#test_review)___

#### Preview changes

[Preview link for example of Scrollable listbox with aria-actions](https://deploy-preview-474--aria-practices.netlify.app/aria/apg/patterns/listbox/examples/listbox-actions/)

[WAI Preview Link](https://deploy-preview-474--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 25 Jun 2026 01:04:39 GMT)._


mcking65:
@curtbellew 

Here's a branch to which you can add the necessary JS and CSS as we discussed last week.

Please read the documentation I wrote in the sections for:
* About This Experimental Example
* Keyboard Support
* Role, Property, State, and Tabindex Attributes

Please do not modify any of the JS or CSS referenced by the other listbox examples. Especially as an experimental example, it is important that we don't change the existing files. Please re-use them as much as possible, but put any new code in separate CSS and JS files. For now, in the `head` of this file, I left in place the links to the existing `js/listbox-scrollable.js`, `js/listbox.js` and `css/listbox.css`.

adampage (changes requested):

curtbellew:
I am getting lint spelling errors on the words 'favorited' and 'unfavorited' and I'm not sure what alternative words would make sense here.  These words are used in aria-live regions to indicate that the 'favorite' button has been clicked to toggle the favorites status on a particular option.  Any suggestions?

adampage:
> I am getting lint spelling errors on the words 'favorited' and 'unfavorited' and I'm not sure what alternative words would make sense here.  These words are used in aria-live regions to indicate that the 'favorite' button has been clicked to toggle the favorites status on a particular option.  Any suggestions?

How about "marked as favorite" and "unmarked as favorite"?

curtbellew:
> > I am getting lint spelling errors on the words 'favorited' and 'unfavorited' and I'm not sure what alternative words would make sense here.  These words are used in aria-live regions to indicate that the 'favorite' button has been clicked to toggle the favorites status on a particular option.  Any suggestions?
> 
> How about "marked as favorite" and "unmarked as favorite"?

This is along the lines of what I was thinking and I do like this.  I was a little worried about it being too verbose but I think this is probably the best alternative. Thanks!

content/patterns/listbox/examples/css/listbox-actions.css
```diff
@@ -0,0 +1,81 @@
+[role="option"] span.actions {
+  display: none;
+}
+
+[role="option"]:hover span.actions,
+[role="listbox"]:focus [role="option"].focused span.actions {
```
ariellalgilmore:
```suggestion
```
I wonder if we should remove this css to prevent this overlap from happening. This way only the options being hovered will show the action buttons and prevent the overlap if one is selected and the user moves the mouse
https://github.com/user-attachments/assets/67b85d47-452f-4a83-bda9-4a4ad3f426d4

jongund:
> > > I am getting lint spelling errors on the words 'favorited' and 'unfavorited' and I'm not sure what alternative words would make sense here.  These words are used in aria-live regions to indicate that the 'favorite' button has been clicked to toggle the favorites status on a particular option.  Any suggestions?
> > 
> > 
> > How about "marked as favorite" and "unmarked as favorite"?
> 
> This is along the lines of what I was thinking and I do like this. I was a little worried about it being too verbose but I think this is probably the best alternative. Thanks!

Can we use a toggle button pattern and the aria-checked state can indicate whether it is a favorite or not?

jongund (approved):
It seems to work as an example.

Interesting behavior when mouse is hovering over the list box and using the keyboard for navigation.  When the list item with focus is under the pointer, the TAB key moves between the action buttons, instead to leaving the listbox.

I only tried with Voice Over, and with Voice Over there is really no information that some of the accessiblle name is coming from action buttons, the action button names just seemed to be merged as part of the accessible name.   Not sure this is the best user experience.  

Seems to me items selected as favorites should also have a a persistent visual state, like a star next to the list item.

content/patterns/listbox/examples/js/listbox-actions.js
```diff
@@ -0,0 +1,428 @@
...
+    let listitemCurrentItemActionsButtonPosition,
+      listboxCurrentItemActionsButton;
+    this.listboxItemCurrent = this.listboxActionsNode.querySelector('.focused');
+    this.activeDescendant = this.listboxActionsNode.getAttribute(
+      'aria-activedescendant'
+    );
+    this.listboxActiveOption =
+      this.listboxActionsNode.querySelector('.focused');
+    this.listboxCurrentItemActionsButtons = Array.from(
+      this.listboxItemCurrent.querySelectorAll(
```
joelamyman: It was only from testing this in the Netlify preview, but I found that if I deleted an option, and then moved focus away from the pattern using the `Tab` key, that this causes an error due to there not being any focused items.

Therefore, to prevent this error, I wonder if it's worth checking if there is a listboxItemCurrent before this line 298, to prevent any errors resulting from this.

joelamyman (commented):
I've completed a review and left a few comments, all of which I think are non-blocking, and more minor pieces of feedback. I think this is looking great, and is such an interesting pattern! I found the JS easy to understand and work my way through, which was helped by the naming and the structure.

This is my first time doing anything like this, so I hope that the feedback is relevant and useful, please let me know if not. Alongside walking through the code and testing the behaviour to better understand it, I mainly relied on the wiki's code guide to help me.

The only other super minor thing I picked up on, again which may or may not be useful, is to prepend the classes used in the JS with `.js-` which I noticed in the wiki's code guide, but I didn't want to pollute the review with lots of these comments if this isn't helpful feedback.

mcking65:
Created #3400 for working on changes to content. It is desirable to keep this branch reasonably stable as the preview is being utilized by w3c/aria#2691.

content/patterns/listbox/examples/listbox-actions.html
```diff
@@ -0,0 +1,443 @@
...
+          options, similar to an HTML <code>select</code> with <code>optgroup</code> children.</li>
+      </ul>
+    </section>
+
+    <section>
+      <div class="example-header">
+        <h2 id="ex_label">Example</h2>
+      </div>
+      <div role="separator" id="ex_start_sep" aria-labelledby="ex_start_sep ex_label" aria-label="Start of"></div>
+      <div class="offscreen">Last change: <span aria-live="polite" id="ss_live_region"></span></div>
```
mcking65: Let's discuss options to make this live region content less prominent or invisible. Currently, it is the first element a screen reader user encounters, and it is confusing because it feels irrelevant.

Minimally, we could change the label to "notification:" and move the element to the bottom of the example.

It would be better if the element did not have a persistent lable and we could make the live content disappear from the DOM after about 10 seconds.

We could consider using ARIA notify, since this is an experimental example.

mcking65:
Some observations from screen reader and keyboard testing:

* Arrow keys are wrapping from last to first and vice versa; should not do that.
* Home/end not working.
*Focus moves to listbox without being focused on an item when a button, e.g., move up, is disabled.
* Accessible name of option does not reflect when item is favorited.

curtbellew:
@mcking65 , I have updated to fix the issues that we discussed.  The only thing I wasn't able to address as the 'Favorite' indication on the favorited option.  The text 'Favorite' already appears in the option name when it's favorited but I'm afraid it's getting lost in the announcements of the action buttons.  Since there is a button called 'Favorite', I'm afraid it's hard to tell that the text is there as well.

mcking65:
@curtbellew 

It's working really well. The focus bugs with the buttons that go away , e.g., when moving the second item to the first position, are fixed.

I hear the "favorite" indicator announcement; you are right, it is masked by the action button names. Often, I recommend that a status like that, e.g., "unread" in an email inbox, be announced before the name of the item. I think that we should do that here. However, that could be a follow up PR.

**Correcting this comment at 7:24 PM on June 24**

Earlier in the day I wrote:

>WRT keyboard support:
>* The arrow keys are still wrapping from first/last and vice versa.
>* Home/end are not functioning.

Testing later in the day, up/down and home/end are working correctly. I think the preview was failing to build so was out-of-sync with the latest commits.

mcking65:
Thank you @curtbellew  for all your fantastic work on this addition to the APG. This is an important milestone!

PR #3387, opened 2025-12-03
Developing a Keyboard Interface Practice: Clarify guidance for “Focusability of disabled controls”

### LINKED ISSUE #2318: Why does the 1.2 APG advocate that disable components are keyboard reachable?

adampage:
Makes small clarifications in the “[Focusability of disabled controls](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols)” section.

Closes #2318.

#### Preview

[Preview the revised keyboard practice page in the compare branch](https://deploy-preview-444--aria-practices.netlify.app/aria/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols)

___
[WAI Preview Link](https://deploy-preview-444--aria-practices.netlify.app/ARIA/apg) _(Last built on Wed, 15 Apr 2026 14:41:50 GMT)._

### DIFF
--- a/content/practices/keyboard-interface/keyboard-interface-practice.html
+++ b/content/practices/keyboard-interface/keyboard-interface-practice.html
@@ -391,9 +391,8 @@
       <section id="kbd_disabled_controls">
         <h2>Focusability of disabled controls</h2>
         <p>
-          By default, disabled HTML input elements are removed from the tab sequence.
-          In most contexts, the normal expectation is that disabled interactive elements are not focusable.
-          However, there are some contexts where it is common for disabled elements to be focusable, especially inside of composite widgets.
+          By default, HTML input elements with the <code>disabled</code> attribute are removed from the tab sequence.
+          However, there are some contexts where it is useful for an element to convey a disabled state while remaining focusable, especially inside of composite widgets.
           For example, as demonstrated in the
           <a href="../../patterns/menubar/menu-and-menubar-pattern.html">menu and menubar pattern</a>,
           disabled items are focusable when navigating through a menu with the arrow keys.
@@ -406,12 +405,12 @@
         </p>
 
         <p>
-          Authors are encouraged to adopt a consistent set of conventions for the focusability of disabled elements.
+          Authors are encouraged to adopt a consistent set of pattern-specific conventions for the focusability of disabled elements.
           The examples in this guide adopt the following conventions, which both reflect common practice and attempt to balance competing concerns.
         </p>
 
         <ol>
-          <li>For elements that are in the tab sequence when enabled, remove them from the tab sequence when disabled.</li>
+          <li>For disabled elements that don’t need to remain discoverable, remove them from the tab sequence.</li>
 
           <li>
             For the following composite widget elements, keep them focusable when disabled:
--- a/content/practices/keyboard-interface/keyboard-interface-practice.html
+++ b/content/practices/keyboard-interface/keyboard-interface-practice.html
@@ -402,7 +402,7 @@
         <p>
           Removing focusability from disabled elements can offer users both advantages and disadvantages.
           Allowing keyboard users to skip disabled elements usually reduces the number of key presses required to complete a task.
-          However, preventing focus from moving to disabled elements can hide their presence from screen reader users who "see" by moving the focus.
+          However, screen reader users are far less likely to discover disabled elements that are not focusable because moving focus is one of their primary methods of discovery.
         </p>
 
         <p>
--- a/content/practices/keyboard-interface/keyboard-interface-practice.html
+++ b/content/practices/keyboard-interface/keyboard-interface-practice.html
@@ -412,7 +412,7 @@ Authors are encouraged to adopt consistent pattern-based conventions for the foc
         <ol>
           <li>
             When users can reasonably infer the presence of a disabled element from nearby
-            focusable elements, it can be removed from the keyboard focus order
+            focusable elements, it is removed from the keyboard focus order
             using the HTML <code>disabled</code> attribute. For example:
             <ul>
               <li>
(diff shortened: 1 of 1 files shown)


mcking65 (changes requested):
made a couple of suggestions and ran out of steam ... struggling with GitHub changes.

adampage:
Heya @mcking65, how about [this latest commit](https://github.com/w3c/aria-practices/pull/3387/commits/781eee7d33f6d36cef0098e236e7b823473c35ae)? Here’s a [direct preview link](https://deploy-preview-444--aria-practices.netlify.app/aria/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols).

mcking65:
@adampage 

Thank you for the revisions and direct link. I'm not going to battle the new GitHub accessibility problems in the diff viewer today. Instead, I will just quote content here to give feedback.

Editorial style note: "We" (the task force) avoid use of personal pronouns, especially first and second person. There are two instances of "we".(irony intended 😄 )

I think we may leave readers scratching their heads a bit wondering why some elements needs to be discoverable and others do not. The text only partially conveys a rationale in the cut/copy/paste example. For the others, there is no explanation.

>1. When a disabled element does not need to remain discoverable, we apply the native disabled attribute so that it will no longer be focusable. For example:

What about something like:

>1. When users could reasonably infer the presence of a disabled element based on the nature of surrounding non-disabled elements that are focusable, disabled elements are removed from the keyboard focus order. To achieve this, the example conveys the disabled state by using the HTML disabled attribute. For example:

Then:

>In the example of a scrollable grid with previous and next buttons for scrolling, users can infer the presence of the disabled "previous" button when the grid is showing the first page of content and the "Next" button receives focus.

content/practices/keyboard-interface/keyboard-interface-practice.html
```diff
@@ -406,41 +406,40 @@ <h2>Focusability of disabled controls</h2>
         </p>
 
         <p>
-          Authors are encouraged to adopt a consistent set of conventions for the focusability of disabled elements.
+          Authors are encouraged to adopt a consistent set of pattern-specific conventions for the focusability of disabled elements.
```
daniel-montalvo: Just minor editorial, as I think the idea is that the conventions need to be based on the pattern, whether or not there's a set I don't think is relevant in this context.

```suggestion
Authors are encouraged to adopt consistent pattern-based conventions for the focusability of disabled elements.
```

content/practices/keyboard-interface/keyboard-interface-practice.html
daniel-montalvo: The idea of discoverability is introduced but I think it should be more clearly explained by relating it with perception and interaction.

```suggestion
However, preventing focus from moving to disabled elements can make them non-discoverable for screen reader users, as the only way for these users to interact with these elements is by moving the focus.
```

daniel-montalvo (dismissed):
Couple comments

content/practices/keyboard-interface/keyboard-interface-practice.html
mcking65: That wording could confuse users because users cannot interact with disabled elements. Perhaps something like:

>However, screen reader users are far less likely to discover disabled elements that are not focusable because moving focus is one of their primary methods of discovery.

daniel-montalvo:
@adampage  Link checker is failing here, I think it's  because of https://www.paciellogroup.com/blog/2013/02/using-wai-aria-landmarks-2013/

We may want to update that link with https://vispero.com/resources/using-wai-aria-landmark-roles/, which seems to be the most up-to-date version they have.

adampage:
Hi @mcking65, I’ve just finished incorporate your and @daniel-montalvo’s suggestions. Please check out the [latest deploy preview](https://deploy-preview-444--aria-practices.netlify.app/aria/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) and let me know what you think?

content/practices/keyboard-interface/keyboard-interface-practice.html
```diff
@@ -402,45 +402,49 @@ <h2>Focusability of disabled controls</h2>
...
           The examples in this guide adopt the following conventions, which both reflect common practice and attempt to balance competing concerns.
         </p>
-
         <ol>
-          <li>For elements that are in the tab sequence when enabled, remove them from the tab sequence when disabled.</li>
-
           <li>
-            For the following composite widget elements, keep them focusable when disabled:
-
+            When users can reasonably infer a disabled element from nearby
```
mcking65:
```suggestion
            When users can reasonably infer the presence of a disabled element from nearby
```

content/practices/keyboard-interface/keyboard-interface-practice.html
```diff
@@ -402,45 +402,49 @@ <h2>Focusability of disabled controls</h2>
...
+            focusable elements, it can be removed from the keyboard focus order
```
mcking65:
```suggestion
            focusable elements, it is removed from the keyboard focus order
```

mcking65 (changes requested):

adampage:
Heya @jongund and @mcking65, it seems [our link checker](https://github.com/w3c/aria-practices/actions/runs/24460911680/job/71474553442?pr=3387#step:5:12) has found [another broken link](https://github.com/w3c/aria-practices/blob/main/content/patterns/landmarks/examples/resources.html#L71), this time  pointing to [`https://ainspector.disability.illinois.edu`](https://ainspector.disability.illinois.edu), which serves up a `403` error.

Would [`https://addons.mozilla.org/en-US/firefox/addon/ainspector-wcag/`](https://addons.mozilla.org/en-US/firefox/addon/ainspector-wcag/) be a suitable replacement, or is it possible the `403` could be resolved or redirected by the university?

daniel-montalvo (approved):

daniel-montalvo (commented):
@mcking65 I approved these changes.

PR #3418, opened 2026-03-06
Landmark Practice: Update alignment between HTML aside element and ARIA complementary role

NakajimaTakuya:
I have revised the description of complementary roles where it now appears to be inaccurate.

* The description that the `aside` element uniformly has the complementary role.
* The description that the complementary role is a top-level landmark role

According to html-aam, it appears that the complementary role is now assigned to aside elements only in two specific cases: when they are in the context of a body or main element, and when they have an accessible name.

* https://www.w3.org/TR/html-aam-1.0/#el-aside-ancestorbodymain
* https://www.w3.org/TR/html-aam-1.0/#el-aside

related: https://github.com/w3c/html-aam/pull/350

(I'm not a native English speaker, so I'd appreciate it if you could suggest corrections if my writing seems awkward.)

#### Preview

[Preview the landmarks practice page in the compare branch](https://deploy-preview-454--aria-practices.netlify.app/aria/apg/practices/landmark-regions/)

___
[WAI Preview Link](https://deploy-preview-454--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 14 May 2026 10:42:30 GMT)._


jugglinmike:
The ARIA Authoring Practices (APG) Task Force just discussed this issue.

<details>
  <summary>The full log of that discussion</summary>

Matt_King: the contributor is trying to align with what HTML says about “aside”
Matt_King: I haven’t reviewed the spec recently
Matt_King: for right now, I would love to designate someone who would love to “run point” on the technical entails of this pull request–does it say what we need it to say, and are the details justified?
joe: I can review
daniel: I can review as well
Matt_King: I really appreciate that someone from the outside is paying close attention


</details>

jongund:
Related to this issue is the [ARIA in HTML spec for aside](https://www.w3.org/TR/html-aria/#el-aside) also needs to be updated.

jongund:
The `aside` element is behaving as specified in the AAM, but seems under specified in the AAM and ARIA in HTML specification based on testing.  

Did some additional testing by making a test page and the `aside` element produces a `complementary` landmark in the context of `body`, `main`, `header` and `footer` elements, but not in the context of `section`, `nav`, `form` and other `aside` elements, unless it has an accessible name.

NakajimaTakuya:
@jongund 

Thank you for verifying this.
I hadn’t realized that they are evaluated as complementary even in the header and footer contexts. I confirmed through my own testing that they are indeed treated that way.
However, since this seems to conflict with the documentation for html-aam, should we wait for the outcome of the discussion regarding html-aria for now?

mcking65:
Ya, seems like there may be bugs in the chrome and Firefox implementations. 

APG should be aligned with the spec, so we should get confirmation that the spec does not intend unnamed asides to be complementary in header and footer. I don't think it would be good for users if aside generated complementary landmarks in header and footer.

content/practices/landmark-regions/landmark-regions-practice.html
```diff
@@ -42,7 +42,7 @@ <h2>HTML Sectioning Elements</h2>
           </thead>
           <tbody>
             <tr>
-              <td><code>aside</code></td>
+              <td><code>aside</code> if it is in the context of a <code>body</code> or <code>main</code> element. Or if it has an accessible name provided by <code>aria-labelledby</code> or <code>aria-label</code></td>
```
joelamyman: After reviewing the [AAM spec](https://www.w3.org/TR/html-aam-1.0/#el-aside), this is accurate. I wonder if it would be useful to mention that the AAM specifies that if it's if the aside is scoped to a sectioning content element and has an accessible name?

For example:
```
<code>aside</code> if it is in the context of a <code>body</code> or <code>main</code> element. Or if it is scoped to another sectioning content element and has an accessible name provided by <code>aria-labelledby</code> or <code>aria-label</code>
```

At the same time, I appreciate this is starting to become a bit lengthy, so please feel free to dismiss this! 😅

content/practices/landmark-regions/landmark-regions-practice.html
```diff
@@ -90,7 +90,7 @@ <h2>General Principles of Landmark Design</h2>
 
         <ul>
           <li>Assign landmark roles based on the type of content in the area.</li>
-          <li><code>banner</code>, <code>main</code>, <code>complementary</code> and <code>contentinfo</code> landmarks should be top level landmarks.</li>
```
joelamyman, line 93: I'm not sure about removing complementary from this list - the ARIA spec states that:
> A [landmark](https://www.w3.org/TR/wai-aria/#landmark) that is designed to be complementary to the main content at a similar level in the DOM hierarchy, but remaining meaningful when separated from the main content.

Therefore, it's my understanding that the primary purpose of this landmark is to be positioned alongside the main content, so it would make sense for it to be a top level landmark. Does this make sense?

content/practices/landmark-regions/landmark-regions-practice.html
```diff
@@ -185,8 +185,6 @@ <h3 class="widget-name">Complementary</h3>
           </p>
 
           <ul>
-            <li><code>complementary</code> landmarks should be top level landmarks (e.g. not contained within any other landmarks).</li>
```
joelamyman, line 188: Given that the AAM specifies that it's possible for the complementary role to apply to an aside scoped to other sectioning content elements, this makes sense to be removed!

content/practices/landmark-regions/landmark-regions-practice.html
```diff
@@ -195,7 +193,7 @@ <h3 class="widget-name">Complementary</h3>
           <section>
             <h4>HTML Technique</h4>
 
-            <p>Use the HTML <code>aside</code> element to define a <code>complementary</code> landmark.</p>
+            <p>The HTML <code>aside</code> element defines a <code>complementary</code> landmark when has an accessible name (e.g. <code>aria-labelledby</code>, <code>aria-label</code>).</p>
```
joelamyman: Similar to my earlier comment, I wonder if this needs to specify both options, so maybe something like:
```
<ul>
  <li><p>The HTML <code>aside</code> element defines a <code>complementary</code> landmark when its context is the <code>body</code> or <code>main</code> element.</p></li>
  <li><p>The HTML <code>aside</code> element defines a <code>complementary</code> landmark when its context is another sectioning content element, and when it has an accessible name (e.g. provided using either the <code>aria-labelledby</code> or <code>aria-label</code> attributes).</p></li>
</ul>
```

joelamyman (commented):
These proposed changes look like an excellent step in the right direction!

I'm not 100% confident in my feedback as landmarks aren't my strong suit, however, after reviewing the AAM and the ARIA spec, I think that there may be a few minor wording changes needed, just to ensure that this accurately reflects the usage specified in both documents.

content/practices/landmark-regions/landmark-regions-practice.html
mcking65: we need a comma before the 'or' and we could shorten the text by eliminating the naming techniques:

<code>aside</code> if it is in the context of a <code>body</code> or <code>main</code> element, or if it is scoped to another sectioning content element and has an accessible name.

content/practices/landmark-regions/landmark-regions-practice.html
mcking65: Correct, it seems to me that complementary should not be removed.

content/practices/landmark-regions/landmark-regions-practice.html
mcking65: if we were to remove the naming technique from the ro in the table for 'aside", then we'd probably want to remove it from the row about 'section' as well for editorial consistency. I'm not sure which is better, in this case -- brevity or clarity.

daniel-montalvo:
Discussed in [29 April meeting](https://www.w3.org/2026/04/29-ARIA-APG-minutes.html#393f)

content/practices/landmark-regions/landmark-regions-practice.html
NakajimaTakuya: @joelamyman @mcking65 
Sorry for the late reply.
You're right—if we assume the goal is to complement the “main” file, then that makes sense.
I've restored it.

[5c4d7bd](https://github.com/w3c/aria-practices/pull/3418/commits/5c4d7bd287041e860707417318c7ffd69f4a09c9)

content/practices/landmark-regions/landmark-regions-practice.html
NakajimaTakuya: @joelamyman @mcking65 

Thank you. I’ve added the description regarding the scope.
Also, regarding the issue of the text being too long, I’ve adopted the method you suggested to simplify the process for setting accessible names.

f47c74554d651f987f3c554570e883042a048728

content/practices/landmark-regions/landmark-regions-practice.html
NakajimaTakuya: @joelamyman 
Thank you. You were right, so I’ve made the correction.
c105f5a7ef7e80539cf17abc71bafab54a67e391

joelamyman (approved):
Sorry for the late re-review - I think the changes that @NakajimaTakuya has made look good to me, and address all of the feedback provided! 😄

daniel-montalvo:
@NakajimaTakuya

Can you please merge the changes from main into this branch?

We've made changes in the link checker, I have confirmed locally that these resolve the failures in here.

NakajimaTakuya:
@joelamyman 
thank you for the review! 😄 

@daniel-montalvo
Merged the changes from main into this branch. 
Thanks for the heads-up! 😄

PR #3448, opened 2026-06-11
Infrastructure: Update User Agent string in link checker  script  

daniel-montalvo:
This is what's causing the 403s, at least from a W3C perspective. We are proposing to identify this script with it's project name and URI, which these days seems to be more reliable than trying to mimic up-to-date user agent strings.
___
[WAI Preview Link](https://deploy-preview-466--aria-practices.netlify.app/ARIA/apg) _(Last built on Wed, 24 Jun 2026 11:19:20 GMT)._

PR #3451, opened 2026-06-29
Infrastructure: Link checker patch

deniak:
That PR:
- replaces `node-fetch` with the native fetch API, and added browser-like request headers (Accept, Accept-Language, Accept-Encoding) to reduce connection resets from Cloudflare-protected sites
- fixes a retry logic bug: the while loop with a broken guard condition was replaced by a for loop, ensuring failed requests are correctly reported after all retries are exhausted
- adds a step variable to retry log messages to identify whether the failure occurred during the fetch, body read, HTML parsing, or id extraction
- fetches external pages in batches of 5 concurrent requests (instead of all at once) to avoid overwhelming servers
- adds per-page timing logs: each completed page prints its index, URL, and elapsed time
- updates the workflow `.github/workflows/link-checker.yml` so it no longer triggers twice for pull requests: push is now scoped to the main branch only, while pull_request continues to run on all PRs
___
[WAI Preview Link](https://deploy-preview-477--aria-practices.netlify.app/ARIA/apg) _(Last built on Tue, 30 Jun 2026 09:48:19 GMT)._

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,6 @@
     "htmlparser2": "^10.0.0",
     "husky": "^9.1.7",
     "lint-staged": "^15.5.0",
-    "node-fetch": "^2.6.7",
     "prettier": "^3.6.2",
     "selenium-webdriver": "^4.34.0",
     "stylelint": "^16.22.0",
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -1,6 +1,10 @@
 name: Link Checker
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   link-checker:
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,6 @@
         "htmlparser2": "^10.0.0",
         "husky": "^9.1.7",
         "lint-staged": "^15.5.0",
-        "node-fetch": "^2.6.7",
         "prettier": "^3.6.2",
         "selenium-webdriver": "^4.34.0",
         "stylelint": "^16.22.0",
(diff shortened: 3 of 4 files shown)


daniel-montalvo:
@mcking65  My colleague @deniak propses further fixes to the link checker, which has again started to fail after we merged #3448

daniel-montalvo (approved):

package.json
```diff
@@ -55,7 +55,6 @@
     "htmlparser2": "^10.0.0",
     "husky": "^9.1.7",
     "lint-staged": "^15.5.0",
-    "node-fetch": "^2.6.7",
```
mcking65, line 58: @howard-e  do you recall the original motivation for using node-fetch?

howard-e: > @howard-e do you recall the original motivation for using node-fetch?

@mcking65 can't say I do, having come in to it. I'd assume the reason was to align with fetch's parity in the browser environment while it wasn't available natively in node.

[+1 for this change though; my comment from a 2025 dependabot PR:](https://github.com/w3c/aria-practices/pull/2766#issuecomment-2733838735)

> Would recommend we now drop usage of node-fetch. Natively supported [node.js fetch](https://nodejs.org/dist/latest-v22.x/docs/api/globals.html#fetch) has been supported without the experimental flag since v18, and considered stable in v21.

howard-e:
@mcking65 said in https://github.com/w3c/aria-practices/pull/3451#issuecomment-4916935468:

> Maybe we can let this sit for a little while. We might be able to get howard-e or CurtBellow to review

I can take a proper look later today/tomorrow!

howard-e (approved):
Changes look good to me, especially the CF considerations

daniel-montalvo:
Hi @mcking65 

Based on @howard-e comments, is this now ready to merge from your perspective?

mcking65:
thank you @howard-e!!

PR #3452, opened 2026-07-10
All Pages: update skipto.js to version 5.11.2

jongund:
Changes should not change the the functionality on the APG page.
* Removed the use of template elements to create HTML content
* Updated about dialog box content to make shortcut key easier to find
* Focus now goes to the first important element in the dialog
* Focus goes to menu button when about dialog box is closed
___
[WAI Preview Link](https://deploy-preview-478--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 20 Aug 2026 17:48:59 GMT)._

### DIFF
--- a/content/shared/js/skipto.js
+++ b/content/shared/js/skipto.js
@@ -5730,11 +5730,11 @@ dialog button:hover {
   const debug$1 = new DebugLogging('skiptoContent', false);
   debug$1.flag = false;
 
-  /* @class SkipToContent5110
+  /* @class SkipToContent5111
    *
    */
 
-  class SkipToContent5110 extends HTMLElement {
+  class SkipToContent5111 extends HTMLElement {
 
     constructor() {
       // Always call super first in constructor
@@ -6258,7 +6258,7 @@ dialog button:hover {
           if (!isExtensionLoaded) {
             if (!isBookmarkletLoaded) {
               removePageSkipTo();
-              window.customElements.define(BOOKMARKLET_ELEMENT_NAME, SkipToContent5110);
+              window.customElements.define(BOOKMARKLET_ELEMENT_NAME, SkipToContent5111);
               skipToContentElem = document.createElement(BOOKMARKLET_ELEMENT_NAME);
               skipToContentElem.setAttribute('version', skipToContentElem.version);
               skipToContentElem.setAttribute('type', type);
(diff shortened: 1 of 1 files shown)


mcking65:
@jongund 

Could you please edit the description and add a list of changes?

lolaodelola (commented):
I've reviewed this for desktop on MacOS both Chrome, Safari & Firefox. The actual skip-to nav works as expected in that it navigates to the proper places. However, there are some things I noticed about the navigation itself:
1. The skip-to link isn't visible on the page and has to be triggered by keyboard input to become visible, the button has a `display: none` before it's triggered which means screen-readers as well as visual users won't see it unless they know the keybinding.
2. The keybinding provided, opt+0, is only available on desktop, it's unclear how I'd trigger this on mobile which means I couldn't review this on mobile.
3. Using voiceOver, the headings in the skip-to menu drop down aren't read so it's unclear that the first list is landmark regions and the second is headings, which may cause confusion when two different items lead to the same space on the page e.g. 'Complementary Feedback' and 'Help Improve this Page'.

mcking65 (changes requested):
Here is what I found from testing in Chrome on Windows using JAWS and NVDA and on iOS in Safari.

* Google Chrome Version 150.0.7871.187
* JAWS version 2026.2605.43.400 
* NVDA version 2026.1.1

#### About Dialog Content


The dialog contains new content that makes the content more difficult to understand than in the previous version.

The new content includes the following:

>"Skip To Content" Button Shortcut
>Alt+0: Open Menu

Semantically, the first line is a heading and the secondline is content under the heading. It is difficult for a screen reader user to figure that out, espcially because:
 * The content is redundnat with the heading. 
 * The content sounds like a menu button widget.

The addition of this new shortcut content makes the overall structure of the dialog content more difficult to parse. The words "Purpose" and "privacy" are floating in the middle of other words. They lack appropriate structure and/or punctuation that would make the content readily understandable.

On iOS, there is some strange additional content that includes the word "games" mixed in with the shortcut content.

One possible solution:

~~~
<h2>Keyboard Shortcut</h2>
<p>Alt+0: Opens the skip to menu and focuses the first landmark menu item.</p>
~~~

Then, add headings for Purpose and Privacy and put a `<hr>` after the privacy statement.

#### Duplicate close buttons

Does the about dialog need two close buttons?

#### Initial focus

>Focus now goes to the dialog element when the about dialog box opens, not the close button

This is counter to APG guidance, which recommends initially focusing the heading containing the title.

A quick test with JAWS and NVDA in Chrome and a not so quick test with iOS Safari is showing that focusing the dialog vs content inside works somewhat better than it did when we last updated that guidance. I didn't test with Firefox, Edge, or anything on macOS (I no longer have a Mac).

We could consider revising the dialog initial focus guidance, but that would require a separate issue and discussion to assess the ramifications and get to consensus. That would likely take a while.

#### Focus after dialog close

Compared to prod, this new version seems to more frequently lose focus after the about dialog is closed. Focus goes to the document instead of the menu button.

lolaodelola:
Regrets for today's call, I have a conflict but will be able to review anything later.

jongund:
@mcking65 
Updated SkipTo.js based on your feedback:
1. Removed first close button
2. Focus now on first important topic, instead of the dialog element
3. Updated information in about box on button keyboard shortcut
4. Focus goes to about menu button when dialog box is closed

mcking65 (approved):
@jongund 

Looks good; thank you for the fixes!

PR #3455, opened 2026-07-22
Quantity Spin Button Example: Add ariaNotify progressive enhancement

adampage:
Re: #3414 and [its discussion in the Wed, July 22, 2026 APG TF call](https://www.w3.org/2026/07/22-aria-apg-minutes.html#1fb4).

A few quick notes:

- By default, the main implementation change —  done in b915a5b — conservatively assumes that `ariaNotify` is **_not_** supported by the user’s browser. It leaves `<output>` live regions in the example’s static HTML and only removes them from the DOM (and bypasses their corresponding content injections) after confirming that `ariaNotify` is available.

- I [updated the regression tests](https://github.com/w3c/aria-practices/commit/15788ff762770978a80b665cc83bf1784e9ca922) to account for either announcement technique depending on the test browser’s capabilities, which was fun. 😅 

- I also added some [new content in the Accessibility Features section](https://github.com/w3c/aria-practices/commit/21f19c9785c245d6bc0caa24a4e18a3c26c87d41).

- While testing, I noticed that the latest macOS Safari no longer [announces the updated value as a percentage](https://www.w3.org/2025/09/09-aria-apg-minutes.html#9c4c:~:text=Adam_Page%3A%20I%20have,it%20currently%20does%20not) when `aria-valuetext` is added. So I indulged in a little scope creep to add [`aria-valuetext` functionality](https://github.com/w3c/aria-practices/commit/5b0b0aaed525e10b2d365021b2099d0d25c66e6b), [additional documentation](https://github.com/w3c/aria-practices/commit/d432a1750c78b4f3956772229ffec5dd87f413a8), and [regression tests](https://github.com/w3c/aria-practices/commit/c2f0cf028513b4fb163eb2802036613bf53fd2f7). Would it be better for me to extract this out as a separate PR?

cc: @mcking65

Issue #3414, opened 2026-02-11
Add guidance and examples for use of AriaNotify


spectranaut:
We just landed :) https://github.com/w3c/aria/pull/2577

adampage:
I used live regions in the [Quantity Spin Button example](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/examples/quantity-spinbutton/) to announce value changes for the increment/decrement buttons.

I could update that to use `ariaNotify` as a progressive enhancement with live regions as the fallback. What do you think, @mcking65?

spectranaut:
Internationalization working group requests that there is an example of how to use ariaNotify when ariaNotify should be in a different language than the majority of the web content: https://github.com/w3c/aria/issues/2828#issuecomment-5281877123

cookiecrook:
Specifically at least three:
- The two primarily expected translation cases:
  - browser auto-translation (nothing for author to do but make sure the lang/dir is correct)
  - author/webapp-translation (translated server- or client-side, and author needs to update the lang/dir at time of translation, in addition to translating any strings)
- as well as the expectedly more rare case "element courier" case:
  - the web-app author needs a one-off notification in an lang/dir that is not used on the page, and therefore generates a new "courier" element to hang the notification from. append a new or utilize an existing empty element to hang the correct lang/dir off of (Noite: This was the agreement we made with i18n for https://github.com/w3c/aria/issues/2828)

PR #3461, opened 2026-08-05
Quantity Spin Button Example: Add ariaNotify progressive enhancement

adampage:
Migrated from [fork PR #3455](https://github.com/w3c/aria-practices/pull/3455) to a branch in the main repo.

Preview deployed to the [Quantity Spin Button Example](https://deploy-preview-482--aria-practices.netlify.app/aria/apg/patterns/spinbutton/examples/quantity-spinbutton/).

***

Re: #3414 and [its discussion in the Wed, July 22, 2026 APG TF call](https://www.w3.org/2026/07/22-aria-apg-minutes.html#1fb4).

A few quick notes:

- By default, the main implementation change —  done in b915a5b — conservatively assumes that `ariaNotify` is **_not_** supported by the user’s browser. It leaves `<output>` live regions in the example’s static HTML and only removes them from the DOM (and bypasses their corresponding content injections) after confirming that `ariaNotify` is available.

- I [updated the regression tests](https://github.com/w3c/aria-practices/commit/15788ff762770978a80b665cc83bf1784e9ca922) to account for either announcement technique depending on the test browser’s capabilities, which was fun. 😅 

- I also added some [new content in the Accessibility Features section](https://github.com/w3c/aria-practices/commit/21f19c9785c245d6bc0caa24a4e18a3c26c87d41).

- ~~While testing, I noticed that the latest macOS Safari no longer [announces the updated value as a percentage](https://www.w3.org/2025/09/09-aria-apg-minutes.html#9c4c:~:text=Adam_Page%3A%20I%20have,it%20currently%20does%20not) when `aria-valuetext` is added. So I indulged in a little scope creep to add [`aria-valuetext` functionality](https://github.com/w3c/aria-practices/commit/5b0b0aaed525e10b2d365021b2099d0d25c66e6b), [additional documentation](https://github.com/w3c/aria-practices/commit/d432a1750c78b4f3956772229ffec5dd87f413a8), and [regression tests](https://github.com/w3c/aria-practices/commit/c2f0cf028513b4fb163eb2802036613bf53fd2f7). Would it be better for me to extract this out as a separate PR?~~
  **Aug 19**: removed `aria-valuetext` per [APG meeting decision](https://github.com/w3c/aria-practices/pull/3461#issuecomment-5345038316)

cc: @mcking65 
___
[WAI Preview Link](https://deploy-preview-482--aria-practices.netlify.app/ARIA/apg) _(Last built on Thu, 17 Sep 2026 23:37:45 GMT)._


mcking65:
We should discuss the use of aria-valuetext given that the spec says:

>Authors SHOULD only set the aria-valuetext attribute when the rendered value cannot be meaningfully represented as a number. For example, a slider might have rendered values of small, medium, and large. In this case, the values of aria-valuenow could range from 1 through 3, which indicate the position of each value in the value space, but the aria-valuetext would be one of the strings: small, medium, or large.

joelamyman (approved):
I've stepped through the changes, and tested using screen readers, and I think this is all working as expected! I tested the preview with:

- Safari + VoiceOver, noting the slightly strange percentage announcements but after reviewing the  conversation in the last meeting, I can see that unfortunately, this is expected
- NVDA + Firefox
- JAWS + Chrome

I'm slightly less familiar with writing tests, but can see that the updates wrap existing live region-related tests with an if statement, so that they don't run if using ariaNotify. Additional tests have also been written for when ariaNotify is used to check things like ariaNotify has announced the expected number of times.

Please let me know if there's anything else I can do to help on this!

content/patterns/spinbutton/examples/quantity-spinbutton.html
```diff
@@ -247,6 +261,16 @@ <h2>Accessibility Features</h2>
               </li>
             </ul>
           </li>
+          <li>
+            When a spin button’s value is changed:
+            <ul>
+              <li>
+                Its <a
+                href="https://www.w3.org/TR/wai-aria-1.3/#spinbutton"><code>aria-valuenow</code>
+                attribute is updated to reflect the new integer value</a>.
```
mcking65:
```suggestion
                Its <a href="#aria-valuenow" class="property-reference"><code>aria-valuenow</code> attribute</a>
                is updated to reflect the new integer value.
```

content/patterns/spinbutton/examples/quantity-spinbutton.html
```diff
@@ -247,6 +261,16 @@ <h2>Accessibility Features</h2>
...
+              </li>
+            </ul>
+          </li>
```
mcking65:
```suggestion
          <li>
            When a spin button’s value is changed its
            <a href="#aria-valuenow" class="property-reference"><code>aria-valuenow</code> attribute</a>
            is updated to reflect the new integer value.
          </li>
```

content/patterns/spinbutton/examples/quantity-spinbutton.html
```diff
@@ -232,9 +232,23 @@ <h2>Accessibility Features</h2>
...
+                    <code>ariaNotify</code> is supported by the user’s browser.
+                  </li>
+                  <li>
+                    Otherwise, by injecting the updated value into an invisible
+                    live region. The live region empties its contents after 3
+                    seconds to avoid leaving stale content in the document.
+                  </li>
+                </ul>
               </li>
               <li>
```
mcking65, line 253: I think these lines are accidentally in the wrong place. I added a suggestion below that puts them at the current line 261. If you accept those changes, you can delete them from here.

content/patterns/spinbutton/examples/quantity-spinbutton.html
mcking65, line 273: Note that we do not use absolute URLs for ARIA spec references; use `class="property-reference" and href="#aria-valuenow"`

```suggestion
          <li>
            When a spin button’s value is changed:
            <ul>
              <li>
                Its
                <a href="#aria-valuenow" class="property-reference"><code>aria-valuenow</code> attribute</a>
                is updated to reflect the new integer value.
              </li>
              <li>
                Screen reader announcement of the new value is triggered using one of these two techniques:
                <ul>
                  <li>
                    By calling the JavaScript
                    <a
                    href="https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaNotify"><code>ariaNotify(announcement)</code></a>
                    method. This technique is implemented as a progressive
                    enhancement only after dynamically confirming that
                    <code>ariaNotify</code> is supported by the user’s browser.
                  </li>
                  <li>
                    Otherwise, by injecting the updated value into an invisible
                    live region. The live region empties its contents after 3
                    seconds to avoid leaving stale content in the document.
                  </li>
                </ul>
              </li>
            </ul>
          </li>
```

mcking65 (changes requested):
It's looking good. I suggested only a couple of changes.

content/patterns/spinbutton/examples/quantity-spinbutton.html
adampage: @mcking65 ah I see, thanks —  I didn’t know about that technique for referencing the ARIA spec. 👍🏻

content/patterns/spinbutton/examples/quantity-spinbutton.html
adampage: @mcking65, I did put them here on purpose since the custom announcements will only fire (whether by live region or ariaNotify) when a user activates the increment and decrement buttons specifically.

If a user inputs a value directly or uses their arrow keys to increment/decrement, there’ll be no custom announcement, just the screen reader’s normal behavior for responding to that interaction.

I went ahead and accepted your [follow-up code suggestion](https://github.com/w3c/aria-practices/pull/3461#discussion_r4022301654) to get the fixed ARIA spec reference, but then backed out the relocation of those notes in [a separate commit](https://github.com/w3c/aria-practices/pull/3461/commits/edb2e20c2ca98cb1103d1570014e54f793c0d8d4).

All that said, have I persuaded you? 😅

